PR #7, opened 2015-02-11
Use names or IDs for calicoctl addtogroup

spikecurtis:
Adds docker-py as a dependency.  We should consider eventually phasing out the sh-based Docker commands we have now.


tomdee:
LGTM - merged.

PR #13, opened 2015-02-12
Streamline dockerfiles

fasaxc:
* Move all the files that we copy in into an overlay tree.
* Combine apt-get calls.
* Remove dist-update (best practice doc says we should rely on upstream for that)
* Combine apt-get update with the following installs so that we can't cache an out-of-date APT packages list and cause the install to fail.
* Make sure we clean up tar extractions etc before the end of a RUN line.


node/Dockerfile
```diff
@@ -4,60 +4,61 @@ FROM phusion/baseimage:0.9.16
...
+    echo "Pin: origin binaries.projectcalico.org" >> /etc/apt/preferences && \
+    echo "Pin-Priority: 1001" >> /etc/apt/preferences
+
+# Install dependencies shared with the master node.  If we can keep this
+# in sync then we'll share an FS layer with the master to speed downloads.
+RUN apt-get update && \
+    apt-get install -qy \
+        calico-felix \
+        calico-acl-manager \
+        curl \
```
tomdee: This line can be removed

tomdee:
As just discussed, can you:
get rid of the apt-get install curl
combine root and home overlays
Use docopt defaulting for the images

fasaxc:
Pushed as squashed commit 822ad3c00055332edeee24d4b10c232b50f8b16f

PR #19, opened 2015-02-18
Reinstate missing '/config' directory for confd.

Symmetric:
Commit 822ad3c00055332edeee24d4b10c232b50f8b16f removed the
/config directory from the node & master Dockerfiles.

This is required for confd to function.

FIXES metaswitch/calico-docker#18


spikecurtis:
Dockerfiles got refactored to make a smaller image.  However, when we did it, they didn't realize that git won't commit empty directories, so that didn't make it into the repo.  That has now been fixed on master.

PR #20, opened 2015-02-18
Refactor datastore code into new file and add a test

tomdee:


spikecurtis:
There is some inconsistency in calicoctl.py functions.  add_container still raises exceptions, rather than printing stuff for the user.  I *think* that the goal is that all functions in calicoctl.py are "user facing" at this point.  If that's not true, let's find a way to indicate which are and which are not (e.g. _private_method or some such).

tomdee:
Yep, just fixing that up at the moment

PR #24, opened 2015-02-19
Fix handling of netns with no default-route.

Symmetric:
netns didn't handle the case with no default route;
the 'ip route | grep default' call returns 1 if there is no match,
and therefore the check_output call fails.

Fix is to parse the returned 'ip route' output in python instead.

Fixes Metaswitch/calico-docker#23


Symmetric:
I was unable to repro the case where there is a route in the netns; Merger, please give this branch a quick spin if possible before merging.

PR #25, opened 2015-02-20
Add Travis CI build for calicoctl

fasaxc:
* Added scripts to install deps and create a virtualenv
* Changed scripts to use dash-case; seemed prettier/more shell-ish
* added basic .travis.yml which, so far, only builds the calicoctl binary.


fasaxc:
Travis build is here: https://travis-ci.org/Metaswitch/calico-docker

tomdee:
I would probably prefer that we don't use a virtualenv and just just a docker container for builds - like this sort of approach https://github.com/docker/fig/blob/master/script/build-linux

Lukasa:
Leaving this here in case anyone's interested: [pydockerize](https://github.com/msabramo/pydockerize).

PR #26, opened 2015-02-23
Decouple powerstrip

Symmetric:
As part of raising a bug in the Powerstrip project, I found it easiest to create a repro by pulling the powerstrip adapter into its own container.

We've discussed this as something that might be desirable; would you take a pull request to decouple the powerstrip adapter alone?

If so I'll do a bit more polish on this branch before merging, just wanted to give you a look at the code before I spent any more time on this.


spikecurtis:
Decoupling the node is a larger conversation.  If we pull powerstrip out, should we also consider breaking up the other Calico bits into a micro services architecture?

Note that you can still compose Calico with other powerstrip adapters by starting up separate powerstrip and pointing it at localhost:2378 for the Calico adapter.

grkvlt:
@Symmetric @spikecurtis this looks interesting, particulalrly in view of wanting to use e.g. Flocker with Calico inside Clocker. I would rather not run multipe Powerstrip adapters chained together.

spikecurtis:
@grkvlt You'll need a Powerstrip adapter for each extension, e.g. one for Calico, one for Flocker.  Is this what you mean?  Are you just trying to avoid running multiple instances of Powerstrip itself?

grkvlt:
@spikecurtis yes, I was wanting to only run one Powestrip instance, I know I'll need to run an adapter service for each extension. It would simplify `DockerHost` setup in Clocker and also prevent the proliferation of large numbers of infrastructure containers.

lxpollitt:
Note that Powerstrip is a temporary thing and we expect to replace it as part of the Docker 1.7 release (mid-June) which will support the Docker network driver plug-in API.  For that API we need to keep Calico as as single container.  So I would caution doing any work in this area or splitting Calico into multiple containers before then.

Symmetric:
I don't think it's a hard requirement to have only one container for the docker plugin -- a possible alternative would be to mount the host docker socket into the network plugin, and start the felix/bird containers from there if they are not running.

For now I think that the hackiness of that workaround slightly outweighs the benefits of disaggregating the containers, but we should be clear about what our options are.

Long-term I think we want calico to run as a rkt pod or docker-compose bundle, when Docker networking supports launching a compose file (per discussion with @tomdee).

Symmetric:
Just putting this here for if/when we get round to this:

[AlpineLinux](https://github.com/gliderlabs/docker-alpine) has a very compact distro that's got better packaging than busybox, with comparable (~15MB) container size.

Python containers can be created in a [super-minimal](http://yasermartinez.com/blog/posts/creating-super-small-docker-images.html) fashion as well, O(10MB) instead of O(100MB)

tomdee:
+1 for Alpine. I was playing with it last week and it seems really nice.

Symmetric:
Just saw a presentation from Robert Witoff at Coinbase -- looks like the "mount the docker socket" aka Global Container paradigm is more widespread than I thought -- c.f. [logspout](https://github.com/gliderlabs/logspout) for a widely used project that does this.

Might not be a bad idea to go down this route after all.

Symmetric:
This PR is outdated; if we do implement proper container separation, it won't be based off this.

PR #27, opened 2015-03-02
Fix docker-py version mismatch in requirements.txt

Symmetric:
apt repo version of docker not yet updated to v1.17 API.


spikecurtis:
I've updated master to explicitly pin the Docker API version when we start up the Docker client.  This means we can use the latest version of docker-py even on backlevel versions of Docker.

If docker-py changes *its* API, then we should pin the version.

spikecurtis:
Pinning to v1.16 is on master.  Closing this.

PR #28, opened 2015-03-02
Pin requests.py version in requirements.txt

Symmetric:
Latest (2.5.2 & 2.5.3) requests.py interacts badly with pyinstaller.

See
http://stackoverflow.com/questions/28775276/pyinstaller-error-importerror-no-module-named-requests-packages-chardet-sys


Lukasa:
FYI, a fix for this is in progress. Should be released in the next few days.

PR #31, opened 2015-03-05
Add support for IPv6 networked containers

spikecurtis:
High level changes:
 - add commands for IPv6 address pools
 - add bird6 to calico-node with Confd configuration
 - move to latest calico-felix (including zeroMQ API update for interface names) for proxy NDP config
 - program containers with local gateway, rather than default via device directly
 - version checking of IP address to program netns correctly

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -59,6 +59,7 @@ mkdir = sh.Command._create('mkdir')
 docker = sh.Command._create('docker')
 modprobe = sh.Command._create('modprobe')
 grep = sh.Command._create('grep')
+sysctl = sh.Command._create("sysctl")
 
 mkdir_p = mkdir.bake('-p')
 
@@ -373,7 +374,8 @@ class CalicoDockerClient(object):
     """
 
     def __init__(self):
-        self.docker_client = pydocker.Client(base_url='unix://var/run/docker.sock')
+        self.docker_client = pydocker.Client(base_url='unix://var/run/docker.sock',
+                                             version="1.16")
 
     def get_container_id(self, container_name):
         """
@@ -477,6 +479,10 @@ def node(ip, node_image, ip6=""):
 
     # Set up etcd
     client = CalicoCmdLineEtcdClient(etcd_authority)
+
+    # Enable IP forwarding.
+    sysctl("-w", "net.ipv4.ip_forward=1")
+    sysctl("-w", "net.ipv6.conf.all.forwarding=1")
 
     master_ip = client.get_master()
     if not master_ip:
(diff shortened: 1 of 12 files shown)


spikecurtis:
Don't merge this until we get an new Felix release!  The zeroMQ API is changed, so it will not work with current Felix.

calicoctl.py
```diff
@@ -228,6 +235,10 @@ def node(ip, node_image):
     # Set up etcd
     client = DatastoreClient()
 
+    # Enable IP forwarding.
```
Symmetric: Might be beneficial to future generations to describe what we're enabling forwarding for.

calicoctl.py
```diff
@@ -237,9 +248,13 @@ def node(ip, node_image):
     if not ipv4_pools:
         print "No IPv4 range defined.  Exiting."
         return
+    ipv6_pools = client.get_ip_pools("v6")
+    if not ipv6_pools:
```
Symmetric: Is it an error to have no ipv6_pools if we're not configured with an IPv6 address? Looks like we bail in that case right now.

Symmetric: Just spotted that we default the pools in master() below. Probably worth a comment here to explicitly call out that we're defaulting the pools elsewhere.

master/root_overlay/plugin.py
```diff
@@ -185,6 +185,7 @@ def do_ep_api():
             rsp = {"rc": "SUCCESS",
                    "message": "Hooray",
                    "type": fields['type'],
+                   "interface_prefix": "tap",
```
Symmetric: Comment this? Unclear why it's being added.

master/root_overlay/plugin.py
```diff
@@ -242,9 +243,11 @@ def send_all_eps(create_sockets, host, resync_id):
     # Send all of the ENDPOINTCREATED messages.
     for ep in eps_by_host.get(host, {}):
         log.info("Sending ENDPOINTCREATED message for endpoint %s", ep)
+        if_name = "tap" + ep[:11]
```
Symmetric: What's 11?

node/root_overlay/adapter/netns.py
```diff
@@ -60,18 +64,25 @@ def remove_endpoint(ep_id):
...
+
+def set_up_endpoint(ip, cpid, next_hop_ips,
+                    in_container=False,
+                    veth_name=VETH_NAME,
+                    proc_alias=PROC_ALIAS):
     """
     Set up an endpoint (veth) in the network namespace idenfitied by the PID.
 
-    :param ip: The IP address to assign to the endpoint (veth)
+    :param ip: The IP address to assign to the endpoint (veth) as Netaddr IPAddress.
```
Symmetric: Much though it pains me to make this complaint, these lines are >79 chars long.

Symmetric:
@spikecurtis - some doc/comment/formatting suggestions, but functionally LGTM.

PR #35, opened 2015-03-13
Fix calicoctl container add to work with IPv6 changes.

spikecurtis:


spikecurtis:
@neiljerram I don't think so---the changes I made were to bring the size down to 79 characters.  (We used 99 on Gulfstream, so some of the early Python in this project had that width.)

tomdee:
Did you just copy that code in datastore.py from the powerstrip code? Probably OK, but we should sort all this copy-paste stuff out when we get rid of master.

Other than that, looks fine.

spikecurtis:
@tomdee yeah, datastore.py and calico_etcd.py need to be merged once we get rid of master.

spikecurtis:
@Symmetric I don't think we should be making exceptions generic---the caller of get_container_info() needs to be able to distinguish between missing keys and more fundamental errors (e.g. Docker isn't running) in order to give a sensible response to the end user.

PR #38, opened 2015-03-17
Fix some trivial typos in readme.

plwhite:


plwhite:
Some typo fixes - should be totally unexciting. This file has a mix of DOS and UNIX line endings, which I have NOT fixed in this checkin (but seem to show up in some diffs).

PR #39, opened 2015-03-18
Etcd felix prototype

spikecurtis:


.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: python
```
tomdee: This file should be deleted

calicoctl.py
```diff
@@ -650,14 +579,13 @@ def process_output(line):
 if __name__ == '__main__':
     arguments = docopt(__doc__)
     if os.geteuid() != 0:
-        print "calicoctl must be run as root"
+        print >> sys.stderr, "calicoctl must be run as root."
+        exit(2)
```
tomdee: This exit is broken

calicoctl.py
```diff
@@ -650,14 +579,13 @@ def process_output(line):
...
+    elif not docker:
```
tomdee: This elif block should be removed

tests/scale.sh
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
tomdee: Should be removed

tomdee:
A few comments but in general I'm happy to merge this whenever we're ready

PR #47, opened 2015-03-31
Basic IPAM support

tomdee:


calicoctl.py
```diff
@@ -115,6 +115,9 @@ def container_add(container_name, ip):
     :param container_name: The name or ID of the container.
     :param ip: An IPAddress object with the desired IP to assign.
     """
+    if os.geteuid() != 0:
+        print >> sys.stderr, "`calicoctl container add` must be run as root."
```
spikecurtis: can you comment why this needs to be root?

node/adapter/ipam.py
```diff
@@ -0,0 +1,115 @@
...
+    def assign_address(self, pool, address):
+        """
+        Attempt to assign an IPAddress in a pool.
+        Fails if the address is already assigned.
+        The directory for storing assignments in this pool must already exist.
+
+        :param IPNetwork pool: The pool that the assignment is from.
+        :param IPAddress address: The address to assign.
+
+        :return: True if the allocation succeeds, false otherwise.
```
spikecurtis: This isn't very Pythony.   You should throw an exception instead.  This is useful because there are lots of different problems we might signal in the future:
 - etcd down
 - pool didn't exist
 - address wasn't available,
etc.

node/adapter/ipam.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self):
+        # Init an etcd client.
+        self.etcd = IPAMClient()
+
+    def allocate(self, pool):
+        """
+        Attempt to allocate an IP address from the provided pool.
+
+        :param IPNetwork pool: The pool to allocate from
+        :return: An IP address which has been allocated or None
```
spikecurtis: Again, throw an exception if we fail.  It's really easy to write didn't-check-the-return-value bugs that are hard to diagnose.

calicoctl.py
```diff
@@ -197,6 +209,15 @@ def container_remove(container_name):
         print "Container %s is in security groups %s. Can't remove." % \
               (container_name, groups)
 
+    # Remove any IP address assignments that this endpoint has
+    endpoint = client.get_endpoint(hostname, container_id, endpoint_id)
+    for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
+        ip = net.ip
```
spikecurtis: assert here that we have /32 or /128

tests/unit/ipam_test.py
```diff
@@ -0,0 +1,46 @@
+from netaddr import IPNetwork, IPAddress
+from nose.tools import assert_equal, assert_true
+from node.adapter.ipam import SequentialAssignment, IPAMClient
+
+pool = IPNetwork("192.168.0.0/16")
+client = IPAMClient()
+
+class TestIPAM:
```
spikecurtis: should be TestIPAMClient for clarity

tests/unit/ipam_test.py
```diff
@@ -0,0 +1,46 @@
+from netaddr import IPNetwork, IPAddress
+from nose.tools import assert_equal, assert_true
+from node.adapter.ipam import SequentialAssignment, IPAMClient
+
+pool = IPNetwork("192.168.0.0/16")
+client = IPAMClient()
```
spikecurtis: you haven't mocked out etcd, so this will modify live data, and will fail if etcd isn't running.

calicoctl.py
```diff
@@ -142,7 +146,11 @@ def container_add(container_name, ip):
     ip = IPAddress(ip)
     version = "v%s" % ip.version
     pools = client.get_ip_pools(version)
-    if not any([ip in pool for pool in pools]):
+    for candidate_pool in pools:
+        if ip in candidate_pool:
+            pool = candidate_pool
+
+    if pool is None:
         print "%s is not in any configured pools" % ip
```
alexwlchan: I think this needs a `pool = None` line before the for loop. Otherwise, if we don’t find the IP in any candidate pools, the value of `pool` will be unassigned and we’ll throw a NameError when we hit the if statement.

node/adapter/datastore.py
```diff
@@ -399,21 +391,33 @@ def get_ep_id_from_cont(self, container_id):
...
-    def create_container(self, hostname, container_id, endpoint):
+    def get_endpoint(self, hostname, container_id, endpoint_id):
         """
-        Set up a container in the /calico/ namespace on this host.  This function assumes 1
-        container, with 1 endpoint.
+        Get all of the details for a single endpoint.
+
+        :param endpoint_id: The ID of the endpoint
+        :return:  an Endpoint Object
+        """
```
alexwlchan: Should this list of parameters include `hostname` or `container_id`?

node/adapter/ipam.py
```diff
@@ -0,0 +1,117 @@
...
+                # We've found an address to try.
+                if self.etcd.assign_address(pool,
+                                            IPAddress(candidate_address)):
+                    return candidate_address
+
+    def _get_next(self, pool, assigned):
+        """
+        Gets the next address in a range.
+        :param IPNetwork pool: The pool to allocate from
+        :return: the next IP address to try (a string), or None if the pool
```
alexwlchan: Should the list of parameters in the docstring include `assigned`?

PR #48, opened 2015-04-01
Profile Commands

spikecurtis:
Add & remove tags
Update profile rules

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -17,6 +17,7 @@ Usage:
   calicoctl profile <PROFILE> tag remove <TAG>
   calicoctl profile <PROFILE> rule show
   calicoctl profile <PROFILE> rule json
+  calicoctl profile <PROFILE> rule update
   calicoctl profile <PROFILE> member add <CONTAINER>
   calicoctl ipv4 pool add <CIDR>
   calicoctl ipv4 pool del <CIDR>
@@ -55,7 +56,8 @@ from prettytable import PrettyTable
 
 from node.adapter.datastore import (DatastoreClient,
                                     ETCD_AUTHORITY_ENV,
-                                    ETCD_AUTHORITY_DEFAULT)
+                                    ETCD_AUTHORITY_DEFAULT,
+                                    Rules)
 from node.adapter import netns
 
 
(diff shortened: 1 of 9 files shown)


calicoctl.py
```diff
@@ -490,12 +470,111 @@ def group_show(detailed):
...
+def profile_tag_show(name):
+    """Show the tags on the profile."""
+    try:
+        profile = client.get_profile(name)
+    except KeyError:
+        print "Profile %s not found." % name
+        sys.exit(1)
+
+    for tag in profile.tags:
+        print tag
```
tomdee: I'd prefer that we carry on using PrettyTable for consistency.

calicoctl.py
```diff
@@ -490,12 +470,111 @@ def group_show(detailed):
...
+
+def profile_rule_show(name, human_readable=False):
+    """Show the rules on the profile."""
+    try:
+        profile = client.get_profile(name)
+    except KeyError:
+        print "Profile %s not found." % name
+        sys.exit(1)
+
+    if human_readable:
```
tomdee: Again, PrettyTable would be good here.

calicoctl.py
```diff
@@ -490,12 +470,111 @@ def group_show(detailed):
...
+        print "Inbound rules:"
+        i = 1
+        for rule in profile.rules.inbound_rules:
```
tomdee: Use enumerate rather than i=1 and i +=1
https://docs.python.org/2/library/functions.html#enumerate

tomdee:
LGTM - I'd prefer that we stick with prettytable for output and one minor point on using enumerate.

calicoctl.py
spikecurtis: Are you ok with keeping the human readable description?  I don't want a bunch of mostly empty cells from breaking out the rule fields individually.

tomdee: Yes, that sounds sensible.

spikecurtis: I mocked this up and I must admit I don't like it.  We're not displaying tabular data (it's two lists), so readability goes down with this approach.

tomdee: OK, fair enough. I’m happy with whatever you think is sensible.

Whilst I think it’s good to be consistent, “A Foolish Consistency is the Hobgoblin of Little Minds”

PR #51, opened 2015-04-07
Datastore Unit Tests

spikecurtis:


tests/unit/datastore_test.py
```diff
@@ -2,9 +2,195 @@
...
+class TestRule(unittest.TestCase):
+
+    def test_create(self):
+        """
+        Test creating a rule from constructor.
+        """
+        rule1 = Rule(action="allow",
+                     src_tag="TEST",
+                     src_ports=[300, 400])
+        self.assertDictEqual({"action": "allow",
```
Symmetric: I  don't think you _need_ to fix this, but nose has bindings to allow more idiomatic (and PEP8-compliant) function calls, e.g. assert_dict_equal, setup, teardown, etc.

Not well documented, but I'm pretty sure it's there.

tests/unit/datastore_test.py
```diff
@@ -2,9 +2,195 @@
...
+        except KeyError:
+            pass
+        else:
+            self.assertTrue(False, "Should raise key error.")
+
+    def test_wrong_value(self):
+        """
+        Test that instantiating a Rule with action not allow|deny will fail.
+        """
+        try:
```
Symmetric: More 'noseish' to use self.assert_raises instead of writing the boilerplate to try/catch? I think it's less code & clearer.

There's also the @raises decorator, but I prefer the former option.

spikecurtis: That didn't seem to work for constructors :(

tests/unit/datastore_test.py
```diff
@@ -2,9 +2,195 @@
...
+                     protocol="icmp",
+                     icmp_type=8,
+                     src_net=IPNetwork("10/8"))
+        self.assertEqual("allow icmp type 8 from 10.0.0.0/8",
+                         rule4.pprint())
+
+
+class TestEndpoint(unittest.TestCase):
+
+    def test_to_json(self):
```
Symmetric: Worth doing invariant testing here too?

e.g. Endpoint.from_json(e.to_json()) == e

spikecurtis: I do that in test_from_json.

Note that we don't have a method that checks Endpoint equality, so I just assert equal values.

tests/unit/datastore_test.py
```diff
@@ -90,6 +331,586 @@ def test_profile_update_rules(self):
...
+    def test_delete_profile(self):
+        """
+        Test deleting a policy profile.
+        """
+        self.datastore.delete_profile("TEST")
+        self.etcd_client.delete.assert_called_once_with(TEST_PROFILE_PATH,
+                                                        recursive=True,
+                                                        dir=True)
+
+    def test_get_profile_names_2(self):
```
Symmetric: why the '_2'?

tests/unit/datastore_test.py
```diff
@@ -90,6 +331,586 @@ def test_profile_update_rules(self):
...
+        self.assertSetEqual(profiles, set())
+
+    def test_get_profile_members(self):
+        """
+        Test get_profile_members() when there are endpoints.
+        """
+        self.etcd_client.read.side_effect = mock_read_4_endpoints
+        members = self.datastore.get_profile_members("TEST")
+        self.assertSetEqual({"567890abcdef", "7890abcdef12"},
+                            set(members))
```
Symmetric: Why are you casting the members result to a set? That could be a destructive operation (e.g. in the case members = [1,2,2]). If you're trying to guarantee a stable ordering, you could use `members.sorted()`.

tests/unit/datastore_test.py
```diff
@@ -90,6 +331,586 @@ def test_profile_update_rules(self):
...
+        members = self.datastore.get_profile_members("UNIT_TEST")
+        self.assertSetEqual(set(), set(members))
+
+    def test_get_endpoint_exists(self):
+        """
+        Test get_endpoint() for an endpoint that exists.
+        """
+        ep = Endpoint("1234567890ab", "active", "11-22-33-44-55-66")
+        self.etcd_client.read.side_effect = mock_read_for_endpoint(ep)
+        ep2 = self.datastore.get_endpoint("TEST_HOST", "1234", "1234567890ab")
```
Symmetric: Here, hostname, container_id, and ep_id should be global constants IMO -- they are baked into the TEST_ENDPOINT_PATH that mock_read_for_endpoint uses.

Symmetric: I'd prefer to use explicit kwargs here, since it's not immediately obvious what "1234" is.

Making the change above might obviate the need to be more explicit though.

Symmetric:
A couple minor naming suggestions which are optional.

LGTM.

tests/unit/datastore_test.py
spikecurtis: 2 profiles get returned.

PR #53, opened 2015-04-08
Allow non calico powerstrip

spikecurtis:
Fix an issue where Powerstrip-calico always sets net=None for containers.  Instead, only do this if CALICO_IP is defined in the environment variables.

This allows non-Calico networked containers, for example, if you need to start a container with host networking.


spikecurtis:
Only the last commit is different from the other open pull request for datastore.py UTs.  Review and merge that one first!

Symmetric:
One minor simplification suggestion, otherwise LGTM.

PR #54, opened 2015-04-08
Fix bug in powerstrip-calico.py for undefined pools

spikecurtis:


spikecurtis:
Only last commit is different than other PRs.

Symmetric:
LGTM

PR #57, opened 2015-04-10
Fix calicoctl.py for missing /etc/defaults/docker

Symmetric:
If the /etc/defaults/docker file was missing, calicoctl would fall
over. this file is not mandatory, so we should tolerate its
absence.

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -302,8 +302,10 @@ def node(ip, force_unix_socket, node_image, ip6=""):
 
         # Set the docker daemon to listen on the docker.real.sock by updating
         # the config, clearing old sockets and restarting.
-        socket_config_exists = \
+        socket_config_exists = (
+            os.path.isfile(DOCKER_DEFAULT_FILENAME) and
             DOCKER_OPTIONS in open(DOCKER_DEFAULT_FILENAME).read()
+        )
         if not socket_config_exists:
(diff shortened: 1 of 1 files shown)


calicoctl.py
```diff
@@ -302,8 +302,10 @@ def node(ip, force_unix_socket, node_image, ip6=""):
 
         # Set the docker daemon to listen on the docker.real.sock by updating
         # the config, clearing old sockets and restarting.
-        socket_config_exists = \
+        socket_config_exists = (
+            os.path.isfile(DOCKER_DEFAULT_FILENAME) and
             DOCKER_OPTIONS in open(DOCKER_DEFAULT_FILENAME).read()
+        )
```
spikecurtis: I don't think this works.  If DOCKER_DEFAULT_FILENAME doesn't exist, then we can't support the --force-unix-socket option.  Writing out that file and restarting docker will not have any effect on Docker if it isn't set to look for its config in that file!

We should just error out if we can't find the file and they've used --force-unix-socket

grkvlt:
I believe that CentOS and RHEL use `/etc/sysconfig/docker` to store the startup args for Docker, is this checked for as well?

tomdee:
I think the code here has moved on and this PR can be closed. @Symmetric do you agree?

Symmetric:
Yep, your latest enhancements cover the same ground more thoroughly.

PR #64, opened 2015-04-14
Add support for configuring a RR

tomdee:
- Bump confd version to 0.9.0
- If /calico/config/bgp_peers exists then peer with values in that directory
  - Otherwise, keep old behavior
- The key is /calico/config/bgp_peers_ipv6 for ipv6

Set values using curl -X POST http://localhost:2379/v2/keys/calico/config/bgp_peers -d value=1.2.3.4


node/templates/bird6.cfg.template
```diff
@@ -30,7 +31,28 @@ protocol direct {
...
+  description "Connection to BGP peer";
+  local as 64511;
+  neighbor {{.Value}} as 64511;
+  multihop;
+  gateway recursive; # This should be the default, but just in case.
+  import filter calico_pools;
+  export filter calico_pools;
+  next hop self;    # Disable next hop processing and always advertise our
+                    # local address as nexthop
+  source address {{getenv "IP"}};  # The local address we use for the TCP connection
```
spikecurtis: s/IP/IP6

spikecurtis:
One typo, which will break bird6, but otherwise looks good.

PR #68, opened 2015-04-15
Unix socket cleanup

tomdee:


calicoctl.py
```diff
@@ -291,59 +302,23 @@ def node(ip, force_unix_socket, node_image, ip6=""):
...
-            clean_restart_docker(REAL_SOCK)
-
-        # Always remove the socket that powerstrip will use, as it gets upset
-        # otherwise.
-        if os.path.exists(POWERSTRIP_SOCK):
-            os.remove(POWERSTRIP_SOCK)
-
-        # At this point, docker is listening on a new port but powerstrip isn't
-        # running, so docker clients need to talk directly to docker.
+    if DOCKER_OPTIONS in open(DOCKER_DEFAULT_FILENAME).read():
```
spikecurtis: need to surround this in a `try:` block or it will fail on systems where `DOCKER_DEFAULT_FILENAME` doesn't exist.

calicoctl.py
```diff
@@ -783,6 +759,47 @@ def ip_pool_show(version):
...
 
+def restart_docker_with_alternative_unix_socket():
+    # Update docker to use a different unix socket, so powerstrip can run
+    # its proxy on the "normal" one. This provides simple access for
+    # existing tools to the powerstrip proxy.
+
+    # Set the docker daemon to listen on the docker.real.sock by updating
+    # the config, clearing old sockets and restarting.
+    socket_config_exists = \
+        DOCKER_OPTIONS in open(DOCKER_DEFAULT_FILENAME).read()
```
spikecurtis: Should also put this in a `try:` block with a helpful error message if `DOCKER_DEFAULT_FILENAME` doesn't exist.

tests/unix_socket.sh
```diff
@@ -2,8 +2,11 @@
 set -e
 set -x
 
+dist/calicoctl restart-docker-without-alternative-unix-socket
+
 # Set it up
 docker rm -f etcd || true
+#DOCKER_HOST="unix:///var/run/docker.real.sock" docker rm -f etcd || true
```
spikecurtis: leftover from testing?

spikecurtis:
A couple places we need to be defensive about the file not existing.  Other than that, looks good!

PR #79, opened 2015-04-16
Get unix socket stuff working on redhat

tomdee:
Tested on Centos 7.0 only (so far)

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -66,6 +66,7 @@ DOCKER_VERSION = "1.16"
 docker_client = docker.Client(version=DOCKER_VERSION,
                               base_url=os.getenv("DOCKER_HOST",
                                                  "unix://var/run/docker.sock"))
+docker_restarter = docker_restart.create_restarter()
 
 try:
     modprobe = sh.Command._create('modprobe')
@@ -288,7 +289,7 @@ def node(ip, node_image, ip6=""):
     node_docker_client = docker_client
 
     enable_socket = "NO"
-    if docker_restart.is_using_alternative_socket():
+    if docker_restarter.is_using_alternative_socket():
         enable_socket = "YES"
         # At this point, docker is listening on a new port but powerstrip
         # might not be running, so docker clients need to talk directly to
@@ -746,14 +747,14 @@ def restart_docker_with_alternative_unix_socket():
     Set the docker daemon to listen on the docker.real.sock by updating
     the config, clearing old sockets and restarting.
     """
-    docker_restart.restart_docker_with_alternative_unix_socket()
+    docker_restarter.restart_docker_with_alternative_unix_socket()
 
 
 def restart_docker_without_alternative_unix_socket():
     """
     Remove any "alternative" unix socket config.
     """
-    docker_restart.restart_docker_without_alternative_unix_socket()
+    docker_restarter.restart_docker_without_alternative_unix_socket()
 
 
 def validate_arguments():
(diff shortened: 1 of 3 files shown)


tomdee:
This could be better at telling the user what's going on, but I really want to get it in today...

node/adapter/docker_restart.py
```diff
@@ -0,0 +1,125 @@
...
+POWERSTRIP_SOCK = "/var/run/docker.sock"
+
+
+def _replace_all(filename, search, replace):
+    for line in fileinput.input(filename, inplace=1):
+        if search in line:
+            line = line.replace(search, replace)
+        sys.stdout.write(line)
+
+def create_restarter():
```
spikecurtis: PEP8 2 newlines for top-level methods

node/adapter/docker_restart.py
```diff
@@ -0,0 +1,125 @@
...
+                       "-H unix://%s \\" % REAL_SOCK
+
+    def _clean_restart_docker(self, sock_to_wait_on):
+        _clean_socks()
+        systemctl = sh.Command._create("systemctl")
+        systemctl("daemon-reload")
+        systemctl("restart", "docker.service")
+
+        # Wait for docker to create the socket
+        while not os.path.exists(sock_to_wait_on):
```
spikecurtis: We probably want to timeout with an error message if we wait too long.  30 seconds?

node/adapter/docker_restart.py
```diff
@@ -0,0 +1,125 @@
...
+
+def _clean_socks():
+    if os.path.exists(REAL_SOCK):
+        os.remove(REAL_SOCK)
+    if os.path.exists(POWERSTRIP_SOCK):
+        os.remove(POWERSTRIP_SOCK)
+
+
+class SystemdRestarter():
+    def __init__(self):
```
spikecurtis: I don't think you need to include an empty constructor.

node/adapter/docker_restart.py
```diff
@@ -0,0 +1,125 @@
...
+    def restart_docker_without_alternative_unix_socket(self):
+        if self.is_using_alternative_socket():
+            _replace_all(self.DOCKER_SYSTEMD_SERVICE,
+                         self.SYSTEMD_MODIFIED,
+                         self.SYSTEMD_DEFAULT)
+            self._clean_restart_docker(POWERSTRIP_SOCK)
+
+
+class UpstartRestarter():
+    def __init__(self):
```
spikecurtis: You don't need to include an empty constructor.

calicoctl.py
```diff
@@ -307,19 +289,13 @@ def node(ip, node_image, ip6=""):
...
-            # At this point, docker is listening on a new port but powerstrip
-            # might not be running, so docker clients need to talk directly to
-            # docker.
-            node_docker_client = docker.Client(version=DOCKER_VERSION,
-                                               base_url="unix://%s" % REAL_SOCK)
-    except IOError:
-        # Can't find Docker config file.  Don't attempt to change anything.
-        # Just carry on.
-        pass
+    if docker_restarter.is_using_alternative_socket():
```
spikecurtis: This code would be more clear if you moves the two above lines into an `else:` branch of this `if:`

node/adapter/docker_restart.py
```diff
@@ -0,0 +1,125 @@
+import os
+import time
+import sh
+import fileinput
+import sys
+REAL_SOCK = "/var/run/docker.real.sock"
+POWERSTRIP_SOCK = "/var/run/docker.sock"
+
```
spikecurtis: This module would really benefit from a block comment explaining the Unix socket business and why we're going to all this trouble.

PR #80, opened 2015-04-17
Fix import policy in bird to allow all routes.

Symmetric:
Previously we applied the same filter to import and export policy,
so only prefixes specified in etcd would be imported.

Relax this restriction, under the assumption that the RR/ToR
will be configured to only export sensible routes to us.


Symmetric:
Reviewed by Spike

PR #83, opened 2015-04-17
Upgrade Bird to 1.4.5, and enable BGP add-path.

Symmetric:
Add-path is required for Bird to advertise ECMP IPs. If it's not
enabled, Bird will refuse to advertise its route to the ECMP
IP if there's another copy of that route already advertised by
another node.


spikecurtis:
LGTM

PR #84, opened 2015-04-17
Refactor bird config files to use bgp template.

Symmetric:
Extract common config into a template, to reduce duplication
in the bird and bird6 config files.

This change should just be structural, and have no effect on the config that's output by confd.


Symmetric:
@spikecurtis , a quick review here -- this is a low priority refactor. I've tested against the /tests/mainline.sh and it looks kosher.

Symmetric:
Closing to re-target on master.

PR #88, opened 2015-04-20
Allow an optional interface name to be passed to calicoctl container add

### LINKED ISSUE #86: Add ability to specify container NIC interface name
When using `calicotl container add` it would be useful to be able to specify the interface name directly, since `eth1` may not be available.

tomdee:
Fixes #86

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -21,7 +21,7 @@ Usage:
   calicoctl (ipv4|ipv6) pool show
   calicoctl (ipv4|ipv6) bgppeer rr (add|remove) <IP>
   calicoctl (ipv4|ipv6) bgppeer rr show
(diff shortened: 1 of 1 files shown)


grkvlt:
@tomdee thanks, this looks like what I need

calicoctl.py
```diff
@@ -174,10 +174,13 @@ def container_add(container_name, ip):
         print "IP address is already assigned in pool %s " % pool
         sys.exit(1)
 
-    # Actually configure the netns.  Use eth1 since eth0 is the docker bridge.
+    # Actually configure the netns. Default to eth1 since eth0 is the
+    # docker bridge.
+    if interface is None:
```
spikecurtis: I think this would be clearer for both devs and users if we used docopt to set the default.

tomdee: I strongly agree

PR #89, opened 2015-04-21
Truncate workload IDs

### LINKED ISSUE #86: Add ability to specify container NIC interface name

tomdee:
Fixes #86

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -595,7 +595,7 @@ def node_show(detailed):
                     for ep_id, endpoint in endpoints.iteritems():
                         x.add_row([host,
                                    container_type,
-                                   workload,
+                                   workload[:12],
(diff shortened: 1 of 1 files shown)


calicoctl.py
```diff
@@ -595,7 +595,7 @@ def node_show(detailed):
                     for ep_id, endpoint in endpoints.iteritems():
                         x.add_row([host,
                                    container_type,
-                                   workload,
+                                   workload[:12],
```
Symmetric: I'd prefer a static var here, or even just a comment explaining where 12 comes from. Not immediately obvious from looking at the code (though I believe it's what docker does?)

PR #97, opened 2015-04-28
Multiple ips - Ready for review

tomdee:
I've added the following option to calicoctl
`calicoctl (ipv4|ipv6) container <CONTAINER> (add|remove) <IP> [--interface=<INTERFACE>]`

I didn't add a "show" command as the IPs can already been seen in the shownodes command.

It's not super urgent that I get this in, but I don't want the branch languishing.

I will do some more manual testing after code review, but it's working fine in my initial testing.


README.md
```diff
@@ -41,33 +41,37 @@ The `calico-node` service is a worker that configures the network endpoints for
...
-  calicoctl ipv4 pool show
-  calicoctl ipv6 pool add <CIDR>
-  calicoctl ipv6 pool del <CIDR>
-  calicoctl ipv6 pool show
-  calicoctl container add <CONTAINER> <IP>
+  calicoctl (ipv4|ipv6) pool (add|remove) <CIDR>
+  calicoctl (ipv4|ipv6) pool show
+  calicoctl (ipv4|ipv6) bgppeer rr (add|remove) <IP>
+  calicoctl (ipv4|ipv6) bgppeer rr show
+  calicoctl (ipv4|ipv6) container <CONTAINER> (add|remove) <IP> [--interface=<INTERFACE>]
```
spikecurtis: I feel like for this command we can drop the (ipv4|ipv6) option and infer it from the address itself.

Actually, this is probably true for all the commands.

calicoctl.py
```diff
@@ -793,16 +785,28 @@ def bgppeer_add(ip, version):
     :param version: v4 or v6
     :return: None
     """
+    address = check_ip_version(ip, version, IPAddress)
+    client.add_bgp_peer(version, address)
+
+
+def check_ip_version(ip, version, type):
```
spikecurtis: Please rename `type` to `cls`, which is the Python convention for that sort of thing.  `type` is a built-in function (and `class` is a reserved word), so it's easy to misread this code and it makes it less portable.

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
         print "No BGP Peers defined."
 
 
+def container_ip_add(container_name, ip, version, interface):
+    """
+    Add an IP address to an existing calico networked container.
```
spikecurtis: Nit: capitalize "Calico"

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
...
+
+    :param container_name: The name of the container.
+    :param ip: The IP to add
+    :param version: The IP version
```
spikecurtis: comment that version is in ("v4", "v6")

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
...
+    if not info["State"]["Running"]:
+        print "%s is not currently running." % container_name
+        sys.exit(1)
+
+    # Check that the container is already networked
+    try:
+        endpoint_id = client.get_ep_id_from_cont(hostname, container_id)
+        endpoint = client.get_endpoint(hostname, container_id, endpoint_id)
+    except KeyError:
+        print "Container is unknown to calico. Tell calico about the " \
```
spikecurtis: Nit: capitalize "Calico"

spikecurtis: It would be helpful to say which command to use to correct the error

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
...
+        sys.exit(1)
+
+    # From here, this method starts having side effects. If something
+    # fails then at least try to leave the system in a clean state.
+    if not client.assign_address(pool, ip):
+        print "IP address is already assigned in pool %s " % pool
+        sys.exit(1)
+
+    try:
+        old_endpoint = copy.deepcopy(endpoint)
```
spikecurtis: Refactor this to be a method on Endpoint --- e.g. `endpoint.copy()`  that saves the user from having to figure out what counts as a safe copy.

node/adapter/datastore.py
```diff
@@ -606,6 +606,24 @@ def set_endpoint(self, hostname, container_id, endpoint):
                                    "endpoint_id": endpoint.ep_id}
         self.etcd_client.write(ep_path, endpoint.to_json())
 
+    def update_endpoint(self, hostname, container_id,
+                        old_endpoint, new_endpoint):
+        """
+        Update a single endpoint object to the datastore. Fails if the
+        old_endpoint that's passed in doesn't match what's in the datastore.
```
spikecurtis: It would be helpful to include an example here, e.g.

```
old_endpoint = datastore.get_endpoint(...)
new_endpoint = old_endpoint.copy()
# modify new endpoint fields
datastore.update_endpoint(..., old_endpoint, new_endpoint)
```

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
...
+
+    try:
+        container_pid = info["State"]["Pid"]
+        netns.ensure_namespace_named(container_pid)
+        netns.add_ip_to_interface(container_pid,
+                                  address,
+                                  interface)
+
+    except CalledProcessError:
+        print "Error updating networking in container. Aborting."
```
spikecurtis: If you hit this condition, you've left the address on the endpoint, but unassigned it from the pool, so it could be auto-assigned to some other container, which is bad news.

I think you need to also clean up the endpoint in the datastore here.

calicoctl.py
```diff
@@ -843,6 +839,135 @@ def bgppeer_show(version):
...
+        sys.exit(1)
+
+    try:
+        old_endpoint = copy.deepcopy(endpoint)
+        if address.version == 4:
+            endpoint.ipv4_nets.remove(IPNetwork(address))
+        else:
+            endpoint.ipv6_nets.remove(IPNetwork(address))
+        client.update_endpoint(hostname, container_id, old_endpoint, endpoint)
+    except (KeyError, ValueError) as e:
```
spikecurtis: I think this is slightly too generic in the case where you attempt to remove an IP address that wasn't added to the container in the first place.  In that case, we should have an error message that says the IP address wasn't found on the container.

README.md
tomdee: Raised https://github.com/Metaswitch/calico-docker/issues/102

PR #98, opened 2015-04-28
Consolidate scripts into own directory.

alexaltair:


tomdee:
I think FV would be a better name

tomdee:
LGTM - merge away.

PR #103, opened 2015-04-29
Replace deprecated .execute with .exec_create & .exec_start.

alexaltair:


calicoctl.py
```diff
@@ -408,24 +408,25 @@ def status():
...
+        apt_output = docker_client.exec_create("calico-node", ["/bin/bash", "-c",
                                            "apt-cache policy calico-felix"])
-        result = re.search(r"Installed: (.*?)\s", apt_output)
+        result = re.search(r"Installed: (.*?)\s", docker_client.exec_start(apt_output))
         if result is not None:
             print "Running felix version %s" % result.group(1)
 
         print "IPv4 Bird (BGP) status"
-        print(docker_client.execute("calico-node",
+        bird_output = docker_client.exec_create("calico-node",
```
tomdee: Is bird_output the right name? Isn't this a docker execute command object, rather than some output?

calicoctl.py
```diff
@@ -408,24 +408,25 @@ def status():
     else:
         print "calico-node container is running. Status: %s" % \
               calico_node_info[0]["Status"]
-
-        apt_output = docker_client.execute("calico-node", ["/bin/bash", "-c",
+        apt_output = docker_client.exec_create("calico-node", ["/bin/bash", "-c",
```
tomdee: Same comment as for bird_output?

tomdee:
Looks fine other than the naming.

calicoctl.py
alexaltair: How about `bird_cmd` etc.?

tomdee: Yep, that would work.

PR #108, opened 2015-05-05
Show calico ip address info in docker inspect

robbrockbank:


tests/fv/auto_ip.sh
```diff
@@ -0,0 +1,77 @@
...
+node1_ip="$(docker inspect node1 | grep IPAddress)"
+node1_ip="${node1_ip#*: \"}"
+node1_ip="${node1_ip%\"*}"
+echo "Node 1 IP address is $node1_ip"
+
+node2_ip="$(docker inspect node2 | grep IPAddress)"
+node2_ip="${node2_ip#*: \"}"
+node2_ip="${node2_ip%\"*}"
+echo "Node 2 IP address is $node2_ip"
+
```
tomdee: I don't think the test beyond here is useful - it's exactly the same as mainline.sh, so just get rid.

You _could_ use a mix of manually assigned addresses instead of just auto, then grep for the known IP and check the return code. But maybe just add a comment to that effect and we'll do the proper checks in the python rewrite.

tomdee:
Looks good, just one comment on the tests.

Have you added the new test to Jenkins?
And don't forget to reference the issue when you merge the PR.

robbrockbank:
I've now moved the auto_ip test into the mainline (since there was so much shared code).

Issue #96, opened 2015-04-28
calicoctl doesn't give clean error messages.


alexaltair:
While trying out the calico-docker demo, I got the following stack traces from the calicoctl executable.
```
core@core-01 ~ $ sudo ./calicoctl node --ip=172.17.8.101    
Traceback (most recent call last):
  File "<string>", line 820, in <module>
  File "<string>", line 286, in node
  File "/code/build/calicoctl/out00-PYZ.pyz/node.adapter.datastore", line 303, in add_ip_pool
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 269, in write
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 582, in api_execute
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 600, in _handle_server_response
etcd.EtcdException: raft.Server: Not current leader
```
```
core@core-01 ~ $ sudo ./calicoctl node --ip=172.17.8.101    
Traceback (most recent call last):
  File "<string>", line 820, in <module>
  File "<string>", line 286, in node
  File "/code/build/calicoctl/out00-PYZ.pyz/node.adapter.datastore", line 303, in add_ip_pool
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 269, in write
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 582, in api_execute
  File "/code/build/calicoctl/out00-PYZ.pyz/etcd.client", line 600, in _handle_server_response
etcd.EtcdException: raft.Server: Not current leader
```
This was useful for me, but it's probably not what we want our users seeing. Should we write generally cleaner-looking error messages?

robbrockbank:
Fixed in pull request #113, pending review.

Fix handles:
-  etcd errors (prints generic is etcd running message)
-  prints error if container not in calico network with details of why and how to fix
-  prints error if docker command needs sudo access

robbrockbank:
Merged in changes from #113.

Closing issue.

alexaltair:
Another example;

```bash
$ sudo calicoctl reset
Removing all data from data store
$ sudo calicoctl pool add 192.168.0.0/16
$ sudo calicoctl container add container1 192.168.1.4
Traceback (most recent call last):
  File "<string>", line 1130, in <module>
  File "<string>", line 185, in container_add
[...]

PR #113, opened 2015-05-06
Issue 96: Improve error msgs for calicoctl command failures

robbrockbank:
This pull request is a fix for issue #96 
See issue for details of fix.

### DIFF
--- a/calicoctl.py
+++ b/calicoctl.py
@@ -55,14 +55,14 @@ from docopt import docopt
 import sh
 import docker
 import docker.utils
-from etcd import EtcdException
 from netaddr import IPNetwork, IPAddress
 from netaddr.core import AddrFormatError
 from prettytable import PrettyTable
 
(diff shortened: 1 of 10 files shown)


tomdee:
We don't want details of etcd leaking into calicoctl. I think you should define new exceptions in datastore.py and catch those in calicoctl.

At some point we want to move docker bits out of calicoctl into a separate module too, but not just yet.

tomdee:
LGTM other than the one comment above. I'd prefer not to be adding more docker specific code to calicoctl.py but it's OK for now. The etcd point still stands though,

robbrockbank:
Added a generic DatastoreError that is currently just raised after catching an EtcdException (with the message passed through).  Could no doubt do better here though.

Use a decorator to decorate all datastore public methods to handle general exception catching.

tomdee:
That's a lot of decorators. Not sure if this is really the right approach but I think it's fine to merge for now.

Issue #87, opened 2015-04-20
How to add multiple IPs per container?


grkvlt:
This is mentioned in the README, but I copuld not find documentation on the lower level API to use?

> Using calicoctl we currently only support one IP address per container, but more than one is possible if you use the lower level APIs.

What should I be doing here? I would actually like to be able to run `calicoctl container add` multiple times with different interface names and IP addresses if possible in future, but until then I am happy to use a different mechanism from [Clocker](http://github.com/brooklyncentral/clocker/) instead.

tomdee:
Do you really need to add multiple interfaces? A safer approach would be to just add multiple IPs to the existing interface.

To add multiple IPs (with the current code) you would need to add the IPs to the ipv(4|6)_nets array in the endpoint info - https://github.com/Metaswitch/calico-docker/blob/master/docs/etcdStructure.md#json-endpoint-config

And then manually adding the IPs to the interfaces (ip addr add ... in the container)

I would propose wrapping all this up in some calicoctl commands, e.g.
* calicoctl container address show 
* calicoctl container address add <IP>
* calicoctl container address remove <IP>

If you're happy with just adding multiple IPs to a single interface then I can look at adding the calicoctl commands for this in the next few days.

tomdee:
Work in progress code for this is on the multiple-ips branch

grkvlt:
@tomdee thanks, this seems like a reasonable approcach. I have provisionally updated Clocker to add mutiple IPs to the `eth1` interface based in the container now, but haven't touched the JSON data in etcd, I will wait until you have merged the `calicoctl` commands.

tomdee:
Fixed in https://github.com/Metaswitch/calico-docker/commit/7f59c8784e7cd720c4fca8e4899e01f65d21c7c0 and in the new release (0.4.0

frostynova:
i tried the new cmd "calicoctl ipv4 container test_ct add 192.168.77.30", it said:
```
ln: failed to create symbolic link ‘/var/run/netns/5742’: No such file or directory
```

tomdee:
@robbrockbank Could you try to repro (and fix) this?

robbrockbank:
@frostynova I've been trying to reproduce this locally without much success.

Do you happen to know what version of calico-docker you are running and also what commands you ran prior to error?

robbrockbank:
Fixed in #114

PR #114, opened 2015-05-06
Use nsenter as it is more generic and faster than docker exec

tomdee:
Also introduces some test refactoring,

Currently breaks on circle CI. We should run the tests in a container so it doesn't break...


robbrockbank:
Tom - largely the changes you made, but with some further cutting out of unused function and some general fix up of errors.

tomdee:
Looks good to me.

robbrockbank:
Fixes #87

PR #119, opened 2015-05-13
Run the UTs in their own container.

alexaltair:


Dockerfile
```diff
@@ -36,11 +36,14 @@ RUN add-apt-repository -y ppa:cz.nic-labs/bird && \
         python-pip \
         python-pyasn1 \
         python-netaddr \
+        python-nose \
```
tomdee: Why are you installing nose into the calico-node container? You shouldn't be doing it without a really good reason...

Dockerfile
```diff
@@ -36,11 +36,14 @@ RUN add-apt-repository -y ppa:cz.nic-labs/bird && \
...
+        software-properties-common \
```
tomdee: What's this needed for?

build_calicoctl/create_binary.sh
```diff
@@ -1,14 +1,32 @@
...
 
 docker build -t calico-build .
 
 cd ..
 mkdir -p `pwd`/dist
 chmod 777 `pwd`/dist
-docker run -u user -v `pwd`/:/code/calico -v `pwd`/dist:/code/dist calico-build pyinstaller calico/calicoctl.py -a -F -s --clean
-docker run -u user -v `pwd`/dist:/code/dist calico-build bash -c 'docopt-completion --manual-bash dist/calicoctl && mv calicoctl.sh dist'
+
+if [[ $CIRCLE_TEST_REPORTS ]]; then
```
tomdee: Hmmm, I don't like this approach. The circle specific parts should go in the circle.yaml

build_calicoctl/create_binary.sh
```diff
@@ -1,14 +1,32 @@
...
+    docker run -v `pwd`/:/code -v $CIRCLE_TEST_REPORTS:/circle_output calico-build \
+     bash -c '/tmp/etcd & \
+      nosetests -c nose.cfg --cover-html-dir=dist --with-xunit --xunit-file=/circle_output/output.xml'
+else
+    docker run --rm -v `pwd`/:/code calico-build \
+     bash -c '/tmp/etcd & \
+      nosetests -c nose.cfg$NOSE_ARGS'
```
tomdee: What is $NOSE_ARGS?

Dockerfile
alexaltair: I thought you'd just wanted to remove the whole RUN apt-get command mess from calico-build, and were willing to take the hit of it being slightly weird to install that stuff on calico/node.

build_calicoctl/requirements.txt
```diff
@@ -1,20 +1,11 @@
-pycparser==2.10
 coverage==3.7.1
 infi.docopt-completion==0.2.6
 docopt==0.6.2
-sh==1.11
-requests==2.5.1
 simplejson==3.6.5
```
tomdee: I think this could be a calicoctl.py thing too. Do you know how to check?

Dockerfile
alexaltair: I think it's for `ca-certificates`.

node/adapter/requirements.txt
```diff
@@ -15,6 +15,6 @@ pyOpenSSL==0.15.1
 pyasn1==0.1.7
 pyasn1-modules==0.0.5
 six==1.9.0
-urllib3==1.10.3
+urllib3>=1.7,<=1.8
```
tomdee: Interesting! Surely there is one right version?

build_calicoctl/create_binary.sh
alexaltair: Yeah, I don't like it either. I'd like to be able to run the UTs locally. So I need a way to say "If you're running this on circleci, go ahead and pass the coverage args to nose. Otherwise don't try." I could do that by putting the call to nose back in circle.yml, but I thought we wanted to fold the UTs into the build pipeline.

build_calicoctl/create_binary.sh
alexaltair: Oops.

build_calicoctl/requirements.txt
alexaltair: I grepped for `simplejson` and didn't see it.

node/adapter/requirements.txt
alexaltair: Shaun has Opinions about this.

Dockerfile
alexaltair: Fixed.

build_calicoctl/create_binary.sh
tomdee: You can make this script more modular. Don't make it run the UTs, just build the image and the calicoctl binary.

Then add the line to call the UTs to circleci.yaml
And optionally add another script to the root direcotry call something like run_uts.sh which just runs the UTs.

build_calicoctl/requirements.txt
tomdee: OK, so can we remove it?

node/adapter/requirements.txt
tomdee: Can you write those opinions here please?

build_calicoctl/Dockerfile
```diff
@@ -1,18 +1,13 @@
...
-        python \
-        python-pip \
-        libffi-dev \
-        libssl-dev \
-        libyaml-dev \
-        python-dev \
-        python-etcd \
-        git && \
-    apt-get clean
+        python-nose \
```
tomdee: Why are you installing this? Isn't nose in the requirements.txt?

build_calicoctl/Dockerfile
```diff
@@ -1,18 +1,13 @@
...
+        software-properties-common \
```
tomdee: Why are you installing this. What is it actually needed for?

node/adapter/requirements.txt
fasaxc: Should be `>=1.7,<1.8`  i.e. we want the newest 1.7.x release.  1.8 added new exceptions that python-etcd doesn't seem to be expecting.  1.9 removed those exceptions and added different ones.

build_calicoctl/create_binary.sh
alexaltair: Is there a reason `create_binary.sh` is inside `build_calicoctl`?

build_calicoctl/requirements.txt
alexaltair: Looking slightly harder; it looks like a soft dependency for `requests`.

node/adapter/requirements.txt
```diff
@@ -15,6 +15,7 @@ pyOpenSSL==0.15.1
 pyasn1==0.1.7
 pyasn1-modules==0.0.5
 six==1.9.0
-urllib3==1.10.3
+urllib3>=1.7,<1.8
```
tomdee: I still don't understand why this isn't an exact version?

run_uts.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -x
+set -e
+date
```
tomdee: You don't need date, pwd and git status here. Remove them.

run_uts.sh
```diff
@@ -0,0 +1,15 @@
...
+pwd
+git status
+
+./create_binary.sh
+docker stop pyinstaller docopt
+docker rm pyinstaller docopt
+docker run -rm -v `pwd`/:/code calico-build bash -c '/tmp/etcd & nosetests -c nose.cfg'
+
+# BE CAREFUL
+rm -rf .coverage cover/ default.etcd/
```
tomdee: What is this for?

run_uts.sh
```diff
@@ -0,0 +1,15 @@
...
+date
+pwd
+git status
+
+./create_binary.sh
+docker stop pyinstaller docopt
+docker rm pyinstaller docopt
+docker run -rm -v `pwd`/:/code calico-build bash -c '/tmp/etcd & nosetests -c nose.cfg'
+
+# BE CAREFUL
```
tomdee: This comment is useless

create_binary.sh
```diff
@@ -0,0 +1,23 @@
...
+set -x
+
+pushd build_calicoctl
+docker build -t calico-build .
+popd
+
+mkdir -p `pwd`/dist
+chmod 777 `pwd`/dist
+
+docker run -v `pwd`/:/code --name pyinstaller calico-build \
```
tomdee: Why are you naming these?
And why have you removed --rm?

run_uts.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -x
+set -e
+date
+pwd
+git status
+
+./create_binary.sh
```
tomdee: Remove this call

run_uts.sh
```diff
@@ -0,0 +1,15 @@
...
+docker stop pyinstaller docopt
```
tomdee: Remove this and the next line

node/adapter/requirements.txt
alexaltair: Basically because I don't have an opinion either way. I can think of reasons for and against, but they don't move me so I just stick with the minimum modifications. But since you do have vocal opinions, I shall find the latest version in that range.

run_uts.sh
alexaltair: This is a convention I've been carrying over from a jenkins config. I quite like them to be honest; they're cheap, and can save some pain in the case where you need that info.

run_uts.sh
alexaltair: When I run this file locally, nose creates these dirs, I think because of what's in `nose.cfg`. They're useless in that case. I can't change `nose.cfg` because it's for circleci. I could make a `local_nose.cfg` but that seems excessive. I tried overriding it with arguments but didn't get it to work. And I'd really rather not type `rm -rf` manually a bunch of times.

run_uts.sh
alexaltair: Interesting assertion! I quite appreciate it when people conspicuously flag things that are potentially dangerous. This is to tell people that modifications or typos might not be harmless, and to think twice instead of once.

create_binary.sh
alexaltair: Circle CI fails if I try to remove a docker container, I think because it has docker-in-docker issues. The goal here is to not have docker containers accumulate on my dev machine or on jenkins. Naming them allows me to delete them without grepping shenanigans.

run_uts.sh
lxpollitt: I agree with Tom that this comment is far too cryptic to be useful.  Why is it cryptic at all?  If there is something important to say then say it explicitly rather than leaving as an exercise for the reader.

create_binary.sh
lxpollitt: If we want this code to be maintainable then it needs much better commenting.  If the code reviewer is having to ask these sort of questions then the code needs better commenting (or is wrong).

create_binary.sh
```diff
@@ -0,0 +1,25 @@
...
+ pyinstaller calicoctl.py -a -F -s --clean
+docker rm -f docopt || true
```
tomdee: You got you names the wrong way round with the rm commands

docs/Building.md
```diff
@@ -14,7 +14,7 @@ This builds the Dockerfile giving it the name `calico/node:latest`, which is the
 
 ## Building `calicoctl` CLI Tool - mainline
 
-    cd ./build_calicoctl
+    ./build_calicoctl
```
tomdee: What does this do? I don't think you want it

tomdee:
Two final comments. Once you've fixed them you should be good to merge.

create_binary.sh
alexaltair: Sounds reasonable. I'm still trying to calibrate to how Maine Coon is making the speed versus quality trade-off.

docs/Building.md
alexaltair: The problem with documentation is that it doesn't get run.

PR #124, opened 2015-05-14
Run FVs on multiple hosts.

alexaltair:


run_fvs.sh
```diff
@@ -6,17 +6,34 @@ date
 pwd
 git status
 
-./build_node.sh
-./create_binary.sh
+docker rm -f host1 || true
+docker run --privileged -v `pwd`:/code --name host1 -tid jpetazzo/dind
+
+docker exec -t host1 bash -c \
+ 'docker -d & \
```
tomdee: Are you _sure_ that this docker call is required?

run_fvs.sh
```diff
@@ -6,17 +6,34 @@ date
...
+# docker run --privileged -v `pwd`:/code --name host2 -t jpetazzo/dind bash -c \
+#  'docker -d -D && \
+#  cd /code && \
+#  ./build_node.sh && \
+#  ./create_binary.sh'
+# docker run --privileged -v `pwd`:/code --name host3 -t jpetazzo/dind bash -c \
+#  'docker -d -D && \
+#  cd /code && \
+#  ./build_node.sh && \
+#  ./create_binary.sh'
```
tomdee: I don't really like checking it commented out lines, but I don't feel too strongly about it.

run_fvs.sh
```diff
@@ -6,17 +6,34 @@ date
...
-sudo ./tests/fv/diags.sh
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/mainline.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/add_container.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/add_ip.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/arg_parsing.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/profile_commands.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/no_powerstrip.sh'
+docker exec -t host1 bash -c 'cd /code && sudo ./tests/fv/diags.sh'
+
+docker rm -f host1 || true
```
tomdee: I think you should take this line out

run_fvs.sh
alexaltair: Yep. If I take out only that, it says;
`FATA[0000] Post http:///var/run/docker.sock/v1.18/build?cpusetcpus=&cpushares=0&dockerfile=Dockerfile&memory=0&memswap=0&rm=1&t=calico%2Fnode: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?`

run_fvs.sh
alexaltair: Why's that, given that this script is for running on jenkins, and not for developer convenience?

tomdee: How is a developer meant to run the (multi host) tests?

run_fvs.sh
tomdee: Hmmm, the dind container will already be running a docker daemon so it's a bit strange to run another. I suspect that by running another daemon you're tramping over the the one that's run for you. 

Something like `while ! docker ps; do sleep 1; done` would capture the meaning of 'waiting for docker to start' better than just starting a new daemon.

run_fvs.sh
alexaltair: Locally? Just by running `./run_fvs.sh`. I suspect this will change quickly, though. Running multi-host tests takes *forever*, so we'll probably want to have some kind of flag.

run_fvs.sh
alexaltair: Ahh, okay. That explains the data.

tomdee:
Code looks good to me. Please don't merge till you've fixed the build though.

PR #128, opened 2015-05-15
Add ability to reinstate networking to container

robbrockbank:
This doesn't work yet as it uncovers a bug in Felix where Felix does not update the ARP cache for a MAC change.

See: https://github.com/Metaswitch/calico/issues/549

We could work around it by deleting the endpoint and creating a new endpoint - but I don't like that as it's not an atomic operation.  That said, it would be a simple temporary solution.

### DIFF
--- a/node/adapter/datastore.py
+++ b/node/adapter/datastore.py
@@ -31,6 +31,10 @@ prefix that appears in all Calico interface names in the root namespace. e.g.
 cali123456789ab.
 """
 
+VETH_NAME = "eth1"
+"""The name to give to the veth in the target container's namespace. Default
+to eth1 because eth0 could be in use"""
+
 
 def handle_errors(fn):
     """
@@ -174,10 +178,17 @@ class Endpoint(object):
     @classmethod
     def from_json(cls, ep_id, json_str):
         json_dict = json.loads(json_str)
+
+        # If there is no container if_name sepcified, assume the default
+        # VETH_NAME.  For containers created prior to this information being
+        # stored, it will be possible to restart the containers, but the
+        # interface may be named differently.
+        if_name = json_dict.get("container:if_name", VETH_NAME)
+
         ep = cls(ep_id=ep_id,
                  state=json_dict["state"],
                  mac=json_dict["mac"],
-                 if_name=json_dict["container:if_name"])
+                 if_name= if_name)
         for net in json_dict["ipv4_nets"]:
             ep.ipv4_nets.add(IPNetwork(net))
         for net in json_dict["ipv6_nets"]:
--- a/node/adapter/datastore.py
+++ b/node/adapter/datastore.py
@@ -736,7 +736,7 @@ class DatastoreClient(object):
     @handle_errors
     def get_endpoints(self, hostname, container_id):
         """
-        Get all of Endpoints for a container.
+        Get all of the Endpoints for a container.
 
         :param hostname: The hostname that the endpoint lives on.
         :param container_id: The container that the endpoint belongs to.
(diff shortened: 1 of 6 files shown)


robbrockbank:
I still need to add UTs and fix up existing UTs.

spikecurtis:
We could also just hard-code the MAC address for every container (i.e. manually set the MAC on the veth pair when it comes up).

robbrockbank:
@spikecurtis Yup was pretty easy just to re-use the original MAC, so that is a much better workaround

tomdee:
Do you know why the jenkins build is failing?

node/adapter/datastore.py
```diff
@@ -160,6 +161,7 @@ def to_json(self):
         json_dict = {"state": self.state,
                      "name": IF_PREFIX + self.ep_id[:11],
                      "mac": self.mac,
+                     "container:if_name": self.if_name,
```
tomdee: You're confident that felix doesn't care about this extra data?

node/adapter/datastore.py
```diff
@@ -174,7 +176,8 @@ def from_json(cls, ep_id, json_str):
         json_dict = json.loads(json_str)
         ep = cls(ep_id=ep_id,
                  state=json_dict["state"],
-                 mac=json_dict["mac"])
+                 mac=json_dict["mac"],
+                 if_name=json_dict["container:if_name"])
```
tomdee: Have you thought about the upgrade path here? What happens someone with existing endpoints takes this code. Will it "just work" or will it break. If it breaks we should decide whether to fix it or just release note it.

node/adapter/netns.py
```diff
@@ -209,3 +218,36 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
         ep.ipv6_nets.add(network)
         ep.ipv6_gateway = next_hop
     return ep
+
+
+def reinstate_endpoint(cpid, old_endpoint, next_hop_ips,
+                       proc_alias=PROC_ALIAS):
+    """
+    Re-instate and endpoint that has been removed.
```
tomdee: typo

robbrockbank:
@tomdee The build seemed to fail because of a download timeout by the looks of it.  So it didn't get as far as actually running the FVs.

FVs pass locally on my box..  Not entirely sure how to retrigger a build from git -any ideas?

Regarding the extra data:  Shaun was under the impression that endpoint data should be fine (just prefix it with docker or container etc. to ensure they don't trample on the data in future).  He mentioned that Rule data was another thing and that you could not arbitrarily add data there.

node/adapter/netns.py
```diff
@@ -209,3 +218,36 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+    :param cpid: The PID and name of the namespace to operate in.
```
tomdee: It's not the name. It's just the PID.

node/adapter/datastore.py
robbrockbank: Good point.  I think making it just work is preferable and not much work.

node/adapter/netns.py
```diff
@@ -209,3 +218,36 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+    :param old_endpoint: The old endpoint that is being re-instated.
+    :param next_hop_ips: Dict of {version: IPAddress} for the next hops of the
+    default routes namespace.
+    :param proc_alias: The head of the /proc filesystem on the host.
```
tomdee: I'm not sure I understand this comment. Something more on the lines of "the location of the /proc filesystem"

PR #129, opened 2015-05-15
Misc Dockerfile cleanups

tomdee:
I've not tested this yet, so please only review it if the tests pass!

The only bit I'm worried about is removing  python-zmq (which should be fine).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,15 +4,6 @@ FROM phusion/baseimage:0.9.16
 # Use baseimage-docker's init system.
 CMD ["/sbin/my_init"]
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -64,3 +50,7 @@ RUN git clone https://www.github.com/Metaswitch/powerstrip.git && \
     cd powerstrip && \
     sed -i s/2375/2377/ powerstrip.tac && \
     python setup.py install
+
+# Copy in our custom configuration files etc. We do this last to speed up
+builds for developer, as it's thing they're most likely to change.
```
alexaltair: No comment here, and missing newline.

alexaltair:
LGTM.

PR #130, opened 2015-05-15
Remove containers within containers before exiting outside container.

alexaltair:


tomdee:
Code looks fine. Add some comments (and link to the right section of the DIND readme) then you're good to merge.

PR #132, opened 2015-05-18
Build containers outside & docker load into hosts.

alexaltair:


run_fvs.sh
```diff
@@ -12,13 +12,23 @@ git status
 docker exec -t host1 bash -c 'docker rm -f $(docker ps -qa) ; \
                               docker rmi $(docker images -qa)' || true
 docker rm -f host1 || true
+
+docker pull busybox:latest
+docker save --output busybox.tar busybox:latest
+./build_node.sh
+docker pull jpetazzo/nsenter
+docker save --output nsenter.tar jpetazzo/nsenter
+docker images | grep quay.io/coreos/etcd || docker pull quay.io/coreos/etcd:v2.0.10
```
tomdee: Why the special casing here?

tomdee: It at least needs a comment

tomdee:
I don't see where you're passing in the calico-node image?

run_fvs.sh
```diff
@@ -12,13 +12,23 @@ git status
 docker exec -t host1 bash -c 'docker rm -f $(docker ps -qa) ; \
                               docker rmi $(docker images -qa)' || true
 docker rm -f host1 || true
+
+docker pull busybox:latest
```
tomdee: It would be great if this new block of code doesn't incur extra run time if it doesn't need to.

The docker pull is pretty safe - it won't do anything (significant) if the image is already pulled
However, the save is a bit wasteful. It would be very easy to just have a simple if at the front to make it not do anything if the file exists already.
`if [ ! -e busybox.tar ] ; then docker save --output busybox.tar busybox:latest ; fi

tomdee:
The approach looks good but you need to comment your code. Add in some echo statements too so the user knows what's going on.

Address this and my other feedback above then let me take another look before you merge.

run_fvs.sh
alexaltair: While I was watching this run, it does actually sit at `docker pull` for several seconds, even if it eventually concludes that it doesn't need to pull the image. I used the `docker images | grep || docker pull` on the etcd one because it was pinned at a version, so presumably we wouldn't care about getting the latest. But if we don't care, I can use the `if [` on all of them.

tomdee:
OK, looks fine. Merge away.

PR #134, opened 2015-05-18
Do not import nsenter as it causes seg fault with certain varions of libc

robbrockbank:
Undo changes to add nsenter until we determine a different approach for library loading.

### DIFF
--- a/node/adapter/netns.py
+++ b/node/adapter/netns.py
@@ -224,7 +224,7 @@ class NamedNamespace(object):
 
     def __exit__(self, _type, _value, _traceback):
         try:
-            os.unlink(self.nsr_dir)
+            os.unlink(self.nsn_dir)
         except BaseException:
             _log.exception("Failed to remove link: %s", self.nsn_dir)
         return False
--- a/node/adapter/netns.py
+++ b/node/adapter/netns.py
@@ -24,7 +24,6 @@ import os.path
 from netaddr import IPNetwork, IPAddress
 
 from datastore import Endpoint, IF_PREFIX
-from nsenter import Namespace
 
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


node/adapter/netns.py
```diff
@@ -209,3 +197,52 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+        self.pid_dir = "%s/%s/ns/net" % (proc, cpid)
+        self.nsn_dir = "/var/run/netns/%s" % self.name
+
+    def __enter__(self):
+        """
+        Add the appropriate configuration to name the namespace.  This links
+        the PID to the namespace name.
+        """
+        _log.debug("Creating link between ns name and PID")
+        check_call("mkdir -p /var/run/netns", shell=True)
```
tomdee: Use python makedirs here

node/adapter/netns.py
```diff
@@ -209,3 +197,52 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+        check_call("ln -s %s %s" % (self.pid_dir, self.nsn_dir),
```
tomdee: ANd use python os.link here

node/adapter/netns.py
```diff
@@ -209,3 +197,52 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+                   shell=True)
+        _log.debug(check_output("ls -l /var/run/netns", shell=True))
```
tomdee: I don't think we need this line and it should just be removed

node/adapter/netns.py
```diff
@@ -209,3 +197,52 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
...
+        return self
+
+    def __exit__(self, _type, _value, _traceback):
+        try:
+            check_call("rm %s" % self.nsn_dir, shell=True)
```
tomdee: Use python os.delete to remove the dir here

tomdee:
A few comments to make the code more pythonic. Once you've resolved them I'm happy for you to merge.

I'm a bit concerned about all the file system interactions that we end up with here, but on balance I think it's OK.

PR #135, opened 2015-05-18
Provide some guidance on getting traffic to/from the calico cluster

tomdee:
@fasaxc Do these iptables commands look OK to you? Do you think they will interact with felix in bad ways?
@lxpollitt Can you review the text?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,3 +87,56 @@ Options:
 ## Can a guest container have multiple networked IP addresses?
 Yes, using the `calicoctl container <CONTAINER> ip (add|remove) <IP>` command.
 
+## How can I get network traffic into and out of my Calico cluster?
+The ideal way to get traffic to/from your Calico network is by peering to 
+your existing data center routers using BGP and by assigning globally 
+routable IPs to your workloads. This avoids the need for NATs, 
+tunnels or overlays. 
+
+Detailed datacenter networking recommendations are given in the main 
+[Project Calico documentation](http://docs.projectcalico.org/en/latest/index.html).
+
+In some cases however, it's not possible to set up such a peering. In such a
+ case, then iptables rules can be configured on the compute hosts to get 
+ traffic into and out of the Calico networked containers.
+
+### Egress traffic from containers
+If you just want to be able to access the internet from your containers and 
+they only have private Calico IP addresses then you can configure SNAT on 
+each compute host.
+```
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -87,3 +87,56 @@ Options:
...
+range (or multiple IP ranges) if you've configured different pools. 
+
+The command will need to be run each time the host is restarted.
+
+### Ingress traffic to containers
+if you're trying to host a public facing service on your Calico network then
+you'll want some way of exposing that service. It's highly desirable that 
+you assign a public IP to your container and peer with your border router 
+so that traffic for that IP is brought to the appropriate compute host. If 
+you can't do this and you're host already has a public facing IP, then you
```
fasaxc: s/you're/your/

README.md
```diff
@@ -87,3 +87,56 @@ Options:
...
+
+### Ingress traffic to containers
+if you're trying to host a public facing service on your Calico network then
+you'll want some way of exposing that service. It's highly desirable that 
+you assign a public IP to your container and peer with your border router 
+so that traffic for that IP is brought to the appropriate compute host. If 
+you can't do this and you're host already has a public facing IP, then you 
+can use port mapping to get the traffic into the container.
+
+You will need to runt he following commands to on the host that the container
```
fasaxc: typo: runt

README.md
```diff
@@ -87,3 +87,56 @@ Options:
...
+is running on.
+
+```
+iptables -A FORWARD -p tcp -d <CALICO_IP> --dport <SERVICE_PORT> -j ACCEPT
```
fasaxc: Why is this FORWARD rule needed; I think it'll just bypass all our security?  If you get the PREROUTING NAT rule right then it should get routed to the VM and accepted/rejected based on its profile rules.

fasaxc:
@tomdee instead of writing a doc about it, why not add the NAT feature to calicoctl?  Feels like only moderately more work, with a lot less room for user mistakes.
```
calicoctl expose port ...
```
or some such.

tomdee:
Adding to calicoctl would be sensible, I'll consider that for a future increment.

README.md
tomdee: Getting rid of this FORWARD would be very desirable. Can you see anything obviously wrong with the PREROUTING rule or do you think it has a chance of working without the FORWARD rule in place?

PR #138, opened 2015-05-19
Refactor node to package

Symmetric:
Move node/*.py to calico/*.py.
Move remaining node/ to node_fs/
Update Dockerfile


Symmetric:
@robbrockbank , can you give this a wee review please? If it looks good I'll give everyone a heads-up before merging.

robbrockbank:
A couple of questions:

-  I'd rename node_fs to be node_filesystem or similar as I think the fs is a little cryptic.
-  Can we put a README in the node_fs directory to indicate what it contains.
-  Does calicoctl.py really live within the calico_container codebase?  Seems to me that it is an application that builds on top of the calico_container library.  I'm not too sure about this though and i could go either way.
- Does the powerstrip code really belong where we've put it, or more correctly, does netns, ipam and datastore really belong in adapter?  Seems like the powerstrip should be in adapter and the other modules elsewhere.
-  Maybe a naive comment, but should we be thinking about more structure in the filesystem to handle different OSs?

Symmetric:
> I'd rename node_fs to be node_filesystem or similar as I think the fs is a little cryptic.

OK
> Can we put a README in the node_fs directory to indicate what it contains.

Good idea, though I don't think this PR should be the place to do all the documentation that's justified for the node_fs. I'll put a brief note in to get things going.
> Does calicoctl.py really live within the calico_container codebase? Seems to me that it is an application that builds on top of the calico_container library. I'm not too sure about this though and i could go either way.

Yes, because I'd like to be able to call calicoctl methods directly from plugins in the calico_container tree instead of having to shell out.
> Does the powerstrip code really belong where we've put it, or more correctly, does netns, ipam and datastore really belong in adapter? Seems like the powerstrip should be in adapter and the other modules elsewhere.

Probably the netns, ipam, etc. should be pulled out into their own module. I think that's a refactor step to be taken as soon as something else uses those modules though, we might never need to.
> Maybe a naive comment, but should we be thinking about more structure in the filesystem to handle different OSs?

I don't think so, we specify the OS in our Dockerfile and don't expect to run this code anywhere else.

tomdee:
If you're calling calicoctl methods directly, that implies we're putting too much logic in calicoctl. We should either be pushing that down into datastore.py/netns.py or adding another layer in between.

In principle, I agree with Rob, calicoctl uses this library and should be at the project root. I'm also keen to see this code merged so I'm not sure if you need to sort it out now.

Likewise, netns/ipam/datastore really shouldn't be under adapter. Even if you're rkt integration doesn't use those modules, libnetwork and mesos integrations will. i.e. we definitely need to pull those out of adapter, but maybe not now.

Symmetric:
Ok, here's another reason -- if we want to test calicoctl (we should
eventually), then it will be easier if it's in the tree.

And another reason -- if we ever move to setup.py (e.g. for specifying
dependencies more sensibly), then it's a pain to include calicoctl.py in
the setup.py install_requires.

I don't agree with putting python files in random locations in the project,
they should all be in a proper package. Gives us more flexibility with no
real cost.

Let's punt discussion of dissecting calicoctl, I'll make that part of the
design review for rkt (once I'm past the prototype phase and looking to get
something serious checked in).

Symmetric:
Ok, here's another reason -- if we want to test calicoctl (we should
eventually), then it will be easier if it's in the tree.

And another reason -- if we ever move to setup.py (e.g. for specifying
dependencies more sensibly), then it's a pain to include calicoctl.py in
the setup.py install_requires.

I don't agree with putting python files in random locations in the project,
they should all be in a proper package. Gives us more flexibility with no
real cost.

Let's punt discussion of dissecting calicoctl, I'll make that part of the
design review for rkt (once I'm past the prototype phase and looking to get
something serious checked in).

tomdee:
SGTM

robbrockbank:
yup good with me too

PR #139, opened 2015-05-19
Add multi-host test in python

alexaltair:
Add multi-host test in python.


tomdee:
@alexaltair I'm not going to be able to look at this today. Maybe @Symmetric or @robbrockbank can.

run_fvs.sh
```diff
@@ -6,39 +6,7 @@ date
...
-docker run --privileged -v `pwd`:/code --name host1 -tid jpetazzo/dind
-
-# Load each of the images saved above.
-docker exec -t host1 bash -c \
- 'while ! docker ps; do sleep 1; done && \
- docker load --input /code/calico-node.tar && \
- docker load --input /code/busybox.tar && \
- docker load --input /code/nsenter.tar && \
- docker load --input /code/etcd.tar'
+nosetests tests/fv --nocapture
```
robbrockbank: Worth adding a commend here that running the python FV tests also recreates the node and calicoctl binaries

tests/fv/__init__.py
```diff
@@ -0,0 +1,75 @@
+import os
+import sh
+from sh import docker
+
+
+def setup_package():
+    pwd = sh.pwd().stdout.rstrip()
+
+    docker_exec = docker.bake("exec")
+    host1_exec = docker_exec.bake("-t", "host1", "bash", "-c")
```
robbrockbank: Not sure I like this.

Better to query which docker in docker hosts there actually are, and then remove the inner containers for all of them.  Perhaps naming them something more uniquely identifiable will make this a safer operation, or perhaps querying the image type.

tests/fv/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        sh.docker.rm("-f", "host1", "host2")
+    except sh.ErrorReturnCode:
+        pass
+    print "Host containers removed."
+
+    # # Save and load each image, so we can use them in the inner host containers.
+    print sh.bash("./build_node.sh").stdout
+    docker.save("--output", "calico-node.tar", "calico/node")
+    if not os.path.isfile("busybox.tar"):
```
robbrockbank: doesn't this logic mean that we can get out of date with the latest busybox and nsenter?

tests/fv/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        docker.pull("busybox:latest")
+        docker.save("--output", "busybox.tar", "busybox:latest")
+    if not os.path.isfile("nsenter.tar"):
+        docker.pull("jpetazzo/nsenter:latest")
+        docker.save("--output", "nsenter.tar", "jpetazzo/nsenter:latest")
+    if not os.path.isfile("etcd.tar"):
+        docker.pull("quay.io/coreos/etcd:v2.0.10")
+        docker.save("--output", "etcd.tar", "quay.io/coreos/etcd:v2.0.10")
+
+    print sh.bash("./create_binary.sh")
```
robbrockbank: I don't really think we should use print here if it's reasonably easy to start using logging.

tests/fv/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+
+    print docker.run("--privileged", "-v", pwd+":/code", "--name", "host1", "-tid", "jpetazzo/dind")
+    docker.run("--privileged", "-v", pwd+":/code", "--name", "host2", "-tid", "jpetazzo/dind")
+
+    host1_exec("while ! docker ps; do sleep 1; done && "
```
robbrockbank: This doesn't feel like it belongs in package setup.  Different tests may use completely different numbers of hosts.

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
+from time import sleep
+import sh
+from sh import docker
+
+
+def multi_host():
```
robbrockbank: Does this actually get run?  Might be mis-remembering but I thought the nose/unittest framework looked for unittest classes with methods prefixed with test_

tests/fv/test_diags.py
```diff
@@ -0,0 +1,7 @@
+import sh
+
+
+def test_diags():
```
robbrockbank: missing doc strings.   and elsewhere too.

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
...
+    etcd_port = "ETCD_AUTHORITY=%s:2379" % host1_ip
+    host1_listen = docker_exec.bake("-i", "host1", "bash", "-s")
+    host2_listen = docker_exec.bake("-i", "host2", "bash", "-s")
+
+    host1_listen(_in="%s %s reset || true" % (etcd_port, calicoctl))
+
+    host1_listen(_in="%s %s node --ip=%s" % (etcd_port, calicoctl, host1_ip))
+    host2_listen(_in="%s %s node --ip=%s" % (etcd_port, calicoctl, host2_ip))
+
+    calico_port = "DOCKER_HOST=localhost:2377"
```
robbrockbank: since you passing this in, I suspect the preferred way to do this is to specify docker -H=localhost:2377

tests/fv/__init__.py
robbrockbank: Should not be adding more "potentially infinite loops.  I think the while needs to timeout.

tests/fv/test_diags.py
robbrockbank: As below, does this actually get run?  Shouldn't it be part of a unittest class?

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
...
+from sh import docker
+
+
+def multi_host():
+    docker_exec = docker.bake("exec")
+
+    host1_ip = docker.inspect("--format", "'{{ .NetworkSettings.IPAddress }}'", "host1").stdout.rstrip()
+    host2_ip = docker.inspect("--format", "'{{ .NetworkSettings.IPAddress }}'", "host2").stdout.rstrip()
+
+    calicoctl = "/code/dist/calicoctl"
```
robbrockbank: Some commenting in here would be nice.  Not exactly sure why the "listen" suffix is as it is?  Are we passing in stdin or something?

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
...
+    host1_listen(_in="%s %s profile add PROF_D" % (etcd_port, calicoctl))
+
+    host1_listen(_in="%s %s profile PROF_A_C_E member add workload-A" % (etcd_port, calicoctl))
+    host1_listen(_in="%s %s profile PROF_B member add workload-B" % (etcd_port, calicoctl))
+    host1_listen(_in="%s %s profile PROF_A_C_E member add workload-C" % (etcd_port, calicoctl))
+
+    host2_listen(_in="%s %s profile PROF_D member add workload-D" % (etcd_port, calicoctl))
+    host2_listen(_in="%s %s profile PROF_A_C_E member add workload-E" % (etcd_port, calicoctl))
+
+    sleep(1)
```
robbrockbank: Why the sleeps?

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
...
+
+    host1_exec = docker_exec.bake("-t", "host1", "bash", "-c")
+
+    host1_exec("docker exec workload-A ping -c 4 192.168.1.3")
+
+    try:
+        host1_exec("docker exec workload-A ping -c 4 192.168.1.2")
+        raise
```
robbrockbank: There is no exception to raise here.  Better to add an else: branch and raise a genuine exception in there.

Same below.

tests/fv/test_multi_host.py
```diff
@@ -0,0 +1,62 @@
+from time import sleep
```
robbrockbank: The multi host test as you've written it is very much an implementation of the IPv4 version of the demo.  It is probably worth stating that.

tests/fv/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+               "docker load --input /code/calico-node.tar && "
+               "docker load --input /code/busybox.tar && "
+               "docker load --input /code/nsenter.tar && "
+               "docker load --input /code/etcd.tar")
+
+    host2_exec("while ! docker ps; do sleep 1; done && "
+               "docker load --input /code/calico-node.tar && "
+               "docker load --input /code/busybox.tar && "
+               "docker load --input /code/nsenter.tar")
```
robbrockbank: So we're only running etcd on the one host at the moment?  Any way to make an actual etcd cluster in this approach?

tests/fv/__init__.py
alexaltair: As it stands, docker-jenkins is *only* allowed to run the docker-fvs job. So it seems quite reasonable to interpolate the entire list of containers into this.

tests/fv/__init__.py
tomdee: I think it would be better to run etcd in it's own container rather than inside a DinD container.
* This avoids the need to tar up the image.
* It becomes easier to turn the one etcd container into a cluster of etcd containers in future.
* It means the hosts becomes homogeneous.

tests/fv/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+    try:
+        sh.docker.rm("-f", "host1", "host2")
+    except sh.ErrorReturnCode:
+        pass
+    print "Host containers removed."
+
+    # Pull and save each image, so we can use them inside the host containers.
+    print sh.bash("./build_node.sh").stdout
+    docker.save("--output", "calico-node.tar", "calico/node")
+    if not os.path.isfile("busybox.tar"):
```
tomdee: Would it make sense to save these tars to /tmp/calico-build is something similar?

This would avoid passing them into the build context when building the node or build containers.

tests/fv/__init__.py
robbrockbank: Not sure that docker jenkins has anything to do with it since anyone can run the FVs.  I'm ok with the interpolation being broad though - although we probably should indicate that running the FVs will tear down all existing containers (one for the list though I suspect)

calico_containers/tests/fv/docker_host.py
```diff
@@ -0,0 +1,76 @@
...
+               "--advertise-client-urls http://%s:2379 "
+               "--listen-client-urls http://0.0.0.0:2379 "
+               "--initial-advertise-peer-urls http://%s:2380 "
+               "--listen-peer-urls http://0.0.0.0:2380 "
+               "--initial-cluster-token etcd-cluster-2 "
+               "--initial-cluster calico=http://%s:2380 "
+               "--initial-cluster-state new" % (host_ip, host_ip, host_ip))
+        self.execute("docker run -d -p 2379:2379 quay.io/coreos/etcd:v2.0.10 %s" % cmd)
+
+    @classmethod
```
robbrockbank: I would pull these out of the class and just have them as useful helper methods in the module.  Not sure it gains much from being a class method.

You could add an instance method delete() which could then take no parameters.

calico_containers/tests/fv/docker_host.py
```diff
@@ -0,0 +1,76 @@
...
+    def delete_container(cls, name):
+        """
```
robbrockbank: Need param strings.

I belive this is actually the name or the container ID no?

calico_containers/tests/fv/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+import sh
+from sh import docker
+from docker_host import DockerHost
+
+
+def setup_package():
+    """
+    Sets up docker images and host containers for running the STs.
+    """
+    containers = docker.ps("-qa").split()
```
robbrockbank: This chunk is repeated here and in setup and teardown.  Just add another helper method to the DockerHost class.

robbrockbank: Actually do you need this at all if the setup/teardown is going to do this?

calico_containers/tests/fv/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        docker.save("--output", "nsenter.tar", "jpetazzo/nsenter:latest")
+    if not os.path.isfile("etcd.tar"):
+        docker.pull("quay.io/coreos/etcd:v2.0.10")
+        docker.save("--output", "etcd.tar", "quay.io/coreos/etcd:v2.0.10")
+
+    # Create the calicoctl binary here so it will be in the volume mounted on the hosts.
+    print sh.bash("./create_binary.sh")
+    print "Calicoctl binary created."
+
+    host1 = DockerHost('host1')
```
robbrockbank: Superfluous no?  This will just be torn down in the test setup

calico_containers/tests/fv/docker_host.py
```diff
@@ -0,0 +1,76 @@
...
+from sh import docker
+
+
+class DockerHost(object):
+    """
+    A host container which will hold workload containers to be networked by calico.
+    """
+    def __init__(self, name):
+        """
+        Create a container using an image made for docker-in-docker. Load saved images into it.
```
robbrockbank: Missing parms in all your doc strings.

robbrockbank:
Looks good.  Few more minor comments and you need to fix the FVs.

Provided there are only minor modifications I don't need to see it again.

calico_containers/tests/fv/docker_host.py
alexaltair: Partly it felt like it was good to have this class be responsible for deleting hosts, and partly I didn't want to make a *forth* new file.

There's an instance method delete in here also.

robbrockbank: I wouldn't add a new file - i'd just make them methods in this file.  I'm not overly worried about this.

My rationale is that the DockerHost encapsulates a DockerHost, but these methods might not even be used on a Docker host - they could be used on any container.

If you left it unchanged I wouldn;t lose sleep.

calico_containers/tests/fv/docker_host.py
alexaltair: I feel like param spec'ing would be too much bloat in these tests for the moment. I'm comfortable describing what non-obvious parameters are about, but if you're advocating the `:param:` lines for every parameter and method, I'd like to have that policy discussion as a group first.

alexaltair: (It is indeed name or ID; I'll write that in the docstring.)

Issue #133, opened 2015-05-18
calicoctl appears to break when specifying etcd


chuckbutler:
When kicking off calicoctl w/ the linked binary in the docs: wget https://github.com/Metaswitch/calico-docker/releases/download/v0.0.6/calicoctl

I get some additional feedback from the binary when providing an external etcd server. It appears the python code is expecting another argument that was not passed:

```
ubuntu@ip-10-35-160-172:~$ sudo ./calicoctl master --ip=10.35.160.72 --etcd=10.230.43.76
Traceback (most recent call last):
  File "<string>", line 513, in <module>
  File "<string>", line 333, in master
  File "<string>", line 79, in __init__
ValueError: need more than 1 value to unpack
```

chuckbutler:
Upon further investigation, it appears this binary linked in the docs is quite old, and this may be resolved with a quick PR to the doc.

tomdee:
@robbrockbank Can you update the docs to point at Alex's special "latest" URL (it's in the video he recorded here http://www.projectcalico.org/getting-started/docker/)

@chuckbutler Does an ENV var work for you or do you need the option of passing on the CLI?

lxpollitt:
@robbrockbank We should include both http://projectcalico.org/latest/calicoctl and an example of getting a specific release from github (as today).  You should flag that projectcalico.org/latest/calicoctl is not HA, so may not be suitable for production orchestration systems.

robbrockbank:
Documentation has been updated.  Closing issue.

Fixed in #140 and #144

chuckbutler:
Sorry about the delay. the ENV Var did indeed work for me, when setting ETCD_AUTHORITY things behaved as expected.

PR #140, opened 2015-05-20
Update docs to give options for using latest beta or specific version

robbrockbank:
To handle #133

### DIFF
--- a/docs/ManualClusterSetup.md
+++ b/docs/ManualClusterSetup.md
@@ -37,12 +37,21 @@ The [example script][example-commands] assumes that your ordinary user account h
 
 ## Getting Calico Binaries
 
-Retrieve the `calicoctl` binary.  On each host run
+Get the calico binary onto each host. It's usually safe to just grab the latest [beta](http://projectcalico.org/latest/calicoctl).  On each host run
 
-    wget https://github.com/Metaswitch/calico-docker/releases/download/v0.4.2/calicoctl
-    chmod +x calicoctl
+    wget http://projectcalico.org/latest/calicoctl
+	chmod +x calicoctl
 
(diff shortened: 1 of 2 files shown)


tomdee:
LGTM

docs/ManualClusterSetup.md
```diff
@@ -37,12 +37,21 @@ The [example script][example-commands] assumes that your ordinary user account h
...
-Retrieve the `calicoctl` binary.  On each host run
+Get the calico binary onto each host. It's usually safe to just grab the latest [beta](http://projectcalico.org/latest/calicoctl).  On each host run
 
-    wget https://github.com/Metaswitch/calico-docker/releases/download/v0.4.2/calicoctl
-    chmod +x calicoctl
+    wget http://projectcalico.org/latest/calicoctl
+	chmod +x calicoctl
 
-Finally, preload the Calico Docker image.  This will make the demo more responsive the first time you run it.
+The beta is not HA so may not be suitable for production environments.  Alternatively, you can download a specific [release](https://github.com/Metaswitch/calico-docker/releases/).  For example, to retrieve the most recent release, on each host run
```
lxpollitt: The point to flag here is not that the beta is not HA, it is that projectcalico.org is not a HA repository.  Something like this would be better: "projectcalico.org is not a HA repository, so using this download URL is not recommend for any automated production installation process."  The "For example, to retrieve the mot recent release" is not right either.  That example downloads a specific versioned release.  Same comments apply to Orchestrators.md.

robbrockbank: OK will update. 

Regarding the latter point, I was thinking that our release process would still update the versions in these documents, but in hindsight I'm not sure that was a good decision on my part.

PR #141, opened 2015-05-20
Add graceful restart to BIRD config, and use confd prefix to minimize config churn

robbrockbank:


node/templates/bird.cfg.template
```diff
@@ -19,6 +21,7 @@ protocol kernel {
   import all;
   device routes;
   export all;     # Default is export none
+  graceful restart;
```
Symmetric: Probably worth leaving a comment explaining what bug this flag fixed

node/templates/bird.cfg.template
```diff
@@ -46,24 +49,25 @@ template bgp bgp_template {
                     # local address as nexthop
   source address {{getenv "IP"}};  # The local address we use for the TCP connection
   add paths on;
+  graceful restart;
```
Symmetric: Worth mentioning the route flap bug here, too -- in case someone is tempted to remove the GR flag in the future.

node/templates/bird.cfg.template
```diff
@@ -1,13 +1,15 @@
-# Generated by confd {{datetime}}
+# Generated by confd
 router id {{getenv "IP"}};
 log "/var/log/calico/bird.log" all;
 
 filter calico_pools {
-{{range gets "/calico/v1/ipam/v4/pool/*"}}
+{{if exists "/ipam/v4/pool"}}
```
Symmetric: It's not obvious why you are doing the {{if}} test first, probably worth a comment.

Symmetric:
All the requests for comments on the v4 template apply to the v6 template too. Happy for you to merge once you've beefed up the commenting a bit.

node/templates/bird.cfg.template
robbrockbank: Is it possible to add a comment in the template without it going into the actual bird config?

Symmetric: http://golang.org/pkg/text/template/ says that {{/* is a comment */}}

PR #143, opened 2015-05-21
Rename FVs to STs.

alexaltair:


Symmetric:
LGTM, minor point about a commented line but happy for this to go in

PR #146, opened 2015-05-22
Fix BIRD issue with unnamed protocols using templates

robbrockbank:
This works around an issue with BIRD that does not handle unnamed protocols when using templates.

The key change here is to add a name to the profiled protocol section.

so
```
protocol bgp  from bgp_template {   ->   protocol bgp bgp_name from bgp_template {
```

This is required to get GR working.  Without this, a reload of BIRD config results in all routes being dropped.

I've named each protocol section according to the IP address of the peer (with . and : replaced with _).


spikecurtis:
LGTM.

PR #147, opened 2015-05-22
Convert the rest of the STs to python.

alexaltair:


calico_containers/tests/st/docker_host.py
```diff
@@ -15,16 +15,21 @@ def __init__(self, name):
         pwd = sh.pwd().stdout.rstrip()
         docker.run("--privileged", "-v", pwd+":/code", "--name", self.name, "-tid", "jpetazzo/dind")
```
spikecurtis: I don't think we're using Docker very well when we use this design pattern.  We go to all the trouble of using Docker to be able to create containerized builds and ST runs, and then we just mount our whole repo into the container.

Instead, what we should be doing is using jpetazzo/dind as a base image, then build a new Docker image that has the code to be tested copied in and the inner Docker images preloaded.

The only volumes we should need to mount are either 

a) persistent storage (don't see why we'd need that for ST)
b) output directories for logs, coverage, etc.

This comment also applies to the way we do builds and UTs, but I'm happy to raise a separate issue for those.

alexaltair: Yeah, it kinda feels that way to me too. It should be noted though that we aren't using the host containers primarily for isolation, but instead for rapid testing of calico on multiple hosts.

Seems like your suggestion is the direction that we should move in, but also that at this point I should probably start doing something other than improving the test framework now that we can easily write multi-host tests in python.

Symmetric: Mounting the code directory in the build container is a common paradigm for
build containers; it means we don't have to rebuild the container every
time we make a code change.

Not sure if the rest of the test code allows this rebuild-free fast path,
but it would be a nice feature to have.

(I'm a fan of not rebuilding docker containers if they are there, and using
a `make clean` to blow everything away in the rare case that requires a
rebuild, rather than trying to guarantee that `make test` will work every
time. The build server can just do `make clean test`.)

tomdee: Yep, I'm very much with @Symmetric here. It _is_ a common pattern to mount code into the container. Docker themselves do it for their build containers.

spikecurtis: This isn't a build container, it's a system test container.  It would be really handy to have the actual ST container artifact as something you can pull down from failed ST runs.  The way this framework is set up, we repeat the work of caching the relevant container images for every host (mercifully we only pull from the external source once per run).

Furthermore, Docker's FS layering means that we should be able to engineer a path where we don't need to build from scratch every time there is a code change---we just re-layer the final changes.

calico_containers/tests/st/test_base.py
```diff
@@ -44,3 +48,32 @@ def start_etcd(self):
             initial_cluster="calico=http://%s:2380" % self.ip,
             initial_cluster_state="new",
         )
+
+    def assert_powerstrip_up(self, host):
```
spikecurtis: I don't get how this asserts that powerstrip is up.  It looks like it just tries `docker ps` until it gets a successful error code, which will happen regardless of whether Powerstrip is up.

calico_containers/tests/st/test_add_container.py
```diff
@@ -0,0 +1,25 @@
...
+        host.execute(calicoctl % "node --ip=127.0.0.1")
+        host.execute(calicoctl % "profile add TEST_GROUP")
+
+        self.assert_powerstrip_up(host)
+
+        host.execute(calicoctl % "container add node 192.168.1.1")
+        host.execute(calicoctl % "profile TEST_GROUP member add node")
+
+        # Wait for felix to program down the route.
+        powerstrip = partial(host.execute, "ip route | grep '192\.168\.1\.1'")
```
spikecurtis: Why is this called `powerstrip`?

calico_containers/tests/st/test_base.py
alexaltair: It's the `docker_host=True` part. That will use port 2377 which will only work if powerstrip is up.

calico_containers/tests/st/test_add_ip.py
```diff
@@ -0,0 +1,69 @@
...
+from test_base import TestBase
+from docker_host import DockerHost
+
+
+class TestAddIp(TestBase):
+    def test_add_ip(self):
+        host = DockerHost('host')
+
+        calicoctl = "/code/dist/calicoctl %s"
+        host.execute(calicoctl % "node --ip=172.17.8.10")
```
spikecurtis: Where does this IP address come from?

calico_containers/tests/st/test_add_ip.py
```diff
@@ -0,0 +1,69 @@
...
+        host.execute(calicoctl % "profile add TEST_GROUP")
+
+        self.assert_powerstrip_up(host)
+
+        host.execute("docker run -e CALICO_IP=192.168.1.1 -tid --name=node1 busybox", docker_host=True)
```
spikecurtis: Thinking about this more, I find the name `docker_host` to be unhelpful.  What this option actually does is determine whether to route the docker command via Powerstrip or not.  Maybe `use_powerstrip` is a better name.

calico_containers/tests/st/test_add_container.py
```diff
@@ -0,0 +1,25 @@
...
+        assert self.retry_until_success(powerstrip, ex_class=ErrorReturnCode)
```
spikecurtis: This feels like a generally useful function, can you pull it out into `DockerHost` as `assert_route_exists` or something.

spikecurtis:
A general comment: most of these tests just convert the original bash scripts to python where you essentially convert the commands to `host.execute` lines.

I'd much rather see some reusable, high level python commands that eventually map down to the command line calls rather than putting the CLI syntax directly into all our STs.  I think we'll find this much harder to maintain and build on in the long run.

calico_containers/tests/st/test_add_container.py
```diff
@@ -0,0 +1,25 @@
...
+from docker_host import DockerHost
+
+
+class TestAddContainer(TestBase):
+    def test_add_container(self):
+        host = DockerHost('host')
+
+        calicoctl = "/code/dist/calicoctl %s"
+
+        host.execute("docker run -tid --name=node busybox")
```
spikecurtis: It's not clear the difference between `host.execute` and `host.listen`.  I think the comments for them are out of date.

calico_containers/tests/st/test_add_ip.py
```diff
@@ -0,0 +1,69 @@
...
+
+        host.execute(calicoctl % "container node2 ip add 192.168.2.2 --interface=hello")
+
+        host.execute("docker exec node1 ping 192.168.2.2 -c 1")
+        host.execute("docker exec node2 ping 192.168.1.1 -c 1")
+        host.execute("docker exec node2 ping 192.168.2.1 -c 1")
+        host.execute("docker exec node2 ping 192.168.3.1 -c 1")
+
+        # Now stop and restart node 1 and node 2.
+        host.execute("sudo docker -H=localhost:2377 stop node1")
```
spikecurtis: why use `-H=localhost:2377` here but `DOCKER_HOST=localhost:2377` elsewhere?

calico_containers/tests/st/test_add_ip.py
```diff
@@ -0,0 +1,69 @@
...
+        host.execute("sudo docker -H=localhost:2377 stop node2")
+        host.execute("sudo docker -H=localhost:2377 start node1")
+        host.execute("sudo docker -H=localhost:2377 start node2")
+
+        assert self.retry_until_success(test_ping, ex_class=ErrorReturnCode)
+
+        # Test pings between the IPs.
+        host.execute("docker exec node1 ping 192.168.1.2 -c 1")
```
spikecurtis: Surely we can make this code more compact and easier to understand with python iteration?

calico_containers/tests/st/test_add_ip.py
```diff
@@ -0,0 +1,69 @@
...
+        host.execute("docker exec node1 ping 192.168.2.2 -c 1")
+        host.execute("docker exec node2 ping 192.168.1.1 -c 1")
+        host.execute("docker exec node2 ping 192.168.2.1 -c 1")
+        host.execute("docker exec node2 ping 192.168.3.1 -c 1")
+        host.execute(calicoctl % "shownodes --detailed")
```
spikecurtis: Why are we running a `shownodes` without checking any of it's output?

calico_containers/tests/st/test_arg_parsing.py
```diff
@@ -0,0 +1,99 @@
...
+        with self.assertRaises(ErrorReturnCode_1):
+            host.execute(calicoctl % "container node1 ip add aa:bb::zz")
+        with self.assertRaises(ErrorReturnCode_1):
+            host.execute(calicoctl % "container add node1 127.a.0.1")
+        with self.assertRaises(ErrorReturnCode_1):
+            host.execute(calicoctl % "container add node1 aa:bb::zz")
+
+        # Add some pools and BGP peers and check the show commands"
+        host.execute(calicoctl % "bgppeer rr add 1.2.3.4")
+        host.execute(calicoctl % "bgppeer rr show")
```
spikecurtis: I don't get why we run this once, dropping the output, then run it again in a few lines.

calico_containers/tests/st/test_arg_parsing.py
```diff
@@ -0,0 +1,99 @@
...
+        host.execute(calicoctl % "bgppeer rr show --ipv4")
+        host.execute(calicoctl % "bgppeer rr show --ipv6")
+        assert "1.2.3.4" in host.execute(calicoctl % "bgppeer rr show").stdout.rstrip()
```
spikecurtis: The `self.assert*` functions in nose give much easier to read output in test logs; please use them.

calico_containers/tests/st/test_base.py
spikecurtis: Ah!  Can you add a comment that explains this.

calico_containers/tests/st/test_arg_parsing.py
```diff
@@ -0,0 +1,99 @@
...
+        assert "aa:bb::ff/128" not in host.execute(calicoctl % "pool show --ipv4").stdout.rstrip()
+        assert "aa:bb::ff/128" in host.execute(calicoctl % "pool show --ipv6").stdout.rstrip()
+        host.execute(calicoctl % "pool remove aa:bb::ff/128")
+        assert "aa:bb::ff/128" not in host.execute(calicoctl % "pool show").stdout.rstrip()
+
+        # Not used anywhere else in the tests; added here for completeness.
+        host.execute(calicoctl % "profile add TEST_PROFILE")
+        host.execute(calicoctl % "profile TEST_PROFILE tag add TEST_TAG")
+        host.execute(calicoctl % "profile TEST_PROFILE tag remove TEST_TAG")
+        host.execute(calicoctl % "profile TEST_PROFILE tag show")
```
spikecurtis: We should test the output of these show commands as well.

calico_containers/tests/st/test_base.py
```diff
@@ -44,3 +48,32 @@ def start_etcd(self):
...
+        :param retries: the maximum number of times to retry the function.
+        A value of 0 will run the function once with no retries.
+        :param ex_class: The class of expected exceptions.
+        :returns: the value returned by function
+        """
+        for retry in range(retries + 1):
+            try:
+                result = function()
+            except Exception as e:
+                if ex_class and not issubclass(e.__class__, ex_class):
```
spikecurtis: What you're looking for is `isinstance(e, ex_class)` rather than the slightly more roundabout `issubclass`

calico_containers/tests/st/test_base.py
```diff
@@ -44,3 +48,32 @@ def start_etcd(self):
...
+        :param function: the function to be repeatedly called
+        :param retries: the maximum number of times to retry the function.
+        A value of 0 will run the function once with no retries.
+        :param ex_class: The class of expected exceptions.
+        :returns: the value returned by function
+        """
+        for retry in range(retries + 1):
+            try:
+                result = function()
+            except Exception as e:
```
spikecurtis: I feel like it would be cleaner to default `ex_class=Exception` and then on this line do `except ex_class as e:`.  That would remove the need for the `if not issubclass` statement.

calico_containers/tests/st/test_mainline.py
```diff
@@ -18,22 +19,16 @@ def run_mainline(self, ip1, ip2):
...
         host.execute("docker run -e CALICO_IP=%s -tid --name=node1 busybox" % ip1, docker_host=True)
         host.execute("docker run -e CALICO_IP=%s -tid --name=node2 busybox" % ip2, docker_host=True)
 
         # Perform a docker inspect to extract the configured IP addresses.
-        node1_ip = host.execute("docker inspect --format '{{ .NetworkSettings.IPAddress }}' node1", docker_host=True).stdout.rstrip()
-        node2_ip = host.execute("docker inspect --format '{{ .NetworkSettings.IPAddress }}' node2", docker_host=True).stdout.rstrip()
+        node1_ip = host.execute("docker inspect --format '{{ .NetworkSettings.IPAddress }}' node1",
+                                docker_host=True).stdout.rstrip()
+        node2_ip = host.execute("docker inspect --format '{{ .NetworkSettings.IPAddress }}' node2",
+                                docker_host=True).stdout.rstrip()
```
spikecurtis: Assert right away that we got the right IPs (if we didn't pass 'auto')

calico_containers/tests/st/test_mainline.py
```diff
@@ -42,15 +37,8 @@ def run_mainline(self, ip1, ip2):
...
-        for i in range(10):
-            try:
-                host.execute("./nsenter -t %s ping %s -c 1 -W 1" % (node1_pid, node2_ip))
-                break
-            except ErrorReturnCode:
-                if i == 9:
-                    raise AssertionError("Network failed to come up.")
-                else:
-                    sleep(1)
+        ping = partial(host.execute, "./nsenter -t %s ping %s -c 1 -W 1" % (node1_pid, node2_ip))
```
spikecurtis: If `nsenter` is our preferred way to run commands in containers, then you should use it everywhere in ST, not just here.

calico_containers/tests/st/test_multi_host.py
```diff
@@ -1,50 +1,37 @@
...
-from time import sleep
 from sh import ErrorReturnCode, ErrorReturnCode_1
+from time import sleep
+import unittest
+
+from test_base import TestBase
 from docker_host import DockerHost
 
 
+@unittest.skip("Waiting for a fix from aa1-separate-etcd")
```
spikecurtis: Hasn't `aa1-separate-etcd` been merged?

calico_containers/tests/st/test_multi_host.py
```diff
@@ -1,50 +1,37 @@
...
+
+        Almost identical in function to the vagrant coreOS demo.
         """
         host1 = DockerHost('host1')
         host2 = DockerHost('host2')
 
         calicoctl = "/code/dist/calicoctl %s"
-
-        host1.execute(calicoctl % "reset || true")
+        try:
```
spikecurtis: why is a reset necessary for this test but not others?

calico_containers/tests/st/test_multi_host.py
```diff
@@ -1,50 +1,37 @@
...
+            host1.execute(calicoctl % "reset")
+        except ErrorReturnCode:
+            pass
 
         host1.execute(calicoctl % ("node --ip=%s" % host1.ip))
         host2.execute(calicoctl % ("node --ip=%s" % host2.ip))
```
spikecurtis: this really should be a method on `DockerHost` object.

calico_containers/tests/st/test_multi_host.py
```diff
@@ -62,18 +49,12 @@ def test_multi_host(self):
         # Wait for the workload networking to converge.
         sleep(1)
```
spikecurtis: can we avoid this hard-coded sleep with a retry loop?

calico_containers/tests/st/test_multi_host.py
```diff
@@ -62,18 +49,12 @@ def test_multi_host(self):
...
 
-        host1.execute("docker exec workload-A ping -c 4 192.168.1.3")
+        assert host1.execute("docker exec workload-A ping -c 4 192.168.1.3")
```
spikecurtis: Why the `assert`? I thought `execute` will raise an exception if the ping fails?

calico_containers/tests/st/test_multi_host.py
```diff
@@ -62,18 +49,12 @@ def test_multi_host(self):
...
 
-        try:
+        with self.assertRaises(ErrorReturnCode_1):
```
spikecurtis: This would be much easier to follow if you pulled the IP addresses out into local variables like `IP_A`, `IP_B` etc.

calico_containers/tests/st/test_no_powerstrip.py
```diff
@@ -0,0 +1,60 @@
...
+        host.execute("docker run --rm  -v `pwd`:/target jpetazzo/nsenter", _ok_code=[0, 1])
+
+        host.execute(calicoctl % "node --ip=127.0.0.1")
+        host.execute(calicoctl % "profile add TEST_GROUP")
+
+        self.assert_powerstrip_up(host)
+
+        # Remove the environment variable such that docker run does not utilize
+        # powerstrip.
+        host.execute("docker run -e CALICO_IP=192.168.1.1 -tid --name=node1 busybox")
```
spikecurtis: `docker_host=False` would be more clear here.  It's confusing when you say remove the environment variable, but have left the `-e CALICO_IP=...` in.

alexaltair:
The goal of the first conversion was basically just that (decided by a conversation with Tom and someone else I can't remember). I come from a mindset of wanting to build as much structure as possible, but that's from my sense of aesthetics and not from practicality. So in this case I tried to push myself pretty hard to *not* make infrastructure, in order to practice being agile and usefulness-based. So I have a pile of things I want to do to improve these tests, but I'm unclear on where to aim along that tradeoff.

That said, it doesn't seem like people are really using the tests all that much. Most of the ST runs are mine. People haven't been wanting to write tests while I've been working on them. So I think we should merge this at it's current level of complexity, and if people want more structure I will gladly implement it.

spikecurtis:
@alexaltair I'm fine merging without infrastructure if that's a trade off we're making from an agile standpoint.  But, I'd like to see that infrastructure in place before we go a lot further on writing additional tests---writing new tests in this current style is going to make more work when we add new structure.

ST runs should become part of our continuous integration infrastructure so they get run all the time by everyone.

calico_containers/tests/st/test_no_powerstrip.py
```diff
@@ -0,0 +1,60 @@
...
+        # Now add the profiles.
+        host.execute(calicoctl % "profile TEST_GROUP member add node1")
+        host.execute(calicoctl % "profile TEST_GROUP member add node2")
+
+        # Inspect the nodes (ensure this works without powerstrip)
+        host.execute("docker inspect node1")
+        host.execute("docker inspect node2")
+
+        # Check it works
+        ping = partial(host.execute, "docker exec node1 ping 192.168.1.2 -c 1 -W 1")
```
spikecurtis: You set up `nsenter` at the beginning of this test and then use `docker exec`.  What gives?

calico_containers/tests/st/test_profile_commands.py
```diff
@@ -0,0 +1,25 @@
...
+        host.execute("echo '%s' | " % json + calicoctl % "profile TEST_PROFILE rule update")
+
+        assert '1 deny' in host.execute(calicoctl % "profile TEST_PROFILE rule show").stdout.rstrip()
+
+        # Test that adding and removing a tag works.
+        assert "TEST_TAG" not in host.execute(calicoctl % "profile TEST_PROFILE tag show").stdout.rstrip()
+        host.execute(calicoctl % "profile TEST_PROFILE tag add TEST_TAG")
+        assert "TEST_TAG" in host.execute(calicoctl % "profile TEST_PROFILE tag show").stdout.rstrip()
+        host.execute(calicoctl % "profile TEST_PROFILE tag remove TEST_TAG")
+        assert "TEST_TAG" not in host.execute(calicoctl % "profile TEST_PROFILE tag show").stdout.rstrip()
```
spikecurtis: Not that it should block merging, but there is clearly a lot more that we can test here.

calico_containers/tests/st/docker_host.py
alexaltair: Just a followup; I don't plan on making this change in this PR. If there's continued disagreement about it, we can go ahead an make an issue for discussion, and I'll make a separate branch for implementing the agreed conclusion.

calico_containers/tests/st/test_base.py
alexaltair: Docstring added.

calico_containers/tests/st/test_add_container.py
alexaltair: Copy-pasta; fixed.

calico_containers/tests/st/test_add_ip.py
alexaltair: The author of the original shell script. Should we default to `127.0.0.1` in single-host tests?

calico_containers/tests/st/test_add_ip.py
alexaltair: Good call. Changed.

calico_containers/tests/st/test_add_container.py
alexaltair: I considered it; but ATM it's only used in this one test, so we might want to wait until it proves itself useful before functioning out. (We might even want to change this to a ping assertion instead.)

calico_containers/tests/st/test_add_container.py
alexaltair: Fixed.

calico_containers/tests/st/test_add_ip.py
alexaltair: If it were just `for i in range(5)` I'd do it. As is I'd like to wait until we figure out the right shape for a `check_mutual_connectivity` function (a la the OpenStack tests).

calico_containers/tests/st/test_add_ip.py
alexaltair: Originally, because the env var didn't make it through the `sudo`. I'm retesting it now without the `sudo`.

calico_containers/tests/st/test_add_ip.py
alexaltair: There used to be a whole bunch of show commands in some of the tests, I think as a sanity check because developers were running them locally. But they never asserted on it, and ultimately they didn't contribute to the pass or fail. So I took them out for this round. We should probably add back some assertions on the show commands next time.

calico_containers/tests/st/test_arg_parsing.py
alexaltair: Probably originally so the developer could see the output. Removing them since they're asserted on.

calico_containers/tests/st/test_arg_parsing.py
alexaltair: Fixed.

calico_containers/tests/st/test_add_ip.py
spikecurtis: We should just use `host.ip` as we do in multi-host tests.

calico_containers/tests/st/test_add_container.py
spikecurtis: ping assertions are used all over, so that's definitely worth pulling out.

calico_containers/tests/st/test_arg_parsing.py
alexaltair: Turns out these are asserted on in `test_profile_commands`.

spikecurtis: Then just drop those from this test, I think.

calico_containers/tests/st/test_base.py
alexaltair: Huh. Thought that didn't work, but I must have tested it wrong. Fixed.

calico_containers/tests/st/test_base.py
alexaltair: Nice!

calico_containers/tests/st/test_mainline.py
alexaltair: Nice catch.

calico_containers/tests/st/test_multi_host.py
alexaltair: Fixed.

calico_containers/tests/st/test_mainline.py
alexaltair: I don't really know why Rob added `nsenter`. I'll consider this if/when I get to writing an `assert_can_ping`.

calico_containers/tests/st/test_multi_host.py
alexaltair: I swear I copied this from somewhere, but I can't find it now. I'll take it out.

calico_containers/tests/st/test_multi_host.py
alexaltair: Agreed; I'll write it for the next round.

calico_containers/tests/st/test_multi_host.py
alexaltair: Fixed.

calico_containers/tests/st/test_multi_host.py
alexaltair: I was trying to be all semantically accurate. But since we don't do this anywhere else I'll remove it and wait for an `assert_can_ping` for semantic correctness.

calico_containers/tests/st/test_multi_host.py
alexaltair: Done.

calico_containers/tests/st/test_no_powerstrip.py
alexaltair: Fixed.

calico_containers/tests/st/test_no_powerstrip.py
alexaltair: Blindly copied from the scripts. (See comment above for my plan re: nsenter.)

PR #151, opened 2015-05-27
Adds the calico-docker charm concept

chuckbutler:
- Includes README for instruction under charm/trusty/calico-docker
- Includes bundle yaml deployable with juju-deployer, details in
  README.md


charm/trusty/calico-docker/hooks/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+import sys
+from path import path
+from charmhelpers.contrib.ansible import AnsibleHooks
+
+# Create the hooks helper, passing a list of hooks which will be
+# handled by default by running all sections of the playbook
+# tagged with the hook name.
+
+def hook_names(here=path(__file__).parent):
+    for name in (x.basename() \
```
Lukasa: This trailing backslash is unnecessary. =)

charm/trusty/calico-docker/playbooks/config-changed.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+
+    - name: Pull CalicoNode image
+      docker:
+          name=caliconode
+          image={{ docker_image }}
+          state=present
+
+    - name: Pull CalicoCTL
+      get_url: url={{calicoctl_package}} sha256sum={{calicoctl_sum}} dest="/usr/local/bin/calicoctl"
```
Lukasa: This is really neither here nor there, but it turns out that ansible doesn't do proper cert validation here, so the SHA sum is really the only protection we have. I'll be opening an issue about this shortly, but hey, now we know!

Lukasa: Issue is here: ansible/ansible#11081.

Lukasa: For the sake of security we're *probably* ok because we're using the sha256 sum, but we're still technically a bit vulnerable. For now, that'll have to do.

Lukasa:
I've made some small notes about one or two things, but generally this *looks* fine. I haven't had a chance to actually deploy any of this yet, but at a top level I'd be happy to have my name down as a co-maintainer here.

charm/bundle/bundles.yaml
```diff
@@ -0,0 +1,21 @@
...
+    docker:
+      charm: cs:trusty/docker
+      num_units: 2
+      options:
+        latest: true
+        version: 1.6.2
+    calico-docker:
+      charm: local:trusty/calico-docker
+      options:
+        docker-image: calico/node:latest
```
tomdee: This should be tied to releases rather than the auto builds from master. It's another thing to update and test when we do a release, but worth it as master can be unstable at times.

charm/trusty/calico-docker/README.md
```diff
@@ -0,0 +1,78 @@
...
+# Configurable Options
+
+The calico-node image is configurable, to limit the possiblity of having
+differing versions of the calico-node image, it's suggested to lock the version
+in the charm configuration so each unit you add at scale will use the proper
+image.
+
+    juju set calico-docker docker-image=calico/node:v0.4.2
+
+The CalicoCTL tool is also versioned, along with a sha256 sum to ensure the
```
tomdee: Interesting capitalization :-) allow lower case is what we normally use when talking about calicoctl.

charm/trusty/calico-docker/README.md
```diff
@@ -0,0 +1,78 @@
...
+payload is delivered correctly:
+
+    juju set calico-docker calicoctl-package=https://github.com/Metaswitch/calico-docker/releases/download/v0.4.2/calicoctl
+    juju set calico-docker calicoctl-sum=90cd21fed0abae9ed524c95d3cff8192dfc133150cdf386951ed9daa28c2d889
+
+## Maintainers:
+
+- Charles Butler &lt;[charles.butler@canonical.com](mailto:charles.butler@canonical.com)&gt;
+- Cory Benfield &lt;[cory.benfield@metaswitch.com](mailto:cory.benfield@metaswitch.com)&gt;
```
tomdee: @Lukasa Would your projectcalico.org address be better here?

Lukasa: I don't think so: my Launchpad.net account is @metaswitch.com, not @projectcalico.org.

charm/bundle/bundles.yaml
tomdee: Ah - just read the readme below. If this file is just used for testing then latest might be the right docker image to use.

charm/trusty/calico-docker/playbooks/stop.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - set_fact:
+        db_relation: "{{ relations.db | default([]) }}"
+
+    - name: Nuke all traces of CalicoNode image
+      docker:
+          name=caliconode
+          image={{ docker_image }}
+          state=absent
+
+    - name: delete CalicoCTL
```
tomdee: Again, just use all lower case for calicoctl

tomdee:
I've taken a quick look and only a few minor comments from me. We should get someone here to test it before I merge and that's unlikely to happen till early next week.

charm/trusty/calico-docker/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Project Calico
+
+- Upstream website
+  - [projectcalico.org](http://projectcalico.org)
+- Upstream bug tracker
+  - [Calico-Docker bugtracker](https://github.com/Metaswitch/calico-docker/issues)
+- Upstream mailing list or contact information
+  - [Calico Mailing List](http://lists.projectcalico.org/listinfo/calico)
```
alexwlchan: The mailing list split in two about a month ago: calico-tech and calico-announce. There’s probably not going to be much more traffic on the calico list.

This link should be changed to point at the current list, and/or the [contact page on the Calico site](http://www.projectcalico.org/contact/).

lxpollitt:
Note that we also need contributors agreement in place before merging per guidelines here: http://docs.projectcalico.org/en/latest/contribute.html.

charm/trusty/calico-docker/README.md
ed-harrison: Yes - I think the correct list is calico-tech (http://lists.projectcalico.org/listinfo/calico-tech) - the original "calico" list is, effectively, dead.

chuckbutler:
Greetings,

Excellent feedback! Thanks for the review of the calico-docker charm. @tomdee I haven't got a good test-case back thus far, and was looking forward to working with @Lukasa on rounding that final stretch where we have verified connectivity captured in tests, and a working Getting Started action demo for anyone passing by.

As far as I can tell, the charm is making all the right moves, and its setting up properly. I feel the problem is between the keyboard and chair - and once we have that lined out, we should be good to go.

I've got an open request in to our team to get moving on the Corp CLA and that should be handled shortly. 

Thanks again for taking a look team! :thumbsup:

Lukasa:
Ok, I gave this a whirl today and that didn't work out. The big problem seems to be this line:

```
unit-calico-docker-1[12565]: 2015-06-22 09:22:32 INFO unit.calico-docker/1.db-relation-changed logger.go$40 fatal: [localhost] => One or more undefined variables: 'dict object' has no attribute 'hostname'
```

This points me to a line in `config-changed.yaml` which looks wrong:

```yaml
    - set_fact:
        etcd_authority: "{{ relations.db[0].hostname }}:{{ relations.db[0].port }}"
      when: db_relation != []
```

Looks like `relations.db[0]` is a dict, not an object.

Lukasa:
While we're here, it doesn't look like the etcd charm sets `port` on the client relation (from some code read). Is that right?

Lukasa:
Nevermind, I see that `port` is set. So the problem is just the dict access I think.

charm/trusty/calico-docker/hooks/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+
+import sys
+from path import path
+from charmhelpers.contrib.ansible import AnsibleHooks
+
+# Create the hooks helper, passing a list of hooks which will be
+# handled by default by running all sections of the playbook
+# tagged with the hook name.
+
+def hook_names(here=path(__file__).parent):
```
Lukasa: It looks like the location of this comment screws with the linter, which leads it to think it's only got one blank line instead of two. I recommend moving it.

charm/trusty/calico-docker/hooks/setup.py
```diff
@@ -0,0 +1,27 @@
...
+def install_charmhelpers():
+    """
+    Install the charmhelpers library, if not present.
+    """
+    try:
+        import charmhelpers  # noqa
+    except ImportError:
+        subprocess.check_call(['apt-get', 'install', '-y', 'python-pip'])
+        subprocess.check_call(['pip', 'install', 'charmhelpers'])
+
```
Lukasa: Can we get an extra blank line here?

charm/trusty/calico-docker/hooks/setup.py
```diff
@@ -0,0 +1,27 @@
...
+def install_python_modules():
+    subprocess.check_call(['pip', 'install', 'ansible'])
+    subprocess.check_call(['pip', 'install', 'path.py'])
+    subprocess.check_call(['pip', 'install', 'docker-py'])
+
+
```
Lukasa: And can we get rid of one of these blank lines? (I basically want `make lint` to succeed here!)

charm/trusty/calico-docker/Makefile
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/make
+
+build: virtualenv lint test
```
Lukasa: `test` is not an actual target in this makefile...

charm/trusty/calico-docker/Makefile
```diff
@@ -0,0 +1,26 @@
...
+	.venv/bin/pip install nose flake8 mock pyyaml charmhelpers charm-tools ecdsa ansible ansible-lint
+
+lint: .venv/bin/python
+	@.venv/bin/flake8 hooks
+	@.venv/bin/ansible-lint playbooks/site.yaml
+	@.venv/bin/charm proof
+
+unit_test: .venv/bin/python
+	@echo Starting tests...
+	@CHARM_DIR=. PYTHONPATH=./hooks .venv/bin/nosetests -vv --nologcapture unit_tests
```
Lukasa: This target doesn't work! I suspect the reason for that is that this charm doesn't actually *have* any unit tests. I'm not too bothered by that, but we shouldn't have a makefile target that doesn't correspond to anything... ;)

charm/trusty/calico-docker/README.md
```diff
@@ -0,0 +1,80 @@
...
+in the charm configuration so each unit you add at scale will use the proper
+image.
+
+    juju set calico-docker docker-image=calico/node:v0.4.2
+
+The calicoctl tool is also versioned, along with a sha256 sum to ensure the
+payload is delivered correctly:
+
+    juju set calico-docker calicoctl-package=https://github.com/Metaswitch/calico-docker/releases/download/v0.4.2/calicoctl
+    juju set calico-docker calicoctl-sum=90cd21fed0abae9ed524c95d3cff8192dfc133150cdf386951ed9daa28c2d889
```
Lukasa: We might want to update this section of the documentation to point out that these are actually the default values, and so don't need to be explicitly set.

chuckbutler:
@Lukasa Thanks for the review feedback. I've addressed the listed concerns above and have resubmit for another review.

Lukasa:
Alright, this generally LGTM. I can't easily test it right now (because I'm on holiday), so ideally @tomdee would take a swing at it, but if not I'll test it when I get back.

tomdee:
I'm not going to be able to look this week. I'm also heading off on vacation soon but will be back on Monday. Let's aim to get this merged early next week.

chuckbutler:
@tomdee obligatory ping to see if there has been any time to evaluate the MP

tomdee:
I was just talking to @Lukasa about this today. I'm slightly concerned about some of the PR being out of date given how things on calico-docker have moved in the last couple of months. Hopefully Lukasa will get a chance to look at this again tomorrow and help me decide what to do (if anything!) - maybe we can just merge it as is.

So - watch this space and hopefully we'll have an update tomorrow.

chuckbutler:
I'm happy to refactor with direction. This has been a work in progress since the first steps, and i'd like to make it a low maintenance burden so its a smooth transition during the maintenance cycles.

Lukasa:
So here's the state of play as I can see it from a quick review: this PR is currently pinned against the 0.4.2 release. This means it currently is still consistent, and using powerstrip. That's all fine, but it does not appear to be what we recommend anymore, so I think we should bring it a bit more up-to-date and to use the newer networking proposal. @tomdee, can you give @chuckbutler some guidance as to how you think this *should* be deployed (e.g. libnetwork or not)?

tomdee:
Since libnetwork is still experimental only, I can't recommend that this is refactored to use that. 

I think it would make sense to base it around the "default-networking" option - see https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/default-networking/Demonstration.md

chuckbutler:
in the interest of keeping history cohesive - adding notes from a call w/ tomdee here.

Powerstrip is now deprecated fully

1.9 stable will expose libnetwork, and is going to land @ public in October.
Targeting the libnetwork integration, would be the best path forward. Supporting this will have a few implications:

- Implement the libnetwork bits to include calico bits.

##  What it works with (docker libkv golang library)
- zookeeper
- consul

Caveats:
- does not work with etcd yet (requires consul + etcd)

pointer for libnetwork: https://github.com/projectcalico/libnetwork-plugin
docs for libnetwork integration: https://github.com/projectcalico/calico-docker/tree/master/docs/getting-started/libnetwork

tomdee:
@chuckbutler 
FYI - libnetwork is now in Docker master (non-experimental) and will be in the 1.9 release in a couple of weeks. And it supports etcd.

We will shortly be doing a release of calico-docker with an updated libnetwork-plugin

tomdee:
@chuckbutler Docker have just released 1.9 and we've just done a calico-docker release to go with it.

tomdee:
Closing this until we find some more time to spend on it.

PR #152, opened 2015-05-29
Fix create_binary build permission issues.

spikecurtis:
Fixes
```
File "/usr/local/bin/pyinstaller", line 9, in <module>
load_entry_point('PyInstaller==2.1', 'console_scripts', 'pyinstaller')()
File "/usr/local/lib/python2.7/dist-packages/PyInstaller/main.py", line 86, in run
spec_file = run_makespec(opts, args)
File "/usr/local/lib/python2.7/dist-packages/PyInstaller/main.py", line 40, in run_makespec
spec_file = PyInstaller.makespec.main(args, **opts.__dict__)
File "/usr/local/lib/python2.7/dist-packages/PyInstaller/makespec.py", line 322, in main
specfile = open(specfnm, 'w')
IOError: [Errno 13] Permission denied: '/code/calicoctl.spec'
```

Root cause is that the `create_binary.sh` script mounts the project root as `/code` inside the build container.  However, the user account inside the container may not have permission to write to this directory.

This worked for a lot of people's boxes because of a coincidence: the UID of the user trying to build is 1000, which is the same UID as the `user` account created inside the container.  If the user trying to build isn't UID 1000, they hit permission errors.

Solution is not to mount the whole root directory to `/code`.  Instead, mount `calico_containers` to `/code/calico_containers`, which just needs read access.  Mount `dist` to `/code/dist` as the output directory.  This gets set with 777 permissions, so the container use can write to it.

### DIFF
--- a/build_calicoctl/Dockerfile
+++ b/build_calicoctl/Dockerfile
@@ -10,7 +10,7 @@ RUN pip install -r requirements.txt
 
 # Can't run pyinstaller as root so add a user.
 RUN useradd -d /home/user -m -s /bin/bash user
-RUN chown -R user /code/
+RUN chown -R user:user /code/
 USER user
 # The HOME env var should be set automatically, but for some reason it's not...
 ENV HOME /home/user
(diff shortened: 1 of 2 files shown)


tomdee:
LGTM - but before you merge, do you know why all the CIs are failing?

PR #153, opened 2015-05-29
Driver container

spikecurtis:


calico_containers/calicoctl.py
```diff
@@ -57,12 +57,12 @@
 from netaddr.core import AddrFormatError
 from prettytable import PrettyTable
 
-from calico_containers.adapter.datastore import (ETCD_AUTHORITY_ENV,
```
tomdee: Can calicoctl.py still be invoked from the command line? Does it require any fiddly pythonpath stuff?

calico_containers/driver/requirements.txt
```diff
@@ -0,0 +1,4 @@
+netaddr==0.7.14
```
tomdee: Nice! Is this really all we need now?

tomdee:
A few little things to check, but otherwise looks good.

plugin.py needs to be moved to an integration directory - the idea is that's we'll have a different directory for each integration - libnetwork, mesos, kubernetes, rkt etc...

calico_containers/calicoctl.py
Symmetric: LMK if this isn't true, I spent a bunch of time fiddling with python path issues in the build to get this working, so I'd like to know what went wrong if it did...

spikecurtis: I've modified it so that calicoctl.py imports `pycalico.datastore` dropping the reference to `calico_containers` in the module path, which is odd.

In order for pyinstaller to build `calicoctl` correctly I had to set the PYTHONPATH for the build. command.

PR #154, opened 2015-05-29
Run etcd outside the host containers.

alexaltair:


calico_containers/tests/st/test_base.py
```diff
@@ -20,3 +24,26 @@ def tearDown(self):
...
+                   advertise_client_urls="http://%s:2379" % self.ip,
+                   listen_client_urls="http://0.0.0.0:2379",
+                   initial_advertise_peer_urls="http://%s:2380" % self.ip,
+                   listen_peer_urls="http://0.0.0.0:2380",
+                   initial_cluster_token="etcd-cluster-2",
+                   initial_cluster="calico=http://%s:2380" % self.ip,
+                   initial_cluster_state="new",
+                  )
+
+    def get_ip(self):
```
spikecurtis: This doesn't need to be a class method.  You can just change it to be a normal function.

calico_containers/tests/st/test_mainline.py
```diff
@@ -10,17 +10,17 @@ def run_mainline(self, ip1, ip2):
         Setup two endpoints on one host and check connectivity.
         """
         host = DockerHost('host')
-        host.start_etcd()
 
-        host_ip = docker.inspect("--format", "'{{ .NetworkSettings.IPAddress }}'", host.name).stdout.rstrip()
-        etcd_port = "ETCD_AUTHORITY=%s:2379" % host_ip
+        # etcd_ip = docker.inspect("--format", "{{ .NetworkSettings.IPAddress }}", "etcd").stdout.rstrip()
```
spikecurtis: Remove this if not needed.

calico_containers/tests/st/test_mainline.py
```diff
@@ -10,17 +10,17 @@ def run_mainline(self, ip1, ip2):
...
+        etcd_ip = self.ip
+        etcd_port = "ETCD_AUTHORITY=%s:2379" % etcd_ip
```
spikecurtis: This naming is confusing.  The etcd "port" is 2379.  The combination of a host and port is called an "authority."

calico_containers/tests/st/test_base.py
robbrockbank: just butting in here too.  not convinced about assuming anything about the host interfaces here.  An oft touted way to do this is along the lines of:

import socket
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.connect(('google.com', 0))
s.getsockname()[0]

not sure that's much better, but it doesn't assume eth0.

calico_containers/tests/st/test_multi_host.py
```diff
@@ -24,7 +24,7 @@ def test_multi_host(self):
         host1.listen(calicoctl % ("node --ip=%s" % host1_ip))
         host2.listen(calicoctl % ("node --ip=%s" % host2_ip))
 
-        calico_port = "DOCKER_HOST=localhost:2377"
+        calico_port = "export DOCKER_HOST=localhost:2377;"
```
spikecurtis: it's not necessary to export this if you're prefixing every command with it.

calico_containers/tests/st/test_base.py
spikecurtis: Assumes connectivity to 'google.com', which isn't great either.

alexaltair: I'm less crazy about using the outside world to figure out my IP than about assuming eth0. Does seem like it could cause a problem, but it works for now.

calico_containers/tests/st/test_multi_host.py
alexaltair: Actually, I was having problems with the `VAR=thing cmd` syntax, and this was my overkill way to fix it and stop worrying. But either way, I overhaul this in aa1-convert-tests, so I'd rather leave it as is here and move further discussion on it to that branch.

calico_containers/tests/st/test_base.py
robbrockbank: actually the example i pasted was prolly a bit rubbish since it uses a name rather than an IP so not sure what happens if DNS isn't working

I usually see this with a "fake" IP of "8.8.8.8" or similar - it doesn't actually need to be valid and so it's not actually talking to the outside world.

@alexaltair when you say it works for now - you mean it works for you because your environment uses eth0 or are you saying this should work for everyone?

alexaltair: It works on my dev machine and on jenkins.

calico_containers/tests/st/test_multi_host.py
spikecurtis: This is a merge to master, so we need to sort it out before merging.

calico_containers/tests/st/utils.py
```diff
@@ -0,0 +1,28 @@
+import sh
+from sh import docker
+
+
+def get_ip():
+    """Return a string of the IP of the hosts eth0 interface."""
+    intf = sh.ifconfig.eth0()
+    return sh.perl(intf, "-ne", 's/dr:(\S+)/print $1/e')
```
spikecurtis: I'd prefer Rob's suggestion here.  I'd rather depend on internet access than having an interface named eth0, which we know isn't the case for RedHat and other popular distros.

calico_containers/tests/st/test_base.py
Symmetric: Note that '8.8.8.8' isn't a fake IP, it's one of Google's DNS servers, so
it's probably one of the most durable IPs in existance for pings. That
durability is why the examples for this approach tend to use 8.8.8.8. (If
you're really paranoid you can fall back to 8.8.4.4, or even some other
cloud provider's HA endpoint).

spikecurtis:
LGTM

calico_containers/tests/st/docker_host.py
```diff
@@ -15,16 +17,21 @@ def __init__(self, name):
...
         self.execute("while ! docker ps; do sleep 1; done && "
-                   "docker load --input /code/calico-node.tar && "
-                   "docker load --input /code/busybox.tar && "
-                   "docker load --input /code/nsenter.tar")
+                     "docker load --input /code/calico_containers/calico-node.tar && "
+                     "docker load --input /code/calico_containers/busybox.tar && "
+                     "docker load --input /code/calico_containers/nsenter.tar")
 
-    def execute(self, command, **kwargs):
+    def execute(self, command, docker_host=False, **kwargs):
```
spikecurtis: Can you add a comment explaining what `docker_host` is for?

spikecurtis:
Looks good.

PR #155, opened 2015-05-29
Test build config for semaphoreci

alexaltair:


tomdee:
Can you do your semaphore explorations without any checkins please? You can do it all through their UI...

alexaltair:
I'm testing changes to the repo itself. I can do it without making a PR, if that's what you mean. What negative effects are you wanting to avoid?

tomdee:
Messing up I git history when there's no need to. I can't see any reason why you need a PR, so closing it.

PR #159, opened 2015-06-02
Start using gunicorn

tomdee:


node_filesystem/etc/service/calico-driver/run
```diff
@@ -1,3 +1,15 @@
...
+ROOT=/calico_containers
+PID=/var/run/gunicorn.pid
+LOG_DIR=/var/log/calico
+APP=docker_plugin:app
+
+if [ -f $PID ]; then rm $PID; fi
+
+exec $GUNICORN --chdir $ROOT --pid=$PID \
+-b unix:///usr/share/docker/plugins/calico.sock $APP \
+--access-logfile $LOG_DIR/plugin-access.log
```
spikecurtis: I assume this is gunicorn-speak, but what is "access" in this context?

tomdee: Nope, standard web server speak. Every web server has support for a log which contains all accesses. Gunicorn is logging in standard apache format.

spikecurtis: Nice.  Do we also get a separate log for the app internals?  Where is its filename specified?

spikecurtis:
LGTM

PR #160, opened 2015-06-02
Add test for multiple endpoints with the same IP.

alexaltair:


calico_containers/tests/st/test_duplicate_ips.py
```diff
@@ -0,0 +1,68 @@
...
+        Start two workloads with the same IP on different hosts. Make sure they
+        can be reached from all places even after one of them is deleted.
+        """
+        host1 = DockerHost('host1')
+        host2 = DockerHost('host2')
+        host3 = DockerHost('host3')
+
+        calicoctl = "/code/dist/calicoctl %s"
+
+        host1.execute(calicoctl % ("node --ip=%s" % host1.ip))
```
fasaxc: How about adding a `calicoctl()` method to the host?  Then you could do something like `host1.calicoctl("node", ip=host1.ip)` and get rid of all those string interpolations.

calico_containers/tests/st/test_duplicate_ips.py
```diff
@@ -0,0 +1,68 @@
...
+        host2.execute(calicoctl % "profile TEST_PROFILE member add workload2")
+        host2.execute(calicoctl % "profile TEST_PROFILE member add dup2")
+        host3.execute(calicoctl % "profile TEST_PROFILE member add workload3")
+
+        # Wait for the workload networking to converge.
+        ping = partial(host1.execute, "docker exec workload1 ping -c 4 192.168.1.4")
+        self.retry_until_success(ping, ex_class=ErrorReturnCode_1)
+
+        # Check for standard connectivity
+        host1.execute("docker exec workload1 ping -c 4 192.168.1.4")
```
fasaxc: Likewise, this could be simplified a lot by having `host1.assert_can_ping(...)`

fasaxc:
Looks pretty straightforward, which is a good sign.  I do think it's worth cleaning up the duplication at this stage.  I expect it'll pay back on the next test.

Seems like there could be some other useful tests that start with this topology.  Make sure you keep the topology setup common as you add more.

alexaltair:
My next branch is cleaning up the things you mentioned et al. Is this one good to go?

PR #161, opened 2015-06-02
Add single-host IPv6 test.

alexaltair:


calico_containers/tests/st/test_ipv6.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_ipv6(self):
+        """
+        Test mainline functionality with IPv6 addresses.
+        """
+        host = DockerHost('host')
+
+        host.execute("docker run --rm  -v `pwd`:/target jpetazzo/nsenter", _ok_code=[0, 1])
+
+        calicoctl = "/code/dist/calicoctl %s"
+        host.execute(calicoctl % "node --ip=127.0.0.1 --ip6=fd80:24e2:f998:72d6::1")
```
spikecurtis: `fd80:24e2:f998:72d6::1` is the same IP as you use for the container.

It'd be better to add an `ip6` field to DockerHost (like the `ip` field it currently has).  You can hard-code it for now, until you get multi-host working.

Also, use `host.ip` for the IPv4 address, rather than `127.0.0.1`

calico_containers/tests/st/test_ipv6.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertEqual(ip2, node2_ip)
+
+        node1_pid = host.execute("docker inspect --format '{{.State.Pid}}' node1").stdout.rstrip()
+        node2_pid = host.execute("docker inspect --format '{{.State.Pid}}' node2").stdout.rstrip()
+
+        ping = partial(host.execute, "./nsenter -t %s ping6 %s -c 1 -W 1" % (node1_pid, node2_ip))
+        self.retry_until_success(ping, ex_class=ErrorReturnCode)
+
+        # Check connectivity.
+        host.execute("./nsenter -t %s ping6 %s -c 1" % (node1_pid, node1_ip))
```
spikecurtis: See https://docs.python.org/2/library/itertools.html#itertools.product

alexaltair: That is awesome, but this is literally the minimum non-trivial product (2x2). Using it here wouldn't really save lines or clarity.

spikecurtis: Saves lines (2 instead of 4).

Maybe it's just me, but even at 2x2 I'd find it more readable to directly state (in python code) that we're doing a product expansion, rather than looking at the list and having to notice that the check is exhaustive.

calico_containers/tests/st/docker_host.py
```diff
@@ -19,6 +19,10 @@ def __init__(self, name):
 
         self.ip = docker.inspect("--format", "{{ .NetworkSettings.IPAddress }}",
                                  self.name).stdout.rstrip()
+        ip6 = docker.inspect("--format", "{{ .NetworkSettings.GlobalIPv6Address }}",
+                             self.name).stdout.rstrip()
+        self.ip6 = ip6 or "fd80:24e2:f998:72d6::1"
```
spikecurtis: Leave a `# TODO` comment that hardcoding is only temporary until we sort out multi-host for IPv6.

spikecurtis:
LGTM!

PR #162, opened 2015-06-02
Make incremental improvements to the ST test framework.

alexaltair:


fasaxc:
Looks like good cleanup to me.  I'd be tempted to make the execute/calicoctl commands more similatr to Popen or sh.  I.e. `host.calicoctl("node", ip="10.0.0.1")`, using `*args` or passing a list as the first argument.

This is test code so it's not a big deal but in production code, it's a big red flag to be passing strings down into the shell without escaping them and the cleanest way I know to do that is to pass lists of arguments down and have the escaping done at a low layer.

alexaltair:
Yeah, I'd really like something like that, but can't see how to do it without overengineering at this point. I'll try to keep it in mind as I change bits.

PR #165, opened 2015-06-02
Move to using svlogd for logs

tomdee:
Add --log-dir option to calicoctl

All services write to stdout/stderr and the logs are picked up my svlogd.

* This makes writing services easy (writing to stdout/err is a lower common denominator).
* This allows people who deploy calico-node to configure logging in a standard way across our components (to filter logs, change level or send logs to syslog)

I've also added a doc describing this docs/logging.md. I suggest starting there.


spikecurtis:
Does svlogd further tie us to runit?

Does it make it easier to hook into everyone's favorite new/cool logging services?

node_filesystem/etc/service/bird/log/run
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+LOGDIR=/var/log/calico/bird
+mkdir -p $LOGDIR
+# Prefix each line with a timestamp
```
spikecurtis: Doesn't bird already timestamp the log lines?

spikecurtis:
One comment, but otherwise LGTM

node_filesystem/etc/service/bird/log/run
tomdee: Not when it's just logging to stdout.

PR #166, opened 2015-06-02
Create Workload class and associated methods.

alexaltair:


calico_containers/tests/st/docker_host.py
```diff
@@ -79,3 +80,6 @@ def assert_powerstrip_up(self):
         """
         powerstrip = partial(self.execute, "docker ps", use_powerstrip=True)
         retry_until_success(powerstrip, ex_class=ErrorReturnCode)
+
+    def create_workload(*args, **kwargs):
```
spikecurtis: Writing a function with arguments like this is basically saying "I'm not telling you anything about how to use this function!"

I get that in this specific case, the constructor for Workload essentially defines the arguments, but I doubt python IDEs are smart enough to figure that out.  I for one use auto-completion all the time.

calico_containers/tests/st/test_add_ip.py
```diff
@@ -13,7 +13,7 @@ def test_add_ip(self):
         """
         host = DockerHost('host')
 
-        host.execute("docker run -e CALICO_IP=192.168.1.1 -tid --name=node1 busybox", use_powerstrip=True)
+        host.create_workload("node1", "192.168.1.1")
         host.execute("docker run -tid --name=node2 busybox")
```
spikecurtis: Feels like this style of creating a workload should also be accommodated in `create_workload`

spikecurtis:
I get that you're working incrementally, but it would be helpful to know your long term plan so I can comment on it.  It's hard to know if I agree with the path we're taking if I can only see one step at a time.  Also, it's hard to spot whether the actual increment I'm reviewing reinforces or undermines the long long term plan.

calico_containers/tests/st/test_base.py
```diff
@@ -46,3 +50,28 @@ def start_etcd(self):
             initial_cluster="calico=http://%s:2380" % self.ip,
             initial_cluster_state="new",
         )
+
+    def assert_can_ping(self, workload, ip, retries=5):
```
spikecurtis: I'd rather have a default of 1 (or maybe 2) as the retry.  In places where the tester knows the network needs time to converge, they should set a higher value.

1 out of 5 packets successful just seems like too low of a default bar to pass and could mask route flapping or other "flaky network" issues.  The underlying network should be rock-solid at this level of utilization, so flakiness is very likely to have been caused by us.

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,90 @@
...
+        assert version in [4, 6]
+        if version == 4:
+            ping = "ping"
+        elif version == 6:
+            ping = "ping6"
+
+        args = [
+            ping,
+            "-c", "1",  # Number of pings
+            "-W", "2",  # Timeout for each ping
```
spikecurtis: I'd like to find ways to shave time off ST runs, and 2 seconds for a ping feels very pessimistic.  Let's go for 1 second.

calico_containers/tests/st/test_duplicate_ips.py
```diff
@@ -17,44 +17,39 @@ def test_duplicate_ips(self):
...
+        host1.calicoctl("profile TEST_PROFILE member add %s" % dup1)
+        host2.calicoctl("profile TEST_PROFILE member add %s" % workload2)
+        host2.calicoctl("profile TEST_PROFILE member add %s" % dup2)
+        host3.calicoctl("profile TEST_PROFILE member add %s" % workload3)
 
         # Check for standard connectivity
-        host1.execute("docker exec workload1 ping -c 4 192.168.1.4")
-        host2.execute("docker exec workload2 ping -c 4 192.168.1.4")
-        host3.execute("docker exec workload3 ping -c 4 192.168.1.4")
+        ping = partial(workload1.ping, dup_ip)
```
spikecurtis: Seems there are two ways to do ping checks: `Workload.ping` and `TestBase.assert_can_ping`  I'd like there to be just one (my preference is `Workload.ping`)

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,90 @@
...
+        if ip and ip != 'auto':
+            assert ip == self.ip, "IP param = %s, configured IP = %s." % (ip, self.ip)
+
+    def execute(self, command, **kwargs):
+        """
+        Execute arbitrary commands on this workload.
+        """
+        return self.host.execute("docker exec %s %s" % (self.name, command))
+
+    def ping(self, ip):
```
spikecurtis: This should be called `assert_can_ping` --- it would also be nice if we raised an Exception that gave the failure in a succinct statement like "workload %s cannot ping IP %s".  Much nicer for debugging system tests.

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,90 @@
...
+
+        args = [
+            ping,
+            "-c", "1",  # Number of pings
+            "-W", "2",  # Timeout for each ping
+            ip,
+        ]
+        command = ' '.join(args)
+        return self.execute(command)
+
```
spikecurtis: add an `assert_cannot_ping` for test readability.  Much clearer than `assertRaises` constructs.

calico_containers/tests/st/workload.py
alexaltair: Sounds reasonable. Do we still have any cases where the first ping can take over a second?

I agree about saving time in the tests, though I suspect the effect of this parameter is far negligible compared to docker setup stuff.

spikecurtis: That may be true for current tests, but a full connectivity check is quadratic in the number of endpoints.  It particularly matters for *negative* checks of connectivity.

Also, we have sorted out the issues with initial pings taking a long time.  I'd consider those reemerging a regression.

calico_containers/tests/st/test_duplicate_ips.py
alexaltair: Yeah, I'm fairly unsure about this.

The idea is to have semantic separation. `Workload.ping` is an action that a `Workload` can take, which is just executing the `ping` command through all the docker stuff. `TestBase.assert_can_ping` is intended to be true to the form of other `TestBase` assertions, as in "it is true that pinging can happen".

The difference practically is the retries. It seems plausible that there could grow to be other differences, like there did in the OpenStack tests.

calico_containers/tests/st/workload.py
alexaltair: Agreed on the exception message. That's the kind of thing that feels to me like it justifies separating the action from the assertion.

calico_containers/tests/st/test_duplicate_ips.py
spikecurtis: I get what you're saying, but in an ST environment a ping is (pretty much?) always a test, not an action to take where the side effect is import.

Executing the ping command already acts like a test in that it throws an exception if it fails.

If the difference is in retries (and you think that distinction is important enough to keep) then you need to name the methods so that difference is clear.

I don't think that the plausibility of other differences is reason for us to take action now.  We can always change things later if it becomes necessary.

calico_containers/tests/st/test_add_ip.py
```diff
@@ -13,54 +13,60 @@ def test_add_ip(self):
...
+        host.calicoctl("container add %s %s --interface=hello" % (node2, ip12))
 
         host.calicoctl("profile add TEST_GROUP")
-        host.calicoctl("profile TEST_GROUP member add node1")
-        host.calicoctl("profile TEST_GROUP member add node2")
+        host.calicoctl("profile TEST_GROUP member add %s" % node1)
+        host.calicoctl("profile TEST_GROUP member add %s" % node2)
 
-        test_ping = partial(host.execute, "docker exec node1 ping 192.168.1.2 -c 1 -W 1")
+        test_ping = partial(node1.assert_can_ping, ip12)
```
spikecurtis: no need for partials here and elsewhere, since `assert_can_ping` has that function built-in.

spikecurtis:
Just one comment on consolidation of retry logic.  Otherwise LGTM.

PR #171, opened 2015-06-03
Some test tweaks

tomdee:
Most important one is moving some of the compile/build steps out of the nosetests command and into run_sts.sh


run_sts.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
+./build_node.sh
```
spikecurtis: Can we avoid rebuilding the node if it's already built?

PR #173, opened 2015-06-04
Pythonize cleanup to catch more errors.

alexaltair:


calico_containers/tests/st/utils.py
```diff
@@ -17,14 +17,22 @@ def cleanup_inside(name):
...
-          )
+    try:
+        inner_containers = docker("exec", "-t", name, "bash", "-c", "docker ps -qa").stdout.split()
+        for container in inner_containers:
+            docker("exec", "-t", name, "bash", "-c", "docker rm -f %s" % container)
+
+        inner_images = docker("exec", "-t", name, "bash", "-c", "docker images -q").stdout.split()
+        for image in inner_images:
+            docker("exec", "-t", name, "bash", "-c", "docker rmi %s" % image)
+    except ErrorReturnCode_127:
```
fasaxc: exception handling needs to be inside the for loops or you'll skip the rest of the cleanup if one rm fails.

alexaltair: @fasaxc Yeah, I wondered whether I should do that. Both caught exceptions are from cases that happen on the first line (one comment is wrong). If an `rm` or `rmi` fails, the exception won't be caught, which is what I'm looking for; then I can go investigate why that happened. It *could* be the case that a `127` or `255` might be raised by some other means, which will then be caught and fail to cleanup the rest of the loop. I feel this is unlikely given that all the commands are so similar. Restructuring it to guarantee catching all also makes it a fair bit messier. Do you think I'm taking the wrong side of the trade-off?

fasaxc: Ah, gotcha. I'd catch the errors only for the first call.  You can also catch multiple exceptions in one except line:
```
try:
    ...
except (ErrorReturnCode_127, ErrorReturnCode_255):
    # These errors come from bash, indicate docker is not present.
   return
```
You may want to check for other cases where commands return non-0 even though they succeeded.  A common one is if you run a "list" command but there's nothing to list (always annoys me when they do that, an empty list is not the same as a failure to build the list!).

calico_containers/tests/st/utils.py
```diff
@@ -38,6 +24,40 @@ def delete_container(name):
...
+    else:
+        for container in inner_containers:
+            docker("exec", "-t", name, "bash", "-c", "docker rm -f %s" % container)
+
+
+def cleanup_inner_images(name):
+    try:
+        inner_images = docker("exec", "-t", name, "bash", "-c", "docker images -q").stdout.split()
+    except (ErrorReturnCode_127, ErrorReturnCode_255):
+        # 127 caused by '"docker": no command found'
```
fasaxc: One last thought: can both of these really occur?  Seems odd that we'd get different errors on different runs.

alexaltair: Good point... well, I can't remember where I saw `ErrorReturnCode_127`. It must have been in a container that was neither `etcd` nor `dind`. In that case, we could 1) ruthlessly delete it or 2) fail to delete it, letting us know of the unexpected case.

PR #174, opened 2015-06-05
Add a multi-host IPv6 test.

alexaltair:


calico_containers/tests/st/docker_host.py
```diff
@@ -67,11 +66,19 @@ def execute(self, command, use_powerstrip=False, **kwargs):
...
         calicoctl = "/code/dist/calicoctl %s"
         return self.execute(calicoctl % command, **kwargs)
 
     def start_calico_node(self, ip=None, ip6=None):
+        """
+        Start calico in a container inside a host by calling through to the
+        calicoctl node command.
+        """
         ip = ip or self.ip
+        ip6 = ip6 or self.ip6
```
spikecurtis: I think we should drop the `ip` and `ip6` arguments and just this method always use the detected IP addresses.

The only place where we're actually passing other IPs is in the arguments parsing test, and I'd argue that test would be easier to understand if it called `host.calicoctl` directly with the arguments under test.

A "would be nice" feature is a way to disable IPv6 support, even if the host container got an IPv6 address, since we do want to test that things don't break if IPv6 isn't available.  That would make it easier to script IPv4 tests that run with and without IPv6 turned on.

calico_containers/tests/st/test_ipv6_multi_host.py
```diff
@@ -0,0 +1,50 @@
...
+        host1 = DockerHost('host1')
+        host2 = DockerHost('host2')
+
+        ip1 = "fd80:24e2:f998:72d6::1:1"
+        ip2 = "fd80:24e2:f998:72d6::1:2"
+        ip3 = "fd80:24e2:f998:72d6::1:3"
+        ip4 = "fd80:24e2:f998:72d6::1:4"
+        ip5 = "fd80:24e2:f998:72d6::1:5"
+
+        workload1 = host1.create_workload("workload1", ip1, image="phusion/baseimage:0.9.16")
```
spikecurtis: Can you add a comment about the choice of image.  E.g. that `busybox` doesn't have IPv6 version of `ping`

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,100 @@
...
+        ]
+        if ip:
+            # Powerstrip passes the CALICO_IP var to calico.
+            args += ["--env", "CALICO_IP=%s" % ip]
+        args.append(image)
+        command = ' '.join(args)
+
+        host.execute(command, use_powerstrip=use_powerstrip)
+
+        # There is an unofficial ip=auto option in addition to ip=None.
```
spikecurtis: What does "unofficial" mean in this context?  If it's just that we haven't documented it I'm not sure how that matters in ST---we still care to know it if regresses.

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,100 @@
...
+    def __init__(self, host, name, ip=None, image="busybox", use_powerstrip=True):
+        """
+        Create the workload and detect its IPs.
+
+        :param host: The host container on which this workload is instantiated.
+        All commands executed by this container will be passed through the host
+        via docker exec.
+        :param name: The name given to the workload container. This name is
+        passed to docker and can be used inside docker commands.
+        :param ip: The IP to be assigned to this workload via calico. May be
```
spikecurtis: It would be really helpful to be more explicit about what is allowed for this parameter.  I think it's "auto", None, or a string representing the IP address.

calico_containers/tests/st/workload.py
```diff
@@ -0,0 +1,100 @@
...
+
+        host.execute(command, use_powerstrip=use_powerstrip)
+
+        # There is an unofficial ip=auto option in addition to ip=None.
+        if ip is None or ip == 'auto':
+            version = None
+        else:
+            version = IPAddress(ip).version
+
+        # if version == 6:
```
spikecurtis: If you're going to leave this commented out code, then you need a comment explaining why it's being worked around and what has to happen before we can remove the workaround.

calico_containers/tests/st/workload.py
alexaltair: Yeah, I'm just explaining why `ip == 'auto'` is in the condition. We still test it.

calico_containers/tests/st/docker_host.py
```diff
@@ -82,10 +80,9 @@ def start_calico_node(self, ip=None, ip6=None):
         Start calico in a container inside a host by calling through to the
         calicoctl node command.
         """
-        ip = ip or self.ip
-        args = ['node', '--ip=%s' % ip]
-        if ip6:
-            args.append('--ip6=%s' % ip6)
+        args = ['node', '--ip=%s' % self.ip]
```
spikecurtis: You should drop the arguments from the def above.

spikecurtis:
One last comment, then LGTM!

PR #175, opened 2015-06-05
Fixes #150, degrades fatal system dep errors to warnings

### LINKED ISSUE #150: calicoctl fails to work when kernel has builtin ip6_tables
I use Linode VPS as my playground; running Ubuntu 15.04 x64. Linode builds the kernels with all things compiled in- e.g., ip6_tables is "builtin" in my case.

```
root at evilsec in ~ # ./calicoctl node --ip='173.255.224.31'
module ip6_tables isn't loaded. Load with `modprobe ip6_tables`
```

djosborne:
Fixes #150, calicoctl now only prints *warnings* for detected issues concerning system dependencies during node creation, but will not cancel the process if it detects them. 

For now, if our system check incorrectly detects issues, the user will just have to read past those warnings knowing that their system is in fact OK. We will have to work towards better detection across host platforms in the future so users won't be bothered with any incorrect messages (see #150).

Additionally, added the `calicoctl checksystem [--fix]` command which will print any detected issues, and will attempt to fix them if the `--fix` flag is provided.

### DIFF
--- a/calico_containers/calicoctl.py
+++ b/calico_containers/calicoctl.py
@@ -282,14 +282,12 @@ def node(ip, node_image, ip6=""):
     enforce_root()
 
     if not module_loaded("ip6_tables"):
-        print >> sys.stderr, "module ip6_tables isn't loaded. Load with " \
+        print >> sys.stderr, "WARNING: calico was unable to detect the ip6_tables module. Load with " \
                              "`modprobe ip6_tables`"
-        sys.exit(2)
 
     if not module_loaded("xt_set"):
-        print >> sys.stderr, "module xt_set isn't loaded. Load with " \
+        print >> sys.stderr, "WARNING: calico was unable to detect the xt_set module. Load with " \
                              "`modprobe xt_set`"
-        sys.exit(2)
 
     # Set up etcd
     ipv4_pools = client.get_ip_pools("v4")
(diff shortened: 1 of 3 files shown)


calico_containers/calicoctl.py
```diff
@@ -388,6 +374,70 @@ def node(ip, node_image, ip6=""):
     print "Calico node is running with id: %s" % cid
 
 
+def checksystem(fix=False, quit_if_error=False):
+    """
+    Checks that the system is setup correctly. If --fix is passed in at command line, this command
+    will attempt to fix any issues it encounters. If any fixes fail, it will exit(1)
+
```
tomdee: Document the parameters using :param

calico_containers/calicoctl.py
```diff
@@ -388,6 +374,70 @@ def node(ip, node_image, ip6=""):
...
+        enforce_root()
+
+    modprobe = sh.Command._create('modprobe')
+    system_ok = True
+    if not module_loaded("ip6_tables"):
+        if fix:
+            try:
+                modprobe('ip6_tables')
+            except:
+                print >> sys.stderr, "ERROR: calico could not enable ip6_tables."
```
tomdee: I think we should drop "calico" here.
Just write
ERROR: Could no enable ip6_tables
and
WARNING: Unable to detect the ip6_tables module. Load with...

tomdee:
Couple of comments. If you fix those, then it LGTM and I'm happy for you to merge.

tomdee:
Actually, you should also update the docs. At a minimum, there's the usage string in the main readme.

djosborne:
Turns out the ip6tables command automatically loads the ip6_tables module if it is not currently loaded. This seems like a cleaner way to actually check if the system supports ip6tables, and more accurately resolves the original issue.

calico_containers/calicoctl.py
```diff
@@ -278,18 +279,8 @@ def module_loaded(module):
 
 
 def node(ip, node_image, ip6=""):
-    # modprobe and sysctl require root privileges.
-    enforce_root()
```
djosborne: @tomdee Do you know offhand if this *only* necessary for modprobe and sysctl?

tomdee: Yes, I believe so.

PR #179, opened 2015-06-08
Remove device routes from BIRD kernel config.

spikecurtis:
The device routes switch tells BIRD to export device routes to the kernel.  This isn't required as the underlying network configuration has already set up device routes within the kernel leading to "File Exists" spam in the BIRD log.


spikecurtis:
CC @alexaltair

robbrockbank:
As discussed, already merged, but for completeness:  LGTM

PR #181, opened 2015-06-08
Stop setting the endpoint name since we can't know it.

tomdee:
Make the endpoint an optional thing, defaulted to None.

Endpoint is broken. We don't know the interface name (libnetwork
decides it) so it shouldn't be a mandatory parameter.


robbrockbank:
LGTM

Issue #169, opened 2015-06-03
New builds don't create calicoctl.sh


djosborne:
Possibly a result from commit 078bce883b03d2c3851ae2a41b79c5610b1853c2, running create_binary.sh only creates a new binary, but doesn't produce a new calicoctl.sh

robbrockbank:
Fixes by #186

PR #186, opened 2015-06-09
cd to dist directory before running docopt-completion

robbrockbank:
docopt-completion writes file to pwd, so cd before running command so output is in the dist directory.


robbrockbank:
Fixes #169

spikecurtis:
LGTM!

PR #190, opened 2015-06-09
Fixed test_arg_parsing.py ST.

spikecurtis:
Also, various refactoring ST to use subprocess rather than sh for host.exec.


tomdee:
LGTM

PR #193, opened 2015-06-09
Make DockerHost context manager that cleans up.

spikecurtis:
DockerHost becomes a context manager so you can use it like

```
with DockerHost('host') as host:
	# Do the stuff.
```

Fixes up existing STs to use this form.

### DIFF
--- a/calico_containers/tests/st/docker_host.py
+++ b/calico_containers/tests/st/docker_host.py
@@ -4,18 +4,23 @@ from functools import partial
 from subprocess import check_output, CalledProcessError
 from calico_containers.tests.st import utils
 
-from utils import get_ip, delete_container, retry_until_success
+from utils import get_ip, retry_until_success
 
 
 CALICO_DRIVER_SOCK = "/usr/share/docker/plugins/calico.sock"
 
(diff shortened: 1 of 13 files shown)


calico_containers/tests/st/docker_host.py
```diff
@@ -85,3 +84,58 @@ def assert_driver_up(self):
...
+    def remove_containers(self):
+        """
+        Remove all containers running on this host.
+
+        Useful for test shut down to ensure the host is cleaned up.
+        :return: None
+        """
+        # TODO: only remove ST created containers for non-dind.
+        cmd = "docker rm -f $(docker ps -qa) ; docker rmi $(docker images -qa)"
+        ok_codes = [0,
```
tomdee: We shouldn't need to catch all these code surely?

tomdee:
LGTM

calico_containers/tests/st/docker_host.py
spikecurtis: Erm, probably.  But, this code is pythonized in master, so I'd like to sort it out after merging to master.

PR #195, opened 2015-06-10
Fix 117, refactor calicoctl endpoint commands

djosborne:


calico_containers/adapter/datastore.py
```diff
@@ -716,6 +792,51 @@ def get_ep_id_from_cont(self, hostname, container_id):
                 "Container with ID %s has no endpoints." % container_id)
 
     @handle_errors
+    def query_endpoint(self, endpoint_id=None, workload_id=None, orchestrator_id=None, hostname=None):
```
djosborne: This function is essentially a more functional `get_endpoints` which doesn't require that all parameters are provided. It behaves fine when used by the new `endpoint show` command, other parts of calicoctl which rely on the current `get_endpoints` command break when they are mmoved over to this new one. I need to investigate further how this is different from `get_endpoints`

calico_containers/adapter/netns.py
```diff
@@ -192,7 +192,7 @@ def set_up_endpoint(ip, cpid, next_hop_ips,
 
     # Return an Endpoint.
     network = IPNetwork(IPAddress(ip))
-    ep = Endpoint(ep_id=ep_id, state="active", mac=mac, if_name=veth_name)
+    ep = Endpoint(hostname, "docker", cpid, endpoint_id, "active", mac)
```
djosborne: This needs work. Now that endpoint objects need hostname, orchestrator_id, and state, this function needs to pass them in. Is this script *always* run by docker, so the orchestrator ID is set here? Since its creating a new node, is its state active? Or does it need to be provided?

calico_containers/adapter/netns.py
```diff
@@ -217,11 +217,12 @@ def reinstate_endpoint(cpid, old_endpoint, next_hop_ips,
     if_name = old_endpoint.if_name
     net = nets.pop()
     new_endpoint = set_up_endpoint(ip=net.ip,
+                                   hostname=HOSTNAME,
```
djosborne: The running system is the appropriate hostname, right?

calico_containers/calicoctl.py
```diff
@@ -460,30 +469,37 @@ def profile_add(profile_name):
         print "Created profile %s" % profile_name
 
 
-def profile_add_container(container_name, profile_name):
+def profile_add_container(container_name, profile_name, workload_id="docker"):
```
djosborne: Is there docker hardcoding okay here?

calico_containers/calicoctl_profile.py
```diff
@@ -0,0 +1 @@
+__author__ = 'dano'
```
djosborne: Oops

djosborne:
@robbrockbank Haven't found any bugs with my manual testing.

calico_containers/adapter/datastore.py
```diff
@@ -228,17 +249,44 @@ def from_json(cls, ep_id, json_str):
...
+        ep.profile_ids = [profile_id] if profile_id else \
+                         json_dict.get("profile_ids", [])
+        ep.if_name = json_dict.get("container:if_name", VETH_NAME)
+
+        # Store the original JSON representation of this Endpoint.
+        ep._original_json = json_str
+
         return ep
 
+    def matches(self, hostname=None, orchestrator_id=None, workload_id=None, endpoint_id=None):
```
robbrockbank: Line length should be max 79 chars plus newline.

calico_containers/adapter/datastore.py
robbrockbank: the original function returns a dictionary of values where the keys provide the hostname and workdload id - perhaps something to do with that.

calico_containers/adapter/datastore.py
```diff
@@ -716,6 +792,51 @@ def get_ep_id_from_cont(self, hostname, container_id):
...
+        :param hostname: The hostname that the endpoint lives on.
+        :param workload_id: The workload (or container) that the endpoint belongs to.
+        :param orchestrator_id: The workload (or container) that the endpoint belongs to.
+
+        :return: A list of Endpoint Objects which match the criteria, or an empty list if none match
+        """
+        # First build the query string as specific as possible. Note, we want the query to be as specific as
+        # possible, so we proceed any variables with known constants e.g. we add '/workload' after the hostname
+        # variable.
+        ep_path = "/calico/v1/host"
```
robbrockbank: We should avoid having these defined in line.  Perhaps rather than string concatenation use the appropriate key template based on what parms you have.

calico_containers/adapter/datastore.py
```diff
@@ -716,6 +792,51 @@ def get_ep_id_from_cont(self, hostname, container_id):
...
+                ep_path += "/%s" % orchestrator_id
+                if workload_id:
+                    ep_path += "/%s/endpoint" % workload_id
+                    if endpoint_id:
+                        ep_path += "/%s" % endpoint_id
+        try:
+            # Search etcd
+            leaves = self.etcd_client.read(ep_path, recursive=True).leaves
+
+            # Filter through result
```
robbrockbank: move this outside the try.

in general it's good practise to minimize the code inside the try block to just that code that will actually raise the exceptions you are catching.

calico_containers/adapter/datastore.py
```diff
@@ -228,17 +249,44 @@ def from_json(cls, ep_id, json_str):
...
+        ep.if_name = json_dict.get("container:if_name", VETH_NAME)
+
+        # Store the original JSON representation of this Endpoint.
+        ep._original_json = json_str
+
         return ep
 
+    def matches(self, hostname=None, orchestrator_id=None, workload_id=None, endpoint_id=None):
+        """
+        A less strict 'equals' function, which compares designated parameters to the current endpoint object.
```
robbrockbank: can be more specific than designated parameters.  this function compares the IDs that fully qualify (or uniquely identify) a particular endpoint

calico_containers/adapter/datastore.py
```diff
@@ -624,12 +669,9 @@ def get_profile_members(self, profile_name):
...
             for child in endpoints:
-                packed = child.key.split("/")
-                if len(packed) == 10:
-                    (_, _, _, _, host, _, ctype, cid, _, ep_id) = packed
-                    ep = Endpoint.from_json(ep_id, child.value)
-                    if ep.profile_id == profile_name:
-                        eps[host][ctype][cid][ep_id] = ep
+                ep = Endpoint.from_json(child.key, child.value)
+                if ep and profile_name in ep.profile_ids:
+                    eps[ep.hostname][ep.orchestrator_id][ep.workload_id][ep.endpoint_id] = ep
```
robbrockbank: Would be nice for this to just return a list rather than a nested dictionary and remote the get_profile_members_endpoint_ids method.  However - I don't suggest we do that just yet.  I'll create a google doc to keep track of future improvements.

calico_containers/calicoctl.py
robbrockbank: workload_id is not correct.  you mean orchestrator id.

calico_containers/calicoctl.py
```diff
@@ -460,30 +469,37 @@ def profile_add(profile_name):
...
-        print "Added %s to %s" % (container_name, profile_name)
+        endpoint_id = client.get_endpoint_id_from_cont(hostname, container_id)
     except KeyError:
-        print "Failed to add container to profile.\n"
+        print "Failed to container to profile.\n"
         print_container_not_in_calico_msg(container_name)
         sys.exit(1)
 
+    # Call through to the container profile set method to add the profile.
+    endpoint_profile_append(endpoint_id, container_name, workload_id, hostname, [profile_name])
```
robbrockbank: should be passing in ep_id, wl_id, orch_id, hostn

calico_containers/calicoctl.py
```diff
@@ -991,13 +1003,181 @@ def container_ip_remove(container_name, ip, version, interface):
 
     print "IP %s removed from %s" % (ip, container_name)
 
+def endpoint_show(endpoint_id, workload_id, orchestrator_id, hostname, detailed):
```
robbrockbank: missing doc string

calico_containers/calicoctl.py
```diff
@@ -991,13 +1003,181 @@ def container_ip_remove(container_name, ip, version, interface):
...
+                       ','.join(endpoint.profile_ids),
+                       endpoint.state])
+    else:
+        x = PrettyTable(["Hostname", "Orchestrator ID", "NumWorkloads", "NumEndpoints"])
+        for endpoint in endpoints:
+            # TODO: remove duplicates rows
+            num_workloads = len([other_endpoint for other_endpoint in endpoints if other_endpoint.workload_id == endpoint.workload_id])
+            num_endpoints = len([other_endpoint for other_endpoint in endpoints if other_endpoint.endpoint_id == endpoint.endpoint_id])
+            x.add_row([endpoint.hostname, endpoint.orchestrator_id, num_workloads, num_endpoints])
+    print str(x) + "\n"
```
robbrockbank: need to make sure the tables are ordered

calico_containers/calicoctl.py
```diff
@@ -991,13 +1003,181 @@ def container_ip_remove(container_name, ip, version, interface):
...
+    :param profile_names: The list of profiles to append.
+
+    :return: None
+    """
+    # Validate the profile list.
+    validate_profile_list(profile_names)
+
+    try:
+        # TODO: Use function parameters to tie-brake endpoints with same endpoint_id
+        client.set_profiles_on_endpoint(endpoint_id, profile_names)
```
robbrockbank: we should be passing in all the parameters now.  this is currently inefficient.

calico_containers/calicoctl.py
```diff
@@ -991,13 +1003,181 @@ def container_ip_remove(container_name, ip, version, interface):
...
+    :param endpoint_id: The endpoint ID.
+    :param profile_names: The list of profiles to append.
+
+    :return: None
+    """
+    # Validate the profile list.
+    validate_profile_list(profile_names)
+
+    try:
+        client.remove_profiles_from_endpoint(endpoint_id, profile_names)
```
robbrockbank: ditto - we should be passing all parms across.

calico_containers/calicoctl.py
```diff
@@ -1127,6 +1306,39 @@ def print_paragraph(msg):
...
+                                            arguments['<PROFILES>'] or [])
+                elif arguments["remove"]:
+                    endpoint_profile_remove(arguments["<ENDPOINT_ID>"],
+                                            arguments["--workload"],
+                                            arguments["--orchestrator"],
+                                            arguments["--host"],
+                                            arguments['<PROFILES>'] or [])
+                elif arguments["set"]:
+                    endpoint_profile_set(arguments["<ENDPOINT_ID>"],
+                                          arguments["--workload"],
```
robbrockbank: argument alignment out

calico_containers/calicoctl.py
```diff
@@ -1193,7 +1405,14 @@ def print_paragraph(msg):
                                         arguments["<IP>"],
                                         ip_version,
                                         arguments["--interface"])
-            elif arguments["container"]:
+                else:
+                    if arguments["add"]:
```
robbrockbank: You can't actually hit this block of code can you?

calico_containers/adapter/datastore.py
```diff
@@ -746,32 +867,31 @@ def set_endpoint(self, hostname, container_id, endpoint):
...
-        :param container_id: The Docker container ID.
-        :param old_endpoint: The existing endpoint to update.
-        :param new_endpoint: The Endpoint to add to the container.
+        :param endpoint: The Endpoint to add to the container.
         """
-        ep_path = ENDPOINT_PATH % {"hostname": hostname,
-                                   "container_id": container_id,
-                                   "endpoint_id": new_endpoint.ep_id}
+        ep_path = ENDPOINT_PATH % {"hostname": endpoint.hostname,
+                                   "container_id": endpoint.workload_id,
```
robbrockbank: Can we rename the container_id in these string templates to be workload_id ... (if its more that 15 minutes work then just add to the google doc to do list)

calico_containers/adapter/datastore.py
```diff
@@ -794,12 +914,28 @@ def get_endpoints(self, hostname, container_id):
...
         eps = []
         for endpoint in endpoints:
-            (_, _, _, _, _, _, _, _, _, endpoint_id) = \
-                                                 endpoint.key.split("/", 9)
-            eps.append(Endpoint.from_json(endpoint_id, endpoint.value))
+            eps.append(Endpoint.from_json(endpoint.key, endpoint.value))
         return eps
 
     @handle_errors
+    def get_endpoint_from_id(self, endpoint_id):
```
robbrockbank: This should really just use query_endpoint.

Raise an Exception if >1 endpoint matching the endpoint ID.

calico_containers/adapter/netns.py
robbrockbank: I think for now hardcoding to docker will be fine.  However, add a TODO comment, and add an item to the google doc so that we don't forget to fix this up.

This will probably be changed by Toms libnetwork stuff too.

calico_containers/adapter/netns.py
robbrockbank: should we just be passing in the hostname into this function, and using that.  It'll all amount to the same thing I believe, but we'd push the hostname determination out the powerstrip which invokes this function.

calico_containers/adapter/datastore.py
```diff
@@ -658,38 +700,72 @@ def profile_update_rules(self, profile):
         self.etcd_client.write(rules_path, profile.rules.to_json())
 
     @handle_errors
-    def add_workload_to_profile(self, hostname, profile_name, container_id):
+    def append_profiles_to_endpoint(self, endpoint_id, profile_names):
```
robbrockbank: these methods need to take the full set of IDs and handle there being multiple endpoints (error)

calico_containers/adapter/datastore.py
robbrockbank: Actually I don't really think this method should exist.

calico_containers/adapter/datastore.py
```diff
@@ -794,12 +914,28 @@ def get_endpoints(self, hostname, container_id):
...
+        # Extract all of the endpoints.
+        for leaf in leaves:
+            ep = Endpoint.from_json(leaf.key, leaf.value)
+            if ep and ep.endpoint_id == endpoint_id:
+                return ep
+        raise KeyError("Endpoint with ID %s was not found." %
+                       endpoint_id)
+
+    @handle_errors
     def get_hosts(self):
```
robbrockbank: WE should retire this.  It's badly named (it's actually getting the endpoints) and the function is effectively duplicated by query_endpoints.

robbrockbank: Not suggesting we do this now - but can you add to the google doc list.

calico_containers/adapter/datastore.py
djosborne: Are you just suggesting rewording the docstring?

calico_containers/calicoctl.py
```diff
@@ -516,30 +532,37 @@ def profile_add(profile_name):
...
         sys.exit(1)
 
+    # Check that the container is already networked
     try:
-        client.add_workload_to_profile(hostname, profile_name, container_id)
-        print "Added %s to %s" % (container_name, profile_name)
+        endpoint_id = client.get_endpoint_id_from_cont(hostname, container_id)
     except KeyError:
-        print "Failed to add container to profile.\n"
+        print "Failed to container to profile.\n"
```
robbrockbank: log does not make sense

calico_containers/calicoctl.py
robbrockbank: you can get rid of workload_id now

PR #200, opened 2015-06-11
Smaller docker image

tomdee:
Reduce calico/node from ~500MB to ~250MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,50 +1,32 @@
-# For details and docs - see https://github.com/phusion/baseimage-docker#getting_started
-FROM phusion/baseimage:0.9.16
+FROM ubuntu:14.04
 
-# Use baseimage-docker's init system.
 CMD ["/sbin/my_init"]
 
-# Ensure UTF-8, required for add-apt-repository call.
-RUN locale-gen en_US.UTF-8
-ENV LANG       en_US.UTF-8
-ENV LC_ALL     en_US.UTF-8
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -1,50 +1,32 @@
...
-        python-etcd \
-        && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ADD calico_containers/pycalico/requirements.txt /pycalico/
 
-# Confd
-RUN curl -L https://www.github.com/kelseyhightower/confd/releases/download/v0.9.0/confd-0.9.0-linux-amd64 -o confd && \
-    chmod +x confd
+# Incremental approach used for dev.
```
spikecurtis: @tomdee Expand these comments to explain.  E.g. "uncomment these lines and comment the 'prod' lines if you are doing lots of repeated builds, since the early scripts take a while and don't change often."

image/base.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
spikecurtis: I think it would help with maintainability if you could, for each file we got from phusion, clearly mark which sections are lifted verbatim and which are your own customizations.

That way if we ever need to upgrade or take patches we'll have a fighting chance...

spikecurtis:
A couple suggestions for additional comments, but otherwise, LGTM.

PR #204, opened 2015-06-14
Allow environment variable to set local IP used in STs

robbrockbank:


robbrockbank:
Agreed with Spike that this is ok to go in (it's copied from Spikes branch - I reviewed)

PR #205, opened 2015-06-15
Add IP in IP support

robbrockbank:


calico_containers/adapter/datastore.py
```diff
@@ -365,12 +405,14 @@ def add_ip_pool(self, version, pool):
...
-        if pool in self.get_ip_pools(version):
-            return
+        key = IP_POOL_KEY % {"version": version,
+                             "pool": str(pool).replace("/", "-")}
 
-        pool_path = IP_POOL_PATH % {"version": version}
-        self.etcd_client.write(pool_path, str(pool), append=True)
+        data = {"cidr" : str(pool)}
+        if ipip:
+            data["ipip"] = "tunl0"
```
tomdee: Do we really want to use a hardcoded value here? tunl0 might conflict with existing tunnels on the host, could we at least make it something a little more calico specific?

robbrockbank: currently mandated by Felix - apparently its the default and is created automatically.  Shaun knows more about this.

node_filesystem/templates/felix.cfg.template
```diff
@@ -1,5 +1,6 @@
 [global]
 MetadataAddr  = None
+IpInIpEnabled = True
```
tomdee: Does this change mean that for calico-docker, we will always get an tunl0 device created? I don't think that's what we want...

robbrockbank: Agree, but leaving as is for MVP.  Will track this getting done at a later date.

PR #206, opened 2015-06-15
New calicoctl endpoint commands to manage profiles and view endpoints

robbrockbank:


calico_containers/calicoctl.py
```diff
@@ -25,6 +25,9 @@
   calicoctl container <CONTAINER> ip (add|remove) <IP> [--interface=<INTERFACE>]
   calicoctl container add <CONTAINER> <IP> [--interface=<INTERFACE>]
   calicoctl container remove <CONTAINER> [--force]
+  calicoctl endpoint show [--host=<HOSTNAME>] [--orchestrator=<ORCHESTRATOR_ID>] [--workload=<WORKLOAD_ID>] [--endpoint=<ENDPOINT_ID>] [--detailed]
```
tomdee: --options should be documented under Options:

calico_containers/calicoctl.py
```diff
@@ -225,22 +239,26 @@ def container_remove(container_name):
     then it is removed.
 
     :param container_name: The name or ID of the container.
+    :param orchestrator_id: If the orchestrator ID for the
```
tomdee: the what?

calico_containers/calicoctl.py
```diff
@@ -518,28 +540,38 @@ def profile_add(profile_name):
...
     :param container_name: The Docker container name or ID.
     :param profile_name:  The Calico policy profile name.
     :return: None.
     """
-    # Resolve the name to ID.
-    container_id = get_container_id(container_name)
+    info = get_container_info_or_exit(container_name)
+    container_id = info["Id"]
 
-    if not client.profile_exists(profile_name):
```
tomdee: Why are we getting rid of this check?

calico_containers/adapter/datastore.py
```diff
@@ -1,15 +1,18 @@
...
 from etcd import EtcdKeyNotFound, EtcdException
 from netaddr import IPNetwork, IPAddress, AddrFormatError
-import os
-import copy
 
 ETCD_AUTHORITY_DEFAULT = "127.0.0.1:4001"
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 
 # etcd paths for Calico
+# TODO: modify string constants to accept orchestrator_id, replacing "docker" hardcoding
```
tomdee: Can you expand this comment a bit. Just be clear that we're not cleaning this up just yet because not all of the methods in datastore.py are aware of the orchestrator ID yet.

calico_containers/calicoctl.py
djosborne: `get_container_info_or_exit` is now used to handle missing profiles

tomdee:
You need to update  def node_stop(force): so that it checks for endpoints on the current host (for any orchestrator)

tomdee:
can shownodes be removed? I think maybe it can...

tomdee:
I'm on the fence about whether endpoint profile show should exist. It seems like it's just duplicating function from endpoint show --endpoint-id=X. WDYT?

djosborne:
The original plan was to have `profile (append|remove|set|show)` all in one command. When we realized that you needed to specify at least an endpoint ID for `profile show`, we split it into a separate command. 

However, to use the `endpoint profile show` command, you need to specify the `endpoint_id`. To get the `endpoint_id`, you will have to run `endpoint show --detailed`, which will end up telling you the profiles on each endpoint anyways. So I would agree with Tom - I can't think of a use case where someone will only have the endpoint ID and will need to know the profile(s) assigned to it.They'll get both pieces of info from `endpoint show --detailed`

However, the other solution I see is not requiring `<ENDPOINT_ID>` on the `endpoint show` command. Then it will look merge all 4 endpoint profile commands back into one command. We can throw an error if they didn't provide enough endpoint filter flags to uniquely identify a target endpoint. 

@robbrockbank what do you think?

djosborne:
`shownodes` and `endpoint show` present the same information. We should remove one

tomdee:
What about adding a command like
calicoctl container getendpointid <CONTAINER>

Which returns the bare endpoint ID and throws an error if the container has more than one endpoint ID.

This means people can run
calicoctl endpoint $(calicoctl container getendpointid "mycontainername") profile append .....

djosborne:
At that point we should just add a --container-name=<CONTAINER_NAME> flag

tomdee:
That's all my comments. I think we need to resolve:
- Removing unneeded/deprecated commands
- providing an easy/scriptable way for people to get endpoint IDs from containers.
- fix up node_stop


And this PR is introducing tech debt that we need to resolve in future. Calicoctl and datastore.py are now getting huge. In the future we need to split and/or clean them

lxpollitt:
I have limited connectivity right now so have not suggested the whole thing, but in general my vote would be to avoid having to do multi command queries. So including --container-name or some other approach that doesn't require a separate commands to find an ID the user doesn't have to hand weeps hey my vote. (I think it is OK to have optional parameters that don't apply to all workload types by the way of it makes dealing with a particular workload type easier.)

djosborne:
@robbrockbank I just made the the latest commit, 6895a67, merging master into this branch. There was a few small conflicts from Shaun's rule branch, but nothing huge. If all tests pass, it should be ready to merge.

PR #209, opened 2015-06-16
Add support for manipulating rules.

fasaxc:


calico_containers/calicoctl.py
```diff
@@ -677,6 +690,62 @@ def profile_rule_update(name):
     print "Successfully updated rules on profile %s" % name
 
 
+def profile_rule_add_remove(
```
robbrockbank: doc string :-)

calico_containers/calicoctl.py
```diff
@@ -1105,6 +1187,39 @@ def print_paragraph(msg):
     print
 
 
+def parse_ports(ports_str):
+    """
+    Parse a a string representing a port list into a list of ports and
```
robbrockbank: a a

calico_containers/calicoctl.py
```diff
@@ -1105,6 +1187,39 @@ def print_paragraph(msg):
...
+        sys.exit(1)
+    splits = ports_str.split(",")
+    parsed_ports = []
+    for split in splits:
+        m = re.match(r'^(\d+)[:-](\d+)$', split)
+        if m:
+            # Got a range, canonicalise it.
+            parsed_ports.append("%s:%s" % (m.group(1), m.group(2)))
+        else:
+            # Shuold be a lone port, convert to int.
```
robbrockbank: Shuold

calico_containers/calicoctl.py
```diff
@@ -1105,6 +1187,39 @@ def print_paragraph(msg):
...
 
 
+def parse_ports(ports_str):
+    """
+    Parse a a string representing a port list into a list of ports and
+    port ranges.
+
+    Returns None if the input is None.
+
+    :param StringTypes|NoneType ports_str: string representing a port list.
```
robbrockbank: Worth validating the port values are in the correct range?

calico_containers/calicoctl.py
```diff
@@ -677,6 +690,77 @@ def profile_rule_update(name):
...
+    else:
+        rules = profile.rules.outbound_rules
+
+    if operation == "add":
+        if position is None:
+            # Default to append.
+            position = len(rules)
+        if not 0 <= position <= len(rules):
+            print "Position %s is out-of-range." % position
+        if rule in rules:
```
robbrockbank: This would prevent you moving a rule to an earlier slot without deleting the rule first.   Is that what we'd want?

calico_containers/calicoctl.py
```diff
@@ -677,6 +690,77 @@ def profile_rule_update(name):
...
+    if direction == "inbound":
+        rules = profile.rules.inbound_rules
+    else:
+        rules = profile.rules.outbound_rules
+
+    if operation == "add":
+        if position is None:
+            # Default to append.
+            position = len(rules)
+        if not 0 <= position <= len(rules):
```
robbrockbank: seems a little odd using different indexing here

calico_containers/calicoctl.py
```diff
@@ -1105,6 +1202,50 @@ def print_paragraph(msg):
...
+                sys.exit(1)
+            if not (0 <= min <= 65535):
+                print "Port minimum (%s) out-of-range." % min
+                sys.exit(1)
+            if not (0 <= max <= 65535):
+                print "Port maximum (%s) out-of-range." % max
+                sys.exit(1)
+            parsed_ports.append("%s:%s" % (min, max))
+        else:
+            # Should be a lone port, convert to int.
```
robbrockbank: not validating this port value.

calico_containers/calicoctl.py
fasaxc: I think that's good enough for now.

calico_containers/calicoctl.py
fasaxc: What do you mean?  insert supports 0 to mean "at the beginning" and `len(list)` to mean "after the end"

calico_containers/calicoctl.py
fasaxc: It gets validated by the regex above

calico_containers/calicoctl.py
robbrockbank: The --at suggests what position you are inserting it at (same as the regular python indexing).

So since our indexing starts at 1, a value of 1 would be insert at 1 and shift everything else along.

calico_containers/calicoctl.py
robbrockbank: but the max min value doesn't does it?

calico_containers/calicoctl.py
fasaxc: Ugh, yeh, how about something like `--before=<POSITION>`

calico_containers/calicoctl.py
fasaxc: Ah, gotcha.  Nope, that's missing.

calico_containers/calicoctl.py
fasaxc: Sorry, being dopey, yes the indexing is wrong.

calico_containers/calicoctl.py
fasaxc: done

robbrockbank:
LGTM - merging.

PR #212, opened 2015-06-17
Improve GCE instructions

fasaxc:
Covers enabling IPIP on the default pool and makes it easier to get calicoctl onto a host.


docs/GCE.md
```diff
@@ -1,63 +1,110 @@
...
 ```
 gcloud config set project PROJECT
 ```
-
 And set a default zone
 ```
 gcloud config set compute/zone us-central1-a
 ```
+## Setting up GCE networking
+GCE blocks traffic between hosts by default, which prevents Calico traffic from flowing.  Run the following command to allow traffic to flow between containers:
```
caseydavenport: This line is a bit confusing to me - It says that GCE blocks traffic between hosts, but then says that this rule allows traffic to flow between containers (which may not be on different hosts).

Maybe just say "allow traffic to flow between containers on different hosts" instead?

tomdee:
The files in the tests/scale directory either need to be updated or deleted.

docs/GCE.md
```diff
@@ -1,63 +1,110 @@
...
 ```
+Note: we disable automated reboots for this demo to avoid interrupting the instructions.
 
-And the compute nodes with
+Then create the cluster with the following command ("calico-1 calico-2 calico-3" is the list of nodes to create):
 ```
-gcloud compute instances create core1 core2 --image https://www.googleapis.com/compute/v1/projects/coreos-cloud/global/images/coreos-alpha-618-0-0-v20150312 --machine-type g1-small --metadata-from-file user-data=cloud-config.yaml --can-ip-forward
+gcloud compute instances create \
+  calico-1 calico-2 calico-3 \
+  --image https://www.googleapis.com/compute/v1/projects/coreos-cloud/global/images/coreos-alpha-709-0-0-v20150611 \
```
lxpollitt: This should work instead if we want to simplify slightly (though it is now two options rather than one):
--image-project coreos-cloud
--image coreos-alpha-709-0-0-v20150611

docs/GCE.md
```diff
@@ -1,63 +1,110 @@
...
-## Setting up GCE networking
-In order for routing to work correctly between hosts, you must set up a firewall rule on GCE to allow IPIP packets to flow between the instances. The command for doing this is as follows.
+# Grab our private IP from the metadata service:
+export metadata_url="http://metadata.google.internal/computeMetadata/v1/"
+export private_ip=$(curl "$metadata_url/instance/network-interfaces/0/ip" -H "Metadata-Flavor: Google")
 
+# Start the calico node service:
+sudo ./calicoctl node --ip=$private_ip
+
+# Work-around a [BIRD routing issue](http://marc.info/?l=bird-users&m=139809577125938&w=2):
```
lxpollitt: I would be temped to say:
# Add a route to point BIRD in the right direction (as suggested [here])

docs/GCE.md
```diff
@@ -1,72 +1,115 @@
...
+```
+IP-in-IP alows Calico to route traffic between containers.  NAT allows the containers to make outgoing connections to the internet.
 
+## Create a couple of containers and check connectivity
+On one host, run:
+```
+export DOCKER_HOST=localhost:2377
+docker run -e CALICO_IP=192.168.1.1 --name container-1 -tid busybox
+./calicoctl profile add TEST
+./calicoctl profile TEST member add container-1
```
lxpollitt: On a 2nd read of this, I wonder if it would be punchier to use "-e CALICO_PROFILE=test" on each node.

docs/GCE.md
```diff
@@ -1,72 +1,112 @@
 # Running calico-docker on GCE
-Calico runs on the Google Compute Engine (GCE), but there are a few tweaks required to the main Getting Started instructions.  The following instructions show the full power of the Calico routing and security model on GCE (and allow GCE to be used for testing)
+Calico runs on the Google Compute Engine (GCE), but there are a few tweaks required to the main Getting Started instructions.  The following instructions show the full power of the Calico routing and security model on GCE (and allow GCE to be used for testing).
 
 ## Getting started
-These instructions assume a total of three GCE hosts running CoreOS. One to run an etcd "cluster", and then two compute nodes.
-Documentation on running CoreOS on GCE is [here](https://coreos.com/docs/running-coreos/cloud-providers/google-compute-engine/)
+These instructions assume a total of three GCE hosts running CoreOS. Documentation on running CoreOS on GCE is [here](https://coreos.com/docs/running-coreos/cloud-providers/google-compute-engine/)
```
alexwlchan: This link should have a more specific description than “here”; how about “on the CoreOS website”?

docs/GCE.md
```diff
@@ -1,72 +1,112 @@
...
 
 Download and install GCE and login to your account. Full documentation on how to install gcloud are given [here](https://cloud.google.com/compute/docs/gcloud-compute/).
```
alexwlchan: And while you’re removing “here” from the previous link, this line needs tidying up too.

fasaxc:
@lxpollitt Are you happy for this to be merged?

lxpollitt:
@fasaxc Yes, I think so.  We may decide to tweak further later, but this is worth getting merged in now.

PR #213, opened 2015-06-17
Fix up the STs

tomdee:
- Stop the calico/build image from depending on calico/node
- Rework the Makefile
- Get tests working
- Add a timer to tests and annotate slow tests
- Sort out deleting containers/images


Makefile
```diff
@@ -1,20 +1,33 @@
...
+SRCDIR=calico_containers
+PYCALICO=$(wildcard $(SRCDIR)/pycalico/*.py) $(wildcard $(SRCDIR)/*.py)
+BUILD_DIR=build_calicoctl
+BUILD_FILES=$(BUILD_DIR)/Dockerfile $(BUILD_DIR)/requirements.txt
+# There are subdirectories so use shell rather than wildcard
+NODE_FILESYSTEM=$(shell find node_filesystem/ -type f)
+NODE_FILES=Dockerfile $(wildcard image/*) $(NODE_FILESYSTEM)
+GIT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null)
+IP_ADDRESS=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
+$(info Local IP address is ${IP_ADDRESS})
```
spikecurtis: Why are we echoing this out on every `make` command?

Makefile
```diff
@@ -50,16 +64,32 @@ ut-circle:
...
+
+fast-st: binary calico_containers/busybox.tar calico_containers/calico-node.tar run-etcd
+	nosetests calico_containers/tests/st/ -sv --nologcapture --with-timer -a '!slow'
+
+run-etcd:
+	-docker rm -f calico-etcd
+	docker run --detach \
+	--publish 2379:2379 \
+	--name calico-etcd quay.io/coreos/etcd:v2.0.11 \
+	--advertise-client-urls http://$(IP_ADDRESS):2379 \
```
spikecurtis: Need a way to override the `IP_ADDRESS` to one of your choosing.  Docker chooses 10.0/16 as a bridge subnet and on Vagrant machines the default NAT interface is 10.0.2.15.  See `get_ip()` in `utils.py`

calico_containers/tests/st/test_arg_parsing.py
```diff
@@ -2,15 +2,15 @@
 
 from test_base import TestBase
 from docker_host import DockerHost
-
+from nose.plugins.attrib import attr
 
 class TestArgParsing(TestBase):
+    @attr('slow')
```
spikecurtis: Very surprising that this test is slow!  All it does is run calicoctl a bunch of times...

calico_containers/tests/st/test_base.py
```diff
@@ -13,55 +14,19 @@ class TestBase(TestCase):
...
-        Clean up host containers before every test.
+        Clean up before every test.
         """
-        # containers = docker.ps("-qa").split()
-        # for container in containers:
-        #     delete_container(container)
-
         self.ip = get_ip()
-        self.start_etcd()
+        # Assert that /calico doesn't exist
```
spikecurtis: Comment is incorrect.  "Delete /calico if it exists.  This ensures each test has an empty data store at start of day."

calico_containers/tests/st/docker_host.py
```diff
@@ -109,18 +114,24 @@ def remove_containers(self):
...
-        cmd = "docker rm -f $(docker ps -qa) ; docker rmi $(docker images -qa)"
-        ok_codes = [0,
-                    1,  # docker: "rm" requires a minimum of 1 argument.
-                    127,  # '"docker": no command found'
-                    255,  # '"bash": executable file not found in $PATH'
-                    ]
+        cmd = "docker rm -f $(docker ps -qa)"
+        try:
+            self.execute(cmd)
+        except CalledProcessError:
```
spikecurtis: Need a comment explaining why such a broad exception handler is justified.

calico_containers/tests/st/workload.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, host, name, ip=None, image="busybox", network=None):
         """
         self.host = host
         self.name = name
+        try:
+            host.execute('docker rm -f %s' % name)
```
spikecurtis: I don't think workloads should try to destroy anything existing with their name.  This is going to lead to hard-to-diagnose test script bugs.

Two ways to fix this instead:

1. Fix `DockerHost` to track and delete workloads when it gets shut down.
2. Stop manually naming workloads altogether.  There just isn't a need now that they are Python Objects.  Let Docker name the workload and `docker inspect` for the name at creation.  Then if the script really needs the name, it can use the `.name` field.

It would be good to do both, actually.

calico_containers/tests/st/workload.py
```diff
@@ -44,9 +48,8 @@ def __init__(self, host, name, ip=None, image="busybox", network=None):
         ]
         assert ip is None, "Static IP assignment not supported by libnetwork."
         if network:
-            assert isinstance(network, DockerNetwork)
-            # Using @squaremo's Docker UI patch, --net accepts
-            # <driver name>:<network name> as an option.
+            if network is not "none":
+                assert isinstance(network, DockerNetwork)
```
spikecurtis: A cleaner way to accomplish this would be to create module constants `NONE`, `BRIDGE` etc., for each of the built-in reserved networks that are instances of DockerNetwork.  Then we can avoid passing strings around as constants which makes maintenance a pain.

PR #214, opened 2015-06-17
Add BGP peer configuration commands

robbrockbank:


calico_containers/calicoctl.py
```diff
@@ -6,8 +6,11 @@
 ETCD_AUTHORITY [default: 127.0.0.1:4001]
 
 Usage:
-  calicoctl node --ip=<IP> [--node-image=<DOCKER_IMAGE_NAME>] [--ip6=<IP6>]
+  calicoctl node --ip=<IP> [--node-image=<DOCKER_IMAGE_NAME>] [--ip6=<IP6>] [--as=<AS_NUM>]
```
tomdee: Remind me why we want nodes to have different ASes to each other?

spikecurtis: L3 fabrics might want an AS per host or an AS per rack.

calico_containers/calicoctl.py
```diff
@@ -6,8 +6,11 @@
...
   calicoctl node stop [--force]
+  calicoctl node bgppeer add <PEER_IP> as <AS_NUM>
+  calicoctl node bgppeer remove <PEER_IP>
+  calicoctl node bgppeer show [--ipv4 | --ipv6]
```
spikecurtis: what about `--show-all-peers` ?

calico_containers/calicoctl.py
```diff
@@ -1217,6 +1342,12 @@ def print_paragraph(msg):
                                   arguments["--interface"])
                 if arguments["remove"]:
                     container_remove(arguments["<CONTAINER>"])
+        elif arguments["bgp-node-mesh"]:
+            enable = arguments["on"]
+            bgp_node_mesh(TRI_STATE_ON_OFF_MAP[(bool(arguments["off"]),
```
spikecurtis: Clever, but it's hard to follow what this is doing.  Why not just

```
if arguments["on"] or arguments["off"]:
    set_bgp_node_mesh(arguments["on"])
else:
    show_bgp_node_mesh()
```

Right now you're doing this mapping to `True`, `False` and `None`, and then as soon as you get into the function you check if it's `None` and do one, and if not do a different thing.

calico_containers/calicoctl.py
```diff
@@ -279,7 +292,7 @@ def module_loaded(module):
     return any(s.startswith(module) for s in open("/proc/modules").readlines())
 
 
-def node(ip, node_image, ip6=""):
+def node(ip, node_image, ip6="", as_num=None):
```
spikecurtis: Past due for this function to have a proper comment explaining its parameters.

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
...
+                data = json.loads(node.value)
+                data["ip"] = IPAddress(data["ip"])
+                peers.append(data)
+            except BaseException:
+                pass
+        return peers
 
     @handle_errors
-    def add_bgp_peer(self, version, ip):
+    def add_node_bgp_peer(self, hostname, version, ip, as_num):
```
spikecurtis: comment is out of date.

calico_containers/adapter/datastore.py
```diff
@@ -302,7 +310,7 @@ def ensure_global_config(self):
         self.etcd_client.write(CALICO_V_PATH + "/Ready", "true")
 
     @handle_errors
-    def create_host(self, hostname, bird_ip, bird6_ip):
+    def create_host(self, hostname, bird_ip, bird6_ip, as_num):
```
spikecurtis: comment is out of date.

calico_containers/adapter/datastore.py
```diff
@@ -322,6 +331,16 @@ def create_host(self, hostname, bird_ip, bird6_ip):
         except EtcdKeyNotFound:
             # Didn't exist, create it now.
             self.etcd_client.write(workload_dir, None, dir=True)
+
+        # Set or delete the node specific BGP AS number as required.
```
spikecurtis: Comment that if the AS number is not present, it will inherit the global value.

calico_containers/adapter/datastore.py
```diff
@@ -936,6 +1029,58 @@ def remove_container(self, hostname, container_id):
...
+        """
+        node_mesh = {"enabled": enable}
+        self.etcd_client.write(BGP_NODE_MESH_PATH, json.dumps(node_mesh))
+
+    @handle_errors
+    def get_bgp_node_mesh(self):
+        """
+        Determine whether the BGP node mesh is enabled or not.
+
+        :param enable: Whether the mesh is enabled or not.  A value of None
```
spikecurtis: method doesn't have any parameters.

calico_containers/adapter/datastore.py
```diff
@@ -936,6 +1029,58 @@ def remove_container(self, hostname, container_id):
...
+        leaves the configuration unchanged.
+        :return: Whether the BGP node mesh is enabled.
```
spikecurtis: Make sure you give the data type.  Is it a boolean?

calico_containers/adapter/datastore.py
```diff
@@ -936,6 +1029,58 @@ def remove_container(self, hostname, container_id):
             raise KeyError("%s is not a configured container on host %s" %
                            (container_id, hostname))
 
+    @handle_errors
+    def set_bgp_node_mesh(self, enable):
+        """
+        Set whether the BGP node mesh is enabled or not.
+
+        :param enable: Whether the mesh is enabled or not.
```
spikecurtis: If the data type is boolean, state that here.

calico_containers/adapter/datastore.py
```diff
@@ -936,6 +1029,58 @@ def remove_container(self, hostname, container_id):
...
+        :return: None.
+        """
```
spikecurtis: `assert enable in [True, False]`

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
         Get the configured BGP Peers
 
         :param version: "v4" for IPv4, "v6" for IPv6
-        :return: List of netaddr.IPAddress IP addresses.
+        :return: List of dictionaries:
```
spikecurtis: Can we avoid passing dictionaries around like this?  I'd much rather see an object (maybe based on `NamedTuple`) to represent this data.  It makes it much easier to catch bugs since you can't set arbitrary keys, can assert on types.

Also, we'd be able to reuse it for the node-specific stuff.  C.f. `Profile` or `Rule` objects in this file.

calico_containers/adapter/datastore.py
```diff
@@ -322,6 +331,16 @@ def create_host(self, hostname, bird_ip, bird6_ip):
         except EtcdKeyNotFound:
             # Didn't exist, create it now.
             self.etcd_client.write(workload_dir, None, dir=True)
+
+        # Set or delete the node specific BGP AS number as required.
+        if as_num is None:
+            try:
+                self.etcd_client.delete(host_path + "bgp_as")
+            except EtcdKeyNotFound:
```
spikecurtis: UTs don't hit this clause.  Please add one that does.

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
...
-    def _get_path_with_keys(self, path):
+        try:
+            nodes = self.etcd_client.read(bgp_peers_path).children
+        except EtcdKeyNotFound:
+            # Path doesn't exist.
+            return []
+
+        peers = []
+        for node in nodes:
+            if node.value is None:
```
spikecurtis: Needs a comment.  When would `node.value` be `None`?

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
...
+                continue
+            try:
+                data = json.loads(node.value)
+                data["ip"] = IPAddress(data["ip"])
+                peers.append(data)
+            except BaseException:
```
spikecurtis: Pokemon exception handling?  (Gotta' catch 'em all.)

This seems very broad.  When would we hit an exception here?

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
...
+        try:
+            nodes = self.etcd_client.read(bgp_peers_path).children
+        except EtcdKeyNotFound:
+            # Path doesn't exist.
+            return []
+
+        peers = []
+        for node in nodes:
+            if node.value is None:
+                continue
```
spikecurtis: Missing UT coverage for this line.

calico_containers/adapter/datastore.py
```diff
@@ -444,36 +463,112 @@ def get_bgp_peers(self, version):
...
-        else:
-            values = {}
-            for child in nodes:
-                value = child.value
-                if value:
-                    values[value] = child.key
-            return values
+            return []
+
+        peers = []
```
spikecurtis: isn't this just copy-and-paste from `get_bgp_peers` with a different path?  Can you refactor to call a helper function with all the common code?

calico_containers/adapter/datastore.py
```diff
@@ -485,16 +580,14 @@ def add_bgp_peer(self, version, ip):
...
         assert version in ("v4", "v6")
         assert isinstance(ip, IPAddress)
-        bgp_peer_path = BGP_PEER_PATH % {"version": version}
-
-        # Check if the peer exists.
-        if ip in self.get_bgp_peers(version):
-            return
-
-        self.etcd_client.write(bgp_peer_path, str(ip), append=True)
+        bgp_peer_path = HOST_BGP_PEER_PATH % {"hostname": hostname,
```
spikecurtis: This subtly changes the behavior of `add_node_bgp_peer`.  Before it did nothing if the peer existed.  Now it will overwrite the AS number of the peer.  Is that what we want?  Or should we not overwrite and throw an error?

I don't much care, but you need to update the comment to say the right thing!

calico_containers/tests/unit/datastore_test.py
```diff
@@ -9,7 +9,9 @@
                                                  Endpoint,
                                                  NoEndpointForContainer,
                                                  CALICO_V_PATH,
-                                                 DataStoreError)
+                                                 DataStoreError,
+                                                 BGP_NODE_MESH_PATH,
```
spikecurtis: I'm not very keen on importing path constants from the module under test.  The UTs should have their own constants that come directly from the spec.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+        assert_list_equal([], peers)
+
+    def test_get_node_bgp_peer_no_peers(self):
+        """
+        Test getting BGP peers from the datastore when the key is there but has
+        no children.
+        :return: None
+        """
+        self.etcd_client.read.side_effect = mock_read_no_bgppeers
+        peers = self.datastore.get_bgp_peers("v4")
```
spikecurtis: test is called `get_node_bgp_peers_no_peers` but you're calling `get_bgp_peers` instead of `get_node_bgp_peers`.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
         :return: None
         """
-        self.etcd_client.read.side_effect = mock_read_2_peers
+        self.etcd_client.delete = Mock()
         peer = IPAddress("192.168.3.1")
-        self.datastore.remove_bgp_peer("v4", peer)
+        self.datastore.remove_node_bgp_peer("TEST_HOST", "v4", peer)
         # 192.168.3.1 has a key ...v4/0 in the ordered list.
-        self.etcd_client.delete.assert_called_once_with(BGP_PEERS_PATH + "0")
+        self.etcd_client.delete.assert_called_once_with(
```
spikecurtis: Comment above is out of date.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+        self.etcd_client.delete.side_effect = EtcdKeyNotFound()
         peer = IPAddress("192.168.100.1")
-        assert_raises(KeyError, self.datastore.remove_bgp_peer, "v4", peer)
-        assert_false(self.etcd_client.delete.called)
+        assert_raises(KeyError, self.datastore.remove_node_bgp_peer,
+                      "BLAH", "v4", peer)
+
+    def test_set_bgp_node_mesh(self):
+        """
+        Set whether the BGP node mesh is enabled or not.
```
spikecurtis: This comment doesn't describe the test.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+        self.datastore.set_bgp_node_mesh(True)
+        self.datastore.set_bgp_node_mesh(False)
+        self.etcd_client.write.assert_has_calls([
+                       call(BGP_NODE_MESH_PATH, json.dumps({"enabled": True})),
+                       call(BGP_NODE_MESH_PATH, json.dumps({"enabled": False}))
+                     ])
+
+    def test_get_bgp_node_mesh(self):
+        """
+        Determine whether the BGP node mesh is enabled or not.
```
spikecurtis: This comment doesn't describe the test.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+            result = Mock(spec=EtcdResult)
+            result.value = "{\"enabled\": true}"
+            return result
+        self.etcd_client.read = mock_read
+
+        assert_true(self.datastore.get_bgp_node_mesh())
+
+    def test_get_bgp_node_mesh_no_config(self):
+        """
+        Determine whether the BGP node mesh is enabled or not.
```
spikecurtis: This comment doesn't describe the test.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+        leaves the configuration unchanged.
+        :return: Whether the BGP node mesh is enabled.
+        """
+        self.etcd_client.read.side_effect = EtcdKeyNotFound()
+
+        assert_true(self.datastore.get_bgp_node_mesh())
+
+    def test_set_default_node_as(self):
+        """
+        Return the default node BGP AS Number
```
spikecurtis: This comment doesn't describe the test.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+
+        :return: The default node BGP AS Number.
+        """
+        self.datastore.set_default_node_as(12345)
+        self.etcd_client.write.assert_called_once_with(BGP_NODE_DEF_AS_PATH,
+                                                       12345)
+
+    def test_get_default_node_as(self):
+        """
+        Return the default node BGP AS Number
```
spikecurtis: This comment doesn't describe the test.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1028,46 +1037,190 @@ def test_add_bgp_peer(self):
...
+            result = Mock(spec=EtcdResult)
+            result.value = 24245
+            return result
+        self.etcd_client.read = mock_read
+
+        assert_equal(self.datastore.get_default_node_as(), 24245)
+
+    def test_get_default_node_as_no_config(self):
+        """
+        Return the default node BGP AS Number
```
spikecurtis: This comment doesn't describe the test.

docs/etcdStructure.md
```diff
@@ -7,11 +7,21 @@ The following illustrates the directory structure calico uses in etcd.
  	+--calico/v1  # root namespace
  	   |--config
  	   |  |--InterfacePrefix # the prefix for Calico interface names
- 	   |  `--LogSeverityFile # Log severity level for writing to file e.g. "DEBUG"
+ 	   |  |--LogSeverityFile # Log severity level for writing to file e.g. "DEBUG"
+ 	   |  |--bgp_as # the default BGP AS number for the node-to-node mesh
```
spikecurtis: the AS number isn't for the mesh.  "the default BGP AS number for nodes"

node_filesystem/templates/bird.cfg.template
```diff
@@ -63,20 +75,40 @@ template bgp bgp_template {
   graceful restart;  # See comment in kernel section about graceful restart.
 }
 
-{{if ls "/config/bgp_peer_rr_v4"}}
-# Peering with configured neighbors.
-{{range gets "/config/bgp_peer_rr_v4/*"}}{{$nums := split .Value "."}}{{$id := join $nums "_"}}
-# For peer {{.Value}}
+{{/* ------------- Node-to-node mesh ------------- */}}
```
spikecurtis: Why not just allow this comment in the output?

calico_containers/adapter/datastore.py
robbrockbank: I agree - and I have this on my list of to-dos.  Any objection to me fixing this up later if I don't have time tonight -- that sounds like a bigger change that will need re-review.

If the only comments are docs strings and comments it'd be nice to get these changes in to tonights release.

node_filesystem/templates/bird.cfg.template
```diff
@@ -63,20 +75,40 @@ template bgp bgp_template {
...
+{{if lsdir "/host"}}
```
spikecurtis: Is this necessary?  Don't we have to at least have ourselves in the `/host` directory?

node_filesystem/templates/bird.cfg.template
```diff
@@ -46,11 +46,23 @@ protocol direct {
   interface -"cali*", "*"; # Exclude cali* but include everything else.
 }
 
+{{/* ------------ Template definitions ----------------- */}}
+{{define "MESH4"}}
+# Peering with all neighbours (mesh).
+{{range $host := lsdir "/host"}}{{$onode_as_key := printf "/host/%s/bgp_as" .}}{{$onode_ip_key := printf "/host/%s/bird_ip" .}}{{$onode_ip := getv $onode_ip_key}}{{$nums := split $onode_ip "."}}{{$id := join $nums "_"}}
```
spikecurtis: Do these need to be all on one line?  It's really hard to read.

spikecurtis:
Ok, done with first look at this.  I'd like to see it again once you've addressed the comments.

calico_containers/adapter/datastore.py
robbrockbank: yes it did turn out that way.  Again this is on my to-do.  I can change - but would require a rereview so won't be going in tonight.

calico_containers/calicoctl.py
robbrockbank: that is not yet implemented.... i don't think that should stop up releasing though.

calico_containers/calicoctl.py
robbrockbank: yeah i like that more... :-)

calico_containers/adapter/datastore.py
robbrockbank: done.

calico_containers/calicoctl.py
robbrockbank: true.  done.

calico_containers/tests/unit/datastore_test.py
robbrockbank: yup perfectly reasonable.  will be fixed in a jiffy.

node_filesystem/templates/bird.cfg.template
robbrockbank: No, not really - just results in more blank lines in the templates, but since we already have that a few more won't hurt.

node_filesystem/templates/bird.cfg.template
robbrockbank: yeah i'll do that it keeps it simpler

node_filesystem/templates/bird.cfg.template
robbrockbank: hmmm i think previous testing indicated that it was required, but checking calicoctl we do set up etcd before creating the node so we should be fine....

and i've just retested without it, so I guess its ok.  I'll remove.

calico_containers/adapter/datastore.py
```diff
@@ -211,6 +219,45 @@ def from_json(cls, json_str):
...
+        Convert the json string into a BGPPeer object.
+        :param json_str: The JSON string representing a BGPPeer.
+        :return: A BGPPeer object.
+        """
+        json_dict = json.loads(json_str)
+        return cls(json_dict["ip"], json_dict["as_num"])
+
+    def __eq__(self, other):
+        if not isinstance(other, BGPPeer):
+            return NotImplemented
```
spikecurtis: This should really be `TypeError()`, and don't forget the `()`

Also, not covered in UT.

spikecurtis:
LGTM!

calico_containers/adapter/datastore.py
robbrockbank: I'll fix up the UTs.

Are you sure about TypeError?  As per https://docs.python.org/2/reference/datamodel.html:

A rich comparison method may return the singleton NotImplemented if it does not implement the operation for a given pair of arguments.

robbrockbank: I've pushed a fix.  Apologies for missing that!  I've re-run code coverage and I can't see any lines from this PR that are not being hit.

PR #218, opened 2015-06-17
Update advanced profile guide with new rules commands.

fasaxc:
@robbrockbank this also removes the default deny from the instructions.

I updated the first part of the example to use the new rules commands but I left the other to show an example of import.


robbrockbank:
LGTM

PR #224, opened 2015-06-18
Instructions for running Calico on AWS

TrimBiggs:


docs/AWS.md
```diff
@@ -0,0 +1,148 @@
...
+Download and install AWS Command Line Interface: 
+```
+curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
+unzip awscli-bundle.zip
+sudo ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws
+```
+For more information, see Amazon's [Installing the AWS Command Line Interface](http://docs.aws.amazon.com/cli/latest/userguide/installing.html#install-bundle-other-os).
+
+Configure the AWS CLI with your User keys.
+```
```
fasaxc: Not clear what's going on here, is `aws configure` an interactive command (i.e. it asks you for those items)?

docs/AWS.md
```diff
@@ -0,0 +1,148 @@
...
+  AWS Access Key ID: <User Access Key>
+  AWS Secret Access Key: <User Secret Access Key>
+  Default region name: us-west-2
+  Default output format: <json, text, or table>
+```
+Your AWS user needs to have the policy AmazonEC2FullAccess or be in a group with this policy in order to run the ec2 commands.  This can be set in the user configuration page of the web console.
+For more information on configuration and keys, see Amazon's [Configuring the AWS Command Line Interface](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html).
+
+## Setting up AWS networking
+The AWS machines will require a login by default.  Create a Key Pair instead to use when SSHing into the instances.
```
fasaxc: I think coreos is always pubkey-only.  Might be better to just say "Create a keypair to use for ssh access to the instances."

docs/AWS.md
```diff
@@ -0,0 +1,148 @@
...
+For more information on configuration and keys, see Amazon's [Configuring the AWS Command Line Interface](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html).
+
+## Setting up AWS networking
+The AWS machines will require a login by default.  Create a Key Pair instead to use when SSHing into the instances.  
+```
+aws ec2 create-key-pair --key-name calicokey --output text > calicokey.pem
+chmod 400 calicokey.pem
+```
+
+A Security Group is required on the instances to control allowed traffic.  Create a Security Group and allow any machine to SSH, but restrict all other traffic that is not within the Security Group.
```
fasaxc: "Create a Security Group and allow any machine to SSH, but restrict all other traffic that is not within the Security Group." verges on a double negative.  How about "Create a security group that allows only SSH from the internet as well as all traffic between instances within the group."

fasaxc:
Generally looking good; a few minor comments.  Suggest you apply those and then get one of us to run through from scratch on a fresh system.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+aws configure
+>  AWS Access Key ID: <User Access Key>
+>  AWS Secret Access Key: <User Secret Access Key>
+>  Default region name: us-west-2
+>  Default output format: <json, text, or table>
+```
+Your AWS user needs to have the policy AmazonEC2FullAccess or be in a group with this policy in order to run the ec2 commands.  This can be set in the IAM User configuration page of the web console.
+For more information on configuration and keys, see Amazon's [Configuring the AWS Command Line Interface](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html).
+
+## Setting up AWS networking
```
lxpollitt: Add something like:  "Before you can use Calico to network your containers you first need to configure AWS to allow your hosts to talk to each other."

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+Create a Key Pair to use for ssh access to the instances.
+```
+aws ec2 create-key-pair --key-name calicokey --output text > calicokey.pem
+chmod 400 calicokey.pem
```
lxpollitt: I think "mykey" might be better than "calicokey" since there is nothing Calico specific about this.  It's just standard AWS setup.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+aws ec2 authorize-security-group-ingress \
+  --group-name CalicoSG \
+  --protocol tcp \
+  --port 22 \
+  --source-group 0.0.0.0/0
+aws ec2 authorize-security-group-ingress \
+  --group-name CalicoSG \
+  --protocol all \
+  --port all \
+  --source-group CalicoSG
```
lxpollitt: Ditto CalicoSG -> MySG

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+    advertise-client-urls: http://$private_ipv4:2379
+    initial-advertise-peer-urls: http://$private_ipv4:2380
+    listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
+    listen-peer-urls: http://$private_ipv4:2380,http://$private_ipv4:7001
+  units:
+    - name: etcd2.service
+      command: start
+
+```
+Note: we disable automated reboots for this demo to avoid interrupting the instructions.
```
lxpollitt: "...we disable automated CoreOS updates for this..."

docs/AWS.md
lxpollitt: Note, this will need updating wherever this value is used later too, of course.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+aws ec2 describe-instances --filter "Name=key-name,Values=calicokey" | grep PublicIpAddress
+```
+
+SSH into each node and run these commands to set up Calico:
+```
+# SSH into a node with the calicokey.pem and username core
+ssh -i calicokey.pem core@<instance IP>
+
+# Download calicoctl and make it executable:
+wget https://github.com/Metaswitch/calico-docker/releases/download/v0.4.5/calicoctl
```
lxpollitt: This will need updating to 0.4.6 (after it is released later today).

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+chmod +x ./calicoctl
+
+# Grab our private IP from the metadata service:
+export metadata_url="http://169.254.169.254/latest/meta-data"
+export private_ip=$(curl "$metadata_url/local-ipv4")
+
+# Start the calico node service:
+sudo ./calicoctl node --ip=$private_ip
+```
+Then, on any one of the hosts, run this command to create an IP pool with IP-in-IP and NAT enabled:
```
lxpollitt: "Then, on any one of the hosts, run create the IP pool Calico will use for your containers:"

Note in particular here, I don't think --ipip or --nat-outgoing are any more worthy of extra explanation than any of the parameters to the other commands in this tutorial.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+```
+./calicoctl pool add 192.168.0.0/16 --ipip --nat-outgoing
+```
+IP-in-IP alows Calico to route traffic between containers.  NAT allows the containers to make outgoing connections to the internet.
```
lxpollitt: Per previous comment, I think you should delete this.  There are lots of parameters that are equally important that we aren't explaining in detail.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+Then, the two containers should be able to ping each other:
+```
+docker exec container-2 ping -c 4 192.168.1.1
+```
+## Next steps
+Now, you may wish to follow the [getting started instructions for creating workloads](https://github.com/Metaswitch/calico-docker/blob/master/docs/GettingStarted.md#creating-networked-endpoints).
+
+## (Optional) Enabling traffic from the internet to containers
+Services running on containers in AWS can be exposed to the internet using Calico using port mapping iptables NAT rules and an appropriate Calico security profile.  For example, you have a container that you've assigned the CALICO_IP of 192.168.7.4 to, and you have NGINX running on port 80 inside the container. If you want to expose this on port 8000, then you should follow the instructions at https://github.com/Metaswitch/calico-docker/blob/master/docs/AdvancedNetworkPolicy.md to expose port 80 on the container and then run the following commands to add the port mapping:
+
```
lxpollitt: I think we should take out all references to port mapping here.  There is no need to do port mapping in this example.

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+```
+iptables -A PREROUTING -t nat -i ens4v1 -p tcp --dport 8000 -j DNAT  --to 172.168.7.4:80
```
lxpollitt: iptables -A PREROUTING -t nat -i ens4v1 -p tcp --dport 80 -j DNAT  --to 172.168.7.4:80

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
...
+```
+Be sure to also update the Security Group for any ports you want to expose:
+```
+aws ec2 authorize-security-group-ingress \
+  --group-name CalicoSG \
+  --protocol tcp \
+  --port 8000
```
lxpollitt: --port 80

docs/AWS.md
```diff
@@ -0,0 +1,147 @@
+# Running calico-docker on AWS
+Calico runs on the Amazon Web Services (AWS), but there are a few tweaks required to the main Getting Started instructions.  The following instructions show the full power of the Calico routing and security model on AWS (and allow AWS to be used for testing).
+
```
lxpollitt: The following instructions show how to network containers using Calico routing and the Calico security model on AWS.

PR #225, opened 2015-06-18
Disable BIRD's recursive route check for GCE compatibility.

fasaxc:
@robbrockbank This pulls the version of BIRD that I built from my PR and removes the work-around from the instructions: https://github.com/Metaswitch/calico-bird/pull/2

Blocked on @spikecurtis's review of that PR, I suppose.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,5 +73,5 @@ COPY calico_containers/adapter /calico_containers/adapter
 COPY calico_containers/__init__.py /calico_containers/
 
 # Copy patched BIRD daemon with tunnel support.
-RUN curl -L https://www.dropbox.com/s/ymbvyi6388h92qg/bird?dl=1 -o /usr/sbin/bird && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -73,5 +73,5 @@ COPY calico_containers/adapter /calico_containers/adapter
 COPY calico_containers/__init__.py /calico_containers/
 
 # Copy patched BIRD daemon with tunnel support.
-RUN curl -L https://www.dropbox.com/s/ymbvyi6388h92qg/bird?dl=1 -o /usr/sbin/bird && \
+RUN curl -L https://www.dropbox.com/s/xjhfckzse25x554/bird-6af4e30d3fccb0c6bd184e9168294c807e1e6d68?dl=0 -o /usr/sbin/bird && \
```
robbrockbank: I think this should be dl=1

fasaxc: Works either way, I think but no harm in being consistent.  It sets the content disposition header, which only affects browsers.

robbrockbank:
LGTM

robbrockbank:
Ran local ST to check multi host scenario.  Looks good, so merging.

PR #229, opened 2015-06-18
Add show all nodes command

robbrockbank:


calico_containers/adapter/datastore.py
```diff
@@ -590,6 +596,87 @@ def remove_ip_pool(self, version, pool):
...
+        Return a list of hostnames
+        :return:  A list of hostnames configured in etcd.
+        """
+        # Get a list of the hosts.
+        hosts = []
+        results = self.etcd_client.read(HOSTS_PATH)
+        for result in results.children:
+            packed = result.key.split("/")
+            # Handle no hosts being configured (etcd returns the parent).
+            if len(packed) != 5:
```
spikecurtis: Key parsing would be more robust if we compute the number of key elements based on HOSTS_PATH.  
```
key_len = len(HOSTS_PATH.split("/")) + 1
...
    if len(packed) != key_len:
    ...
```

robbrockbank: That feels more fragile to me since theoretically you could have HOSTS_PATH with and without the trailing / and get the same results from etcd ... but obviously HOSTS_PATH.split("/") would be different.

I'd quite like to sort out all of the key parsing to use actual regexes with names groups.

calico_containers/adapter/datastore.py
```diff
@@ -590,6 +596,87 @@ def remove_ip_pool(self, version, pool):
...
+                continue
+            hosts.append(packed[-1])
+        return hosts
+
+    @handle_errors
+    def get_node_as_peer(self, version, hostname):
```
spikecurtis: I found this name very confusing, since "as" = Autonomous System.  Can we rename this?

Also, I don't think it's a very appropriate method to become part of the public interface to datastore.py.  You only really want it as a sub-procedure of generating the "mesh" peers.  So we should give it a leading `_`.

I know you use it in calicoctl.py, but only as a test of whether or not the node's own BGP config exists.  However, that's actually an error condition an we should just throw an exception in get_node_to_node_mesh_peers.

calico_containers/adapter/datastore.py
```diff
@@ -590,6 +596,87 @@ def remove_ip_pool(self, version, pool):
...
+            # We have no BGP information for the requested version and so
+            # we will not peer with other nodes.
+            return []
+
+        peers = []
+        hostnames = self.get_hostnames()
+        for ohostname in hostnames:
+            if ohostname == hostname:
+                continue
+            peer = self.get_node_as_peer(version, ohostname)
```
spikecurtis: Implementing it this way doubles the number of etcd requests we make, since the `get_hostnames()` already walks the tree once, throwing away all the returned AS & IP data.

Do we have other reasons to want to `get_hostnames()`?

calico_containers/tests/unit/datastore_test.py
```diff
@@ -1211,6 +1212,136 @@ def test_get_default_node_as_no_config(self):
 
         assert_equal(self.datastore.get_default_node_as(), 64511)
 
+    def test_get_hostnames(self):
+        """
+        Test get_hostnames()
+        """
+        def mock_read(path):
```
spikecurtis: This isn't a very realistic read, since the host path in a live system will include all kinds of shit like containers & endpoints.  I'd like to test that the code deals with that.

calico_containers/adapter/datastore.py
robbrockbank: If you don't know the hostnames, is it possible to get the per host bgp config in a single shot without doing a recursive query?  What would the query look like?

calico_containers/tests/unit/datastore_test.py
robbrockbank: Oh lordy - so you mean that a single etcd read at the top level will return the full tree even if recursive is False?

calico_containers/adapter/datastore.py
robbrockbank: Yup that seems reasonable.

How about  _node_to_bgppeer or _bgppeer_from_node etc.

PR #232, opened 2015-06-19
Updated GettingStarted.md to use new endpoint commands

djosborne:


djosborne:
Decided against bash substitution in favor of copy-pasting endpoint id

PR #236, opened 2015-06-19
Fix ICMP handling in rules/parsing

fasaxc:
We were using strings instead of ints when generating rules and the Rule object was missing support for icmp_code.


fasaxc:
@tomdee can you review this?

tomdee:
LGTM - merged.

PR #243, opened 2015-06-24
Rework STs using obosolete profile member add commands.

robbrockbank:


calico_containers/tests/st/test_add_container.py
```diff
@@ -29,12 +29,13 @@ def test_add_container(self):
...
+            host.calicoctl("container add %s 192.168.1.1" % node)
 
-            host.calicoctl("container add %s 192.168.1.1" % node.name)
-
-            # Add the container to a profile so felix will pick it up.
+            # Create the profile, get the endpoint IDs for the containers and
+            # add the profile to the endpoint so felix will pick it up.
             host.calicoctl("profile add TEST_GROUP")
-            host.calicoctl("profile TEST_GROUP member add %s" % node.name)
+            ep = host.calicoctl("container %s endpoint-id show" % node).strip()
```
tomdee: Are ou sure you need this strip? Would it be better to put the strip in the calicoctl method?

tomdee:
One small comment. The tests are passing on semaphore now. Woohoo!

calico_containers/tests/st/test_add_container.py
robbrockbank: Yes that might be nicer.  We'd never really want the trailing /n from the command output.  I'll update.

robbrockbank:
Tom, strip() is now pulled into the execute() method.  Also updated a couple of doc strings that were missing some details.

tomdee:
LGTM

PR #248, opened 2015-06-25
Add show all peers functionality

robbrockbank:


robbrockbank:
No longer require this functionality as it is covered by the calicoctl status command which actually gives more detailed information.

PR #251, opened 2015-06-26
Djo update endpoint

djosborne:
@robbrockbank 
Turns out we were incorrectly setting the `endpoint.workload_id`, but since it wasn't being used by `set_endpoint` we didn't notice it.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -1020,19 +1020,15 @@ class DatastoreClient(object):
             return eps.pop()
 
     @handle_errors
-    def set_endpoint(self, hostname, orchestrator_id, container_id, endpoint):
+    def set_endpoint(self, endpoint):
         """
         Write a single endpoint object to the datastore.
 
-        :param hostname: The hostname for the Docker hosting this container.
-        :param orchestrator_id: The orchestrator the the workload (or container)
-        belongs to.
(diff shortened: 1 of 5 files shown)


calico_containers/pycalico/netns.py
```diff
@@ -110,7 +110,7 @@ def remove_ip_from_interface(container_pid, ip, interface_name,
                       shell=True)
 
 
-def set_up_endpoint(ip, hostname, orchestrator_id, cpid, next_hop_ips,
+def set_up_endpoint(ip, hostname, orchestrator_id, workload_id, cpid, next_hop_ips,
                     veth_name=VETH_NAME,
                     proc_alias=PROC_ALIAS,
```
robbrockbank: Looks like doc string needs updating.

calico_containers/pycalico/datastore.py
```diff
@@ -1020,7 +1020,7 @@ def get_endpoint(self, hostname=None, orchestrator_id=None,
             return eps.pop()
 
     @handle_errors
-    def set_endpoint(self, hostname, orchestrator_id, container_id, endpoint):
+    def set_endpoint(self, endpoint):
         """
```
robbrockbank: Docstring needs updating.

robbrockbank:
Fix up the parms in the doc strings and then LGTM.

Issue #233, opened 2015-06-19
calicoctl node should check IP address and warn


spikecurtis:
A common problem setting up Calico for users is fat-fingering the IP address on `calicoctl node` command.  We can help avoid this by querying the system for configured IP addresses and verify that the specified IP is among them, dropping a big helpful warning if it is not.

tomdee:
We could also default to using the first IP address. This would make demos nicer,

PR #252, opened 2015-06-26
added auto ip detection and hostname collision detection

### LINKED ISSUE #234: calicoctl node should warn if hostname already exists
A common problem new users have is trying to set up Calico on a few nodes that don't have unique hostnames.  This is particularly common if they use Vagrant, since they'll often just leave the default hostname.

I propose that `calicoctl node` do a little proactive checking to warn the user if they're likely to run up against this issue:

If the hostname already exists in the etcd database and it has a different IP address than the one the user is configuring, drop a warning:  "Hostname %s is already in use with IP address %s.  Calico requires each compute host to have a unique hostname.  You may ignore this warning if you are reconfiguring the host with name %s"

We should disable this check if `calico-node` is already running on the host, since it's likely that the user is just trying to reconfigure the host.

djosborne:
Fixes #234 and #233

### DIFF
--- a/calico_containers/calicoctl.py
+++ b/calico_containers/calicoctl.py
@@ -20,7 +20,7 @@ Override the host:port of the ETCD server by setting the environment variable
 ETCD_AUTHORITY [default: 127.0.0.1:4001]
 
 Usage:
-  calicoctl node --ip=<IP> [--node-image=<DOCKER_IMAGE_NAME>] [--ip6=<IP6>]
+  calicoctl node [--ip=<IP>] [--ip6=<IP6>] [--node-image=<DOCKER_IMAGE_NAME>]
                            [--as=<AS_NUM>] [--log-dir=<LOG_DIR>]
   calicoctl node stop [--force]
   calicoctl node bgppeer add <PEER_IP> as <AS_NUM>
@@ -334,7 +334,7 @@ def module_loaded(module):
     return any(s.startswith(module) for s in open("/proc/modules").readlines())
 
 
-def node(ip, node_image, log_dir, ip6="", as_num=None):
+def node(node_image, log_dir, ip="", ip6="", as_num=None):
     """
     Create the calico-node container and establish Calico networking on this
     host.
@@ -352,6 +352,39 @@ def node(ip, node_image, log_dir, ip6="", as_num=None):
 
     # Print warnings for any known system issues before continuing
     checksystem(fix=False, quit_if_error=False)
+
+    # Get IP address of host, if none was specified
+    if not ip:
+        ip = get_first_ip()
+        if ip:
+            print "No IP provided. Using detected IP: %s" % ip
+        else:
+            print "Couldn't autodetect a management IP address. Please rerun" \
+                  "command with the --ip=<IP_ADDRESS> flag."
+            sys.exit(1)
+
+    # Warn if this hostname is already in use.
+    # If there's already a calico-node container on this host, they're probably
+    # just re-running node to do an update, so only issue a warning
+    if len(docker_client.containers(filters={'name': 'calico-node'})) == 0:
+        # Otherwise, check if another host with the same hostname
+        # is already configured
+        try:
+            current_ipv4, current_ipv6 = client.get_host_ips(hostname)
+        except KeyError:
+            # No other machine has registered configuration under this hostname.
+            # This must be a new host with a unique hostname, which is the
+            # expected behavior.
+            pass
+        else:
+            if current_ipv4 != "" and current_ipv4 != ip:
+                # User is either reconfiguring this node to use another IP,
+                # or another host may share the same hostname.
+                print "WARNING: Hostname %s is already in use with IP address " \
+                      "%s. Calico requires each compute host to have a " \
+                      "unique hostname. If this is your first time running" \
+                      "'calicoctl node' on this host, check to make sure" \
+                      "another host is not already using the same hostname"  % (hostname, ip)
 
     # Set up etcd
     ipv4_pools = client.get_ip_pools("v4")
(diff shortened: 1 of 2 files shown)


calico_containers/calicoctl.py
```diff
@@ -437,6 +470,23 @@ def normalize_version(version):
...
     return [int(x) for x in re.sub(r'(\.0+)*$','', version).split(".")]
 
+
+def get_first_ip():
+    """
+    Searches for the first IP address (that isn't the docker0 veth).
+    Since already has a dependency on 'ip' binary, we'll use that to
+    find the address.
+    """
+    ip = sh.Command._create("ip")
```
spikecurtis: the usual python way to do this is:

```
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.connect(("8.8.8.8", 80))
ip = s.getsockname()[0]
s.close()
```

calico_containers/pycalico/datastore.py
```diff
@@ -491,7 +491,8 @@ def ensure_global_config(self):
     @handle_errors
     def create_host(self, hostname, bird_ip, bird6_ip, as_num):
         """
-        Create a new Calico host.
+        Create a new Calico host configuration in etcd.
+        This does not start a host container.
```
spikecurtis: I don't know what "start a host container" means in this context.

calico_containers/calicoctl.py
```diff
@@ -356,6 +356,39 @@ def node(ip, node_image, log_dir, ip6="", as_num=None):
...
+        if ip:
+            print "No IP provided. Using detected IP: %s" % ip
+        else:
+            print "Couldn't autodetect a management IP address. Please rerun" \
+                  "command with the --ip=<IP_ADDRESS> flag."
+            sys.exit(1)
+
+    # Warn if this hostname is already in use.
+    # If there's already a calico-node container on this host, they're probably
+    # just re-running node to do an update, so only issue a warning
```
spikecurtis: comment says "only issue a warning" but I think the code will not issue a warning if there is already a calico-node container.

calico_containers/calicoctl.py
```diff
@@ -356,6 +356,39 @@ def node(ip, node_image, log_dir, ip6="", as_num=None):
...
+    # Get IP address of host, if none was specified
+    if not ip:
+        ip = get_first_ip()
+        if ip:
+            print "No IP provided. Using detected IP: %s" % ip
+        else:
+            print "Couldn't autodetect a management IP address. Please rerun" \
+                  "command with the --ip=<IP_ADDRESS> flag."
+            sys.exit(1)
+
```
spikecurtis: This doesn't do what https://github.com/Metaswitch/calico-docker/issues/233 suggested, which is check that the address specified on the command line exists.

spikecurtis: Should do this for IPv6 as well.

calico_containers/calicoctl.py
```diff
@@ -356,6 +356,39 @@ def node(ip, node_image, log_dir, ip6="", as_num=None):
...
+        else:
+            if current_ipv4 != "" and current_ipv4 != ip:
+                # User is either reconfiguring this node to use another IP,
+                # or another host may share the same hostname.
+                print "WARNING: Hostname %s is already in use with IP address " \
+                      "%s. Calico requires each compute host to have a " \
+                      "unique hostname. If this is your first time running" \
+                      "'calicoctl node' on this host, check to make sure" \
+                      "another host is not already using the same hostname"  % (hostname, ip)
+
```
spikecurtis: move these procedures into their own functions.  `node()` is too cluttered as it is!

calico_containers/calicoctl.py
djosborne: Wouldn't that attempt to connect to Google's DNS server? What if the container / host doesn't have a connection to the WAN? Wouldn't this fail?

calico_containers/calicoctl.py
djosborne: You're right. Its the comment that needs fixing, not the behavior, right?

The desired behavior is that - If there is a calico-node on this container, the etcd entry for this hostname must be this host. Right?

spikecurtis: Yeah, it's the comment that's wrong.  I think the behavior is ok.

calico_containers/calicoctl.py
spikecurtis: Yeah, I think it would.  It's fine to leave this as is; it's meant to be a convenience on systems with only one interface.  Doesn't need to be foolproof.

calico_containers/calicoctl.py
```diff
@@ -437,6 +461,64 @@ def normalize_version(version):
     """
     return [int(x) for x in re.sub(r'(\.0+)*$','', version).split(".")]
 
+
+def get_host_ips():
```
djosborne: Looking for advice on whether to use regex or split the output by spaces and endlines (see get_host_ip6s). The ipv6 regex is a bit rediculous, with the double colon shortener.

spikecurtis: ip route is structured text; you shouldn't need a regex.  Just split the output and use `netaddr` to parse addresses.

calico_containers/calicoctl.py
```diff
@@ -437,6 +461,64 @@ def normalize_version(version):
...
+    """
+    Searches for the first IP address (that isn't the docker0 veth).
```
spikecurtis: comment is out of date.

calico_containers/calicoctl.py
```diff
@@ -437,6 +461,64 @@ def normalize_version(version):
...
+    """
+    # Use the IP command since calico-docker has it as a dependency
+    ip = sh.Command._create("ip")
+    route_output = ip("-6", "route").stdout.strip()
+    routes = route_output.split("\n")
+    ip_6_addrs = []
+    for route in routes:
+        if "docker0" not in route:
+            ip_6_addr = route.split()[0]
```
spikecurtis: this is wrong.  The `ip -6 route` command doesn't give the local address assigned to the interface, just the prefix.  Try `ip -o -6 addr`

calico_containers/calicoctl.py
```diff
@@ -356,6 +356,30 @@ def node(ip, node_image, log_dir, ip6="", as_num=None):
...
+        print "WARNING: Could not confirm that the provided IPv4 address is assigned" \
+              "to this host."
+
+    if ip6 and ip6 not in get_host_ip6s():
+        print "WARNING: Could not confirm that the provided IPv6 address is assigned" \
+              "to this host."
+
+    # Warn if this hostname conflicts with an existing host
+    warn_if_hostname_conflict(ip)
+
```
spikecurtis: can you move all these checks into a separate function?

spikecurtis:
Couple nits.  Looks good, though and I don't need to see it again after you fix them up.

PR #253, opened 2015-06-29
Add BGP Peer STs (and fix bugs in existing function)

robbrockbank:


calico_containers/tests/st/utils/workload.py
```diff
@@ -125,15 +128,44 @@ def assert_can_ping(self, ip, retries=0):
...
+            try:
+                ping()
+            except CommandExecError:
+                pass
+            else:
+                raise Exception("%s can ping %s" % (self, ip))
+
+        return retry_until_success(cant_ping,
+                                   retries=retries,
+                                   ex_class=Exception)
```
spikecurtis: Overly broad.  Use a specific exception.

calico_containers/tests/st/utils/workload.py
```diff
@@ -125,15 +128,44 @@ def assert_can_ping(self, ip, retries=0):
...
     def assert_cant_ping(self, ip, retries=0):
-        for retry in range(retries + 1):
-            self.assert_can_ping(ip)
-            if retry >= retries:
-                raise Exception("Workload can unexpectedly ping %s" % ip)
+        """
+        Execute a ping from this workload to the ip.  Assert that the workload
+        cannot ping an IP.  Use retries to compensate for network uncertainty
+        and convergence.
+
```
spikecurtis: This function is odd.  The goal is to verify that containers cannot reach each other.  Network uncertainty results in dropped packets, so retrying the function will just make it more likely that a dropped packet will represent a false positive.

If we're going to support retries at all on this function (and maybe we shouldn't), then you need separate numbers for

 1. Number of successful pings before giving up and failing.
 2. Number of dropped pings before declaring a pass.

The first gives you control over how long to wait for network convergence, the second allows you to protect yourself from random network events.  Note that for `assert_can_ping` these factors act in concert, here they act opposite directions.

calico_containers/tests/unit/datastore_test.py
```diff
@@ -398,17 +398,17 @@ def test_ensure_global_config(self):
         """
         self.etcd_client.read.side_effect = EtcdKeyNotFound 
         self.datastore.ensure_global_config()
-        expected_writes = [call(CALICO_V_PATH + "/Ready", "true"),
-                           call(CONFIG_PATH + "InterfacePrefix", "cali")]
-        self.etcd_client.write.assert_has_calls(expected_writes,
-                                                any_order=True)
+        expected_writes = [call(CONFIG_PATH + "InterfacePrefix", "cali"),
+                           call(CALICO_V_PATH + "/Ready", "true")]
+        self.etcd_client.write.assert_has_calls(expected_writes)
```
spikecurtis: Please add a test case that covers the bug you found, e.g. `CONFIG_PATH` exists, but `...InterfacePrefix` does not.

robbrockbank: I'm not sure how that would look as a UT.  If CONFIG_PATH exists but InterfacePrefix doesn't then the read side effect would be EtcdKeyNotFound as per this UT.

I can change this test to verify that the read is called with CONFIG_PATH+"InterfacePrefix".

One of the STs (the one where the default is set before creating the nodes) tests this with a real etcd backend.

spikecurtis: I like the idea of asserting the path that gets read.

calico_containers/tests/st/utils/workload.py
robbrockbank: Yes I can create a specific exception for this - the reason I didn't was simplicity - it wasn't correct to raise a CommandExecError and I felt that if we have a proper exception for this then we probably should for the assert_can_ping.

I can update.

spikecurtis:
There is a glaring weakness in our ST of this functionality, which is that we don't test with any BGP speakers that aren't Docker host peers.  I don't think we should block merging this PR, but we need to actually test some interesting networking networking environments, for example simulating a L3 fabric.

calico_containers/tests/st/utils/workload.py
robbrockbank: Yes I was thinking about something similar actually, but since I was moving from a ping to a no-ping scenario I wasn't convinced it was necessary ... but I guess the "no ping" could be a transient state.

I still need convincing that the assert_can_ping is immune to this problem.  For example, you have scanerio A.  Pings should work and do.  Move to scenario B ... pings should still work, so you use the assert_can_ping which succeeds.... but wait for a bit and pings stop working.  I think it is always possible to think of scenarios in both cases where stopping on the first success (be it ping or no ping) is insufficient.

IMHO, both of these should have the two sets of values.

spikecurtis:
LGTM!

spikecurtis:
semaphore is breaking, though...

robbrockbank:
Rebased and re-ran semaphore tests.

Semaphore is currently having some disk space issues.  I've contacted them via chat and am awaiting a response.

PR #258, opened 2015-06-30
Split up datatore.py into three files

tomdee:
There is no functional change from this. Just code movement


calico_containers/calicoctl.py
```diff
@@ -108,13 +108,15 @@
...
                                Rule,
-                               Rules,
-                               DataStoreError,
-                               ProfileNotInEndpoint,
+                                ProfileNotInEndpoint,
                                ProfileAlreadyInEndpoint,
                                MultipleEndpointsMatch,
                                BGPPeer,
                                IPPool)
+from calico_containers.pycalico.datastore_errors import DataStoreError, \
```
robbrockbank: Can we use a common allignment style of "from import" at least within the same file.

robbrockbank:
Worth adding soc strings to the modules to make sure we don't start putting things in the wrong place :-)

Somewhat personal pref to have datastore_data -> datastore_types or datastore_datatypes   ... but not a strong feeling.


Otherwise LGTM.

PR #262, opened 2015-07-01
Stop enforcing a profile limit of 40 chars

### LINKED ISSUE #260: Why profile name length must be less than 40 characters?
Some of our profile name is longer than 40 characters, so i wonder if this is a hard limit by something or not?

tomdee:
Fixes #260


spikecurtis:
LGTM!

PR #264, opened 2015-07-01
Only refer to workloads (not containers) in datastore code

tomdee:


calico_containers/tests/unit/datastore_test.py
```diff
@@ -1139,11 +1139,11 @@ def test_remove_container(self):
     @raises(KeyError)
     def test_remove_container_missing(self):
```
robbrockbank: test name should be updated too..

robbrockbank:
minor comment but otherwise LGTM.

PR #265, opened 2015-07-01
Kube plugin

Symmetric:
Initial checkin of Kubernetes plugin (plus a Makefile to build it into an executable binary)


Symmetric:
This PR is outdated; if we do implement proper container separation, it won't be based off this.

calico_containers/calicoctl.py
```diff
@@ -742,23 +745,9 @@ def profile_rule_add_remove(
     :return:
     """
     if icmp_type is not None:
-        try:
```
caseydavenport: Why remove this validation from calicoctl?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+# Append to existing env, to avoid losing PATH etc.
+# Need to edit the path here since calicoctl loads client on import.
+# TODO-PAT: This shouldn't be hardcoded
+os.environ['ETCD_AUTHORITY'] = 'kubernetes-master:6666'
+
+from calicoctl import container_add
+from pycalico.datastore import IF_PREFIX
+from pycalico.util import generate_cali_interface_name
+
+CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/home/vagrant/calicoctl')
```
caseydavenport: The long-term fix here is still to directly call into calicoctl using Python, correct?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+import socket
+import sys
+from subprocess import check_output, CalledProcessError, check_call
+import requests
+import sh
+
+# Append to existing env, to avoid losing PATH etc.
+# Need to edit the path here since calicoctl loads client on import.
+# TODO-PAT: This shouldn't be hardcoded
+os.environ['ETCD_AUTHORITY'] = 'kubernetes-master:6666'
```
caseydavenport: Can we make this configurable via an environment variable as well?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+
+from calicoctl import container_add
+from pycalico.datastore import IF_PREFIX
+from pycalico.util import generate_cali_interface_name
+
+CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/home/vagrant/calicoctl')
+calicoctl = sh.Command(CALICOCTL_PATH).bake(_env=os.environ)
+
+ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
```
caseydavenport: Doesn't seem like any of these are used.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+        interface_name = generate_cali_interface_name(IF_PREFIX, ep.endpoint_id)
+        node_ip = self._get_node_ip()
+        print('Adding IP %s to interface %s' % (node_ip, interface_name))
+        check_call(['ip', 'addr', 'add', node_ip + '/32',
+                    'dev', interface_name])
+        print('Finished configuring network interface')
+        return ep
+
+    def _get_node_ip(self):
+        """Get the IP for the host node from the k8s API."""
```
caseydavenport: This docstring looks wrong, since we're not using the k8s API to get the host's IP.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+        self.pod_name = None
+        self.docker_id = None
+
+    def create(self, args):
+        """"Create a pod."""
+        # Calicoctl only
+        self.pod_name = args[3].replace('-', '_')
+        self.docker_id = args[4]
+
+        print('Configuring docker container %s' % self.docker_id)
```
caseydavenport: Where are we expecting all of the print statements to end up? In the Kubelet logfiles?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+
+if __name__ == '__main__':
+    print('Args: %s' % sys.argv)
+    mode = sys.argv[1]
+
+    if mode == 'init':
+        print('No initialization work to perform')
+    elif mode == 'setup':
+        print('Executing Calico pod-creation hook')
+        NetworkPlugin().create(sys.argv)
```
caseydavenport: I think this code should be parsing the command line arguments and passing them to the correct methods, rather than just blindly passing the full args list - creates a better separation of integration code and plugin code, and makes the above methods easier to read.

calico_containers/pycalico/util.py
```diff
@@ -0,0 +1,12 @@
+def generate_cali_interface_name(prefix, ep_id):
```
caseydavenport: Is there no code somewhere already that does this that could be commonized?  How does felix usually decide which veth name to use?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+class NetworkPlugin(object):
+    def __init__(self):
+        self.pod_name = None
+        self.docker_id = None
+
+    def create(self, args):
+        """"Create a pod."""
+        # Calicoctl only
```
caseydavenport: What does "# Calicoctl only" mean here?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+    def _read_docker_ip(self):
+        """Get the IP for the pod's infra container."""
+        ip = check_output([
+            'docker', 'inspect', '-format', '{{ .NetworkSettings.IPAddress }}',
+            self.docker_id
+        ])
+        # Clean trailing whitespace (expect a '\n' at least).
+        ip = ip.strip()
+
+        print('Docker-assigned IP was %s' % ip)
```
caseydavenport: Logs probably want to include the docker ID / pod so we know which container / pod we're talking about.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+    def _configure_interface(self):
+        """Configure the Calico interface for a pod."""
+        container_ip = self._read_docker_ip()
+        self._delete_docker_interface()
+        print('Configuring Calico networking.')
+        ep = container_add(self.docker_id, container_ip, 'eth0')
+        interface_name = generate_cali_interface_name(IF_PREFIX, ep.endpoint_id)
+        node_ip = self._get_node_ip()
+        print('Adding IP %s to interface %s' % (node_ip, interface_name))
+        check_call(['ip', 'addr', 'add', node_ip + '/32',
```
caseydavenport: Could you explain what this is meant to be doing?

It appears to be adding the host node's IP address to the new calico interface for the pod.  Shouldn't it be adding the container_ip instead? Or am I missing something?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def _configure_profile(self, endpoint):
+        """
+        Configure the calico profile for a pod.
+
+        Currently assumes one pod with each name.
+        """
+        calicoctl('profile', 'add', self.pod_name)
+        pod = self._get_pod_config()
+        profile_name = self.pod_name
```
caseydavenport: Move this line to the top so that we can use profile_name everywhere that expects a profile_name, and we can replace self.pod_name in the calicoctl call above.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+
+        :param path: The relative path to an API endpoint.
+        :return: A list of JSON API objects
+        :rtype list
+        """
+        bearer_token = self._get_api_token()
+        session = requests.Session()
+        session.headers.update({'Authorization': 'Bearer ' + bearer_token})
+        response = session.get(
+            'https://kubernetes-master:6443/api/v1beta3/' + path,
```
caseydavenport: This hardcoded URL should be configurable - another environment variable?

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+        # under the 'items' key.
+        return json.loads(response_body)['items']
+
+    def _get_api_token(self):
+        """
+        Get the kubelet Bearer token for this node, used for HTTPS auth.
+        :return: The token.
+        :rtype: str
+        """
+        with open('/var/lib/kubelet/kubernetes_auth') as f:
```
caseydavenport: This needs to be resilient to the auth file not existing / being in a different location.  This auth token is not necessarily going to exist on all kubernetes deployments.  Maybe hardcoding this is fine for now, I'm not sure - but we should track this inflexibility.

From the kubelet logs: "Could not load kubernetes auth path: stat /var/lib/kubelet/kubernetes_auth: no such file or directory. Continuing with defaults."

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,309 @@
...
+            'inbound_rules': inbound,
+            'outbound_rules': outbound,
+        }
+        profile_json = json.dumps(profile, indent=2)
+        print('Final profile "%s": %s' % (profile_name, profile_json))
+        return profile_json
+
+    def _apply_rules(self, profile_name, pod):
+        """
+        Generate a new profile with the specified rules.
```
caseydavenport: A bit confusing since the caller doesn't actually specify any rules - they are generated within this method.

calico_containers/calicoctl.py
Symmetric: Accidentally pulled in a commit from Shaun, I've removed it.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yep, that's v1.1. (More specifically, we're going to pull the common logic out of calicoctl.py and into a common module, leaving calicoctl as just a CLI wrapper)

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yep, this should only set the ETCD_AUTHORITY if it's not there already.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yep, they will be logged at level 5 by kubelet (--v=5 as an argument to kubelet to allow them to be logged).

I was planning to refactor this to use the logging module, and just set stdout as the stream handler for the k8s plugin. TODO in a later PR.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Unfortunately we get a different set of args depending on the mode, so we'd have to put (some of) the parsing code inside the if branches... at which point it feels like something that should be encapsulated in the per-mode methods.

calico_containers/pycalico/util.py
Symmetric: No, I created this commonized method by pulling out the logic from the one place that it was inlined.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yeah this is hacky so it deserves a comment, good point. It's magic to ensure that there's a valid IP on all of the caliXXX interfaces, which K8s requires (iptables is painful)

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: There are a few authentication methods (c.f. https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/authentication.md) but for now I'm going to mandate token-based auth. This file is standard across the salt-provisioned kubernetes distros fwiw.

Also surprising that this is working for you without the file -- it looks like there is a way to set up an --insecure-port (defaults to localhost:8080) but this shouldn't be exposed within your cluster.

Happy for you to make this more robust though.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,327 @@
...
+
+CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/home/vagrant/calicoctl')
+print("Using CALICOCTL_PATH=%s" % CALICOCTL_PATH)
+calicoctl = sh.Command(CALICOCTL_PATH).bake(_env=os.environ)
+
+KUBE_API_ROOT = os.environ.get('KUBE_API_ROOT',
+                               'https://kubernetes-master:6443/api/v1beta3/')
+print("Using KUBE_API_ROOT=%s" % KUBE_API_ROOT)
+
+class NetworkPlugin(object):
```
caseydavenport: Nitpick: two lines before a new class.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,327 @@
...
+        # no traffic for the node_ip's subnet will use the /32 route.
+        check_call(['ip', 'addr', 'add', node_ip + '/32',
+                    'dev', interface_name])
+        print('Finished configuring network interface')
+        return ep
+
+    def _get_node_ip(self):
+        """Determine the IP for the host node.
+
+        This hits a well-known IP (the Google public DNS).
```
caseydavenport: Just a thought - does this mean we can't use the plugin on deployments that are offline? (since connecting to 8.8.8.8 will fail?)

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -0,0 +1,327 @@
...
+                this_pod = pod
+                break
+        else:
+            raise KeyError('Pod not found: ' + self.pod_name)
+        print('Got pod data %s' % this_pod)
+        return this_pod
+
+    def _get_api_path(self, path):
+        """
+        Get the list of pods from the Kube API server.
```
caseydavenport: Docstring for this guy is wrong - this doesn't just get pods, it can get any arbitrary API item.

calico_containers/integrations/kubernetes/calico_kubernetes.py
caseydavenport: Hm, interesting.  I'll have to figure out why this is working for me without the file.  For now, we can leave as it and I will make sure my demo uses this authentication file.

We can track supporting different types of auth on the backlog.

calico_containers/integrations/kubernetes/calico_kubernetes.py
caseydavenport: Didn't you mention that the rkt plugin would use a decent chunk of the code laid out in this plugin?  If they plugins share the create() and delete() methods, then it would probably be good to split out the args here.  If they don't, then this is fine how it is.

caseydavenport:
I added one or two more minor comments, but other than those I'm happy for this to go in.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yep -- I'd like to make this better, but alas the obvious way of hitting the kube API doesn't actually work in the vagrant case. We could call 'ip addr' and parse the output I suppose.

I'm going to leave as-is for now and address later, since we need to get this in tonight.

calico_containers/integrations/kubernetes/calico_kubernetes.py
Symmetric: Yep, fair cop; not that ugly in the end, so I've done it here.

Symmetric:
Closing in favour of https://github.com/Metaswitch/calico-docker/pull/295, which targets master instead of libnetwork-release

PR #271, opened 2015-07-02
Sort out the mess that is our python path

tomdee:
This checkin introduces no functional changes
It simplifies the Makefile and leaves everything sane (wheras before it was a mess)


spikecurtis:
LGTM, but breaks semaphore...

PR #272, opened 2015-07-02
Checks on proc_alias & /var/run/netns

spikecurtis:


tomdee:
LGTM (but you'll need to change the target branch)

PR #273, opened 2015-07-02
Update demo calicoctl version to be static, not latest

TrimBiggs:
The calicoctl version will soon change to work strictly with libnetwork, so the demos that use powerstrip should have a static version of calicoctl that supports powerstrip.


robbrockbank:
LGTM

PR #275, opened 2015-07-02
Build pycalico as a python wheel.

spikecurtis:


Makefile
```diff
@@ -44,6 +45,11 @@ dist/calicoctl: $(PYCALICO) calicobuild.created
 	# `user` account in the container to write to it.
 	-docker run -v `pwd`/dist:/code/dist --rm -w /code/dist calico/build \
 
+dist/pycalico-0.4.8-py2-none-any.whl: $(PYCALICO)
```
tomdee: Use a variable for the version

setup.py
```diff
@@ -0,0 +1,43 @@
...
+    keywords='calico docker etcd mesos kubernetes rkt openstack',
+
+    package_dir={"": "calico_containers"},
+    packages=["pycalico"],
+
+    # List run-time dependencies here.  These will be installed by pip when
+    # your project is installed. For an analysis of "install_requires" vs pip's
+    # requirements files see:
+    # https://packaging.python.org/en/latest/requirements.html
+    install_requires=['netaddr', 'python-etcd'],
```
tomdee: Is it intentional that this uses the upstream python-etcd rather than the calico customized version.

spikecurtis: It's not intentional.  I don't believe it's possible to control the source for the package install via this mechanism.  So, unless we've pushed our custom package to pip under a different name, this is all we can do.

FWIW, I haven't seen any issues using the upstream `python-etcd` in my Mesos plug in.

Symmetric: [This article](http://mike.zwobble.org/2013/05/adding-git-or-hg-or-svn-dependencies-in-setup-py/) claims that you can say `git+ssh://git@github.com/user/repo.git` as an item in the `dependency_links` list -- if we do need to pin to our version. IIRC that is just needed for Felix watches/streaming, so for now I think keeping it simple is fine.

Symmetric:
LGTM

setup.py
spikecurtis: Good to know for future reference!

As you say, in the name of simplicity, I'll just leave it as is.

PR #280, opened 2015-07-06
Fix bug where calicoctl artifcat isn't created in Circle

TrimBiggs:


tomdee:
LGTM. But then it would since I told you what to type :-)

There's still an issue where the calicoctl.sh files isn't being created. I will look into that.

PR #282, opened 2015-07-07
Make STs not DinD now that the daemon doesn't crash all the time

tomdee:


robbrockbank:
LGTM (once the tests pass)

PR #283, opened 2015-07-07
Build Kube-plugin binary in makefile

alexhersh:


Makefile
```diff
@@ -12,9 +12,11 @@ NODE_FILES=Dockerfile $(wildcard image/*) $(NODE_FILESYSTEM)
 LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
 ST_TO_RUN?=calico_containers/tests/st/
 
+# Scale - add var for building binaries
+
 default: all
 all: test
-binary: dist/calicoctl
+binary: dist/calicoctl /dist/calico_kubernetes
```
caseydavenport: This tries to build "/dist/calico_kubernetes" instead of "dist/calico_kubernetes" - the key bit being that the former is in the root directory.

PR #286, opened 2015-07-07
Fix stale APT cache when building calicoctl binary

spikecurtis:
Before this fix, the Dockerfile for calico/build had apt-get update and apt-get install on separate RUN lines.  This means Docker can cache the filesystem after the update.

After a few days, the apt-cache from that update gets stale, and an apt-get install tries to pull a discontinued point release of some library, and builds fail.

This fixes that by forcing them all on the same RUN line.


tomdee:
LGTM

PR #294, opened 2015-07-08
Run UT as root inside the container

spikecurtis:
The code coverage module requires the /code directory to be
writable, which it won't be if the UID of the user outside the container
doesn't match the UID of the `user` user inside the container.

Running under `user` is only required for pyinstaller, not UTs.


tomdee:
LGTM

PR #300, opened 2015-07-09
Fix up calicoctl version and update ubuntu vm names

TrimBiggs:


docs/ManualClusterSetup.md
```diff
@@ -39,21 +39,12 @@ The [example script][example-commands] assumes that your ordinary user account h
...
 
 	wget https://github.com/Metaswitch/calico-docker/releases/download/v0.5.0/calicoctl
 	chmod +x calicoctl
 
-Finally, preload the Calico Docker image.  This will make the demo more responsive the first time you run it.  Select the same version of the Calico Docker image as you selected for the calico binary.  For example, to pull the latest beta
-
-    docker pull calico/node:latest
-
-or to pull the latest released version
+Finally, preload the Calico Docker image.  This will make the demo more responsive the first time you run it.  Select the same version of the Calico Docker image as you selected for the calico binary.  For example, to pull the latest released version
```
tomdee: It should be clearer that this is an optional step.
demo is the wrong word to use here.
What is the "calico" binary?

tomdee:
One comment above then LGTM

PR #302, opened 2015-07-10
Add BGP documentation

robbrockbank:
First iteration just includes the commands.

Example topologies to follow - but I'll do as a separate PR so that we can at least get the commands documented.

### DIFF
--- a/docs/bgp.md
+++ b/docs/bgp.md
@@ -1,0 +1,201 @@
+# Configuration of BGP peering
+
+By default, when running Calico on Docker, a full BGP mesh is used to
+distribute routes between all of the Calico nodes.  This approach has scale
+limitations, and alternative topologies would be recommended for building a
+large scale network.
+
+This document describes the commands available in calicoctl for managing BGP.
+
+The calicoctl tool allows you to configure three types of BGP peer.  The first
+is the full node-to-node BGP mesh which is configured automatically and can be
+enabled and disabled globally.  The second are global peers, these are peers
+that are common to all Calico nodes.  The third are node specific peers, these
+are peers specific to a particular Calico node.  All three types of peer work
+in unison, so the automatic mesh, global and node-specific peers will all be
+established on a node when configured.
+
+
+### Configuring the default node AS number
+
+When creating a Calico node, you can optionally specify an AS number to use for
+the node.  If no AS number if specified, the node will use the global default
+value.
+
+Use the `calicoctl bgp default-node-as` command to set the global default AS
+number.  If no value is configured, the default AS number is 64511.
+
+If all of your Calico nodes are within the same AS, but you require a
+different AS number to be used (e.g because you are peering with a border
+router), changing the default AS number to the value you require eliminates
+the need to explicitly set it on a per Calico node basis.  For more
+complicated toplogies where you are explicitly setting the AS number on each
+node, the default value will not be used and therefore using this command is
+not necessary.
+
+#### Example
+
+To set the default AS number to 64512, run the following calicoctl command on
+any node:
+
+	$ calicoctl bgp default-node-as 64512
+	
+To view the current default value, run the command without specifying an AS
+number, the command will output the current value.
+
+	$ calicoctl bgp default-node-as
+	64512
+	
+
+### Disabling the full node-to-node BGP mesh
+
+If you are explicitly configuring the BGP topology for your Calico network, 
+you may wish to disable the full node-to-node mesh.  Use the
+`calicoctl bgp node-mesh` command to disable or re-enable the mesh.
+
+If you are building your network from scratch and do not need the full
+node-to-node mesh we recommend turning off the mesh before configuring your
+nodes.  If you are updating your network from a full-mesh topology to a
+different topology (e.g. to start using a cluster of route reflectors to
+increase scaling), configure the appropriate peers before disabling the mesh
+to ensure continuity of service.
+
+#### Example
+
+To turn off the full BGP node-to-node mesh run the following command on any
+node:
+
+	$ calicoctl bgp node-mesh off
+
+If you need to turn the full BGP node-to-node mesh back on run the following 
+command on any node:
+
+	$ calicoctl bgp node-mesh on
+
+To view whether the BGP node-to-node mesh is on or off, enter the command
+without specifying the parameter, the command will output the current state.
+
+	$ calicoctl bgp node-mesh
+	on
+
+
+### Configuring a global BGP peer
+
+If your network topology includes BGP speakers that will be peered with *every* 
+Calico node in your deployment, you can use the `calicoctl bgp peer` commands
+to set up the peering on your Calico nodes.  We refer to these as global peers 
+because they peer with every Calico node.
+
+Two situations where global BGP peer configuration is useful are when adding
+a border router that is peering into a full node-to-node mesh, or configuring
+a Calico network that uses one or two route reflectors to provide moderate
+scale-out capabilities.  In the latter case, each Calico node would peer to
+each of the Route Reflectors and the full node-to-node mesh would be
+disabled.
+
+#### Example
+
+To add a global BGP peer at IP address 192.20.30.40 with AS number 64567 run
+the following command on any node:
+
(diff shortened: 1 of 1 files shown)


docs/bgp.md
```diff
@@ -0,0 +1,154 @@
...
+	> calicoctl bgp default-node-as 64512
+	
+To view the current default value, enter the command without specifying a number, 
+the command will output the current value.
+
+	> calicoctl bgp default-node-as
+	64512
+	
+
+### Disabling the ful node-to-node BGP mesh
```
alexwlchan: “ful” ~> “full”

tomdee:
@robbrockbank Is this ready for review?

robbrockbank:
@tomdee Yes - sorry decided to update the status command output just before I was about to assign it.  That's done now.  Assigning to you for review :-)

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
+# Configuration of BGP peering
+
+By default, a Calico-Docker network uses a full BGP mesh to distribute routes
```
tomdee: I don't really like the term Calico-Docker network. Could you should say By default, when running Calico on Docker, a full BGP mesh is used to distribute...?

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
+# Configuration of BGP peering
+
+By default, a Calico-Docker network uses a full BGP mesh to distribute routes 
+between all of the Calico nodes.  This approach has scale limitations and so 
+various different topologies are recommended for scaling out a network.
+
+This document describes the commands available in calicoctl for managing BGP 
+in a Calico-Docker networks.
```
tomdee: I think you can just drop 'in a Calico-Docker networks.'

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
+# Configuration of BGP peering
+
+By default, a Calico-Docker network uses a full BGP mesh to distribute routes 
+between all of the Calico nodes.  This approach has scale limitations and so 
+various different topologies are recommended for scaling out a network.
+
+This document describes the commands available in calicoctl for managing BGP
```
tomdee: Every time `calicoctl` or a `calicoctl command` is mentioned, you should stick backticks around it.

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+between all of the Calico nodes.  This approach has scale limitations and so 
+various different topologies are recommended for scaling out a network.
+
+This document describes the commands available in calicoctl for managing BGP 
+in a Calico-Docker networks.
+
+### Configuring the default node AS Number
+
+When creating a Calico node, you can optionally specify an AS number to use for
+the node.  If no AS Number if specified, the node will use the global default
```
tomdee: I don't think Number should have a capital at the start.

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+value.
+
+Use the "calicoctl bgp default-node-as" command to set the global default AS 
+Number.  If no value is configured, the default AS Number is 64511.
```
tomdee: I don't think Number should have a capital at the start.

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+
+The command does not need to be run on any specific node.
+
+For example, to set the default AS Number to 64512, enter the following 
+calicoctl command:
+
+	> calicoctl bgp default-node-as 64512
```
tomdee: Would a $ be better than a >

What do we use elsewhere?

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+number, the command will output the current value.
+
+	> calicoctl bgp default-node-as
+	64512
+	
+
+### Disabling the full node-to-node BGP mesh
+
+If you are explicitly configuring the BGP topology for your Calico network, 
+first disable the full node-to-node mesh using the "calicoctl bgp node-mesh"
```
tomdee: Should they really do this first? Is there no live migration path - e.g. can they set up RRs first then turn off the mesh?

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+	64512
+	
+
+### Disabling the full node-to-node BGP mesh
+
+If you are explicitly configuring the BGP topology for your Calico network, 
+first disable the full node-to-node mesh using the "calicoctl bgp node-mesh" 
+command.
+
+The command does not need to be run on any specific node.
```
tomdee: Rather than a whole new paragraph for this, why not just tack "on any node" to the para above?

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+	| Node specific IPv6 BGP Peer | AS Num |
+	+-----------------------------+--------+
+	| aa:bb::ff                   | 64514  |
+	+-----------------------------+--------+
+
+
+### Checking the status of the BGP peers
+
+To display the status of all BGP peerings for a specific node, use the 
+"calicoctl status" command.  This displays the status of all BIRD protocols
```
tomdee: To some readers of this doc, the introduction of the term `BIRD` here is going to be confusing. Could we avoid referencing bird at all? Maybe we should remove it from the calicoctl status output too?

docs/bgp.md
```diff
@@ -0,0 +1,161 @@
...
+
+	Node1> calicoctl status
+	calico-node container is running. Status: Up 43 seconds
+
+	IPv4 Bird (BGP) status
+	BIRD 1.5.0 ready.
+	+---------+----------+---------------+-------+----------+---------+
+	|   Name  | Protocol |   Peer type   | State |  Since   |   Info  |
+	+---------+----------+---------------+-------+----------+---------+
+	| kernel1 |  Kernel  |               |   up  | 21:59:07 |         |
```
tomdee: I wonder if there's any value in including the kernel, device and Direct protocol lines in this output. WDYT?

tomdee:
I think two overall changes would be useful for this doc.
1) You should add more detail for people who need a more guidance on *when* and *why* they need all these BGP options.
2) Some people already going to understand the *why* of what they're trying to do (e.g. network admins). But they'll need a bit more detail on the impacts of these commands. e.g. What BGP sessions will be dropped? How quickly will that happen? Do we actually support running any of them at any time, and/or do we have suitable checks to stop it from happening?

tomdee:
Actually, I've just reread your PR comment and maybe my last comment is moot if you're already addressing this in your next PR.

docs/bgp.md
robbrockbank: Yup will change.  For the vagrant demo we display the full hostname which terminates with a $.

docs/bgp.md
robbrockbank: hmmm good point - yes I'll reword.

docs/bgp.md
robbrockbank: seems reasonable - was toying with getting rid of it too.  i'll update the status output.

docs/bgp.md
robbrockbank: yeah I'd be happy getting rid of it too.

robbrockbank:
Regarding point 2).... Yes some of that I had intended to be in the follow up describing real use cases - but I'll make sure all of your points are included.

tomdee:
Just reviewing now. Please do squash this before merging. We don't need 9 commits to introduce a single new file! Give me a call if you need help doing this.

docs/bgp.md
```diff
@@ -0,0 +1,193 @@
...
+If you are explicitly configuring the BGP topology for your Calico network, 
+you may wish to disable the full node-to-node mesh.  Use the
+`calicoctl bgp node-mesh` command to disable or re-enable the mesh.
+
+If you are building your network from scratch and do not need the full
+node-to-node mesh we recommend turning off the mesh before configuring your
+nodes.  If you are updating your network from a full-mesh topology to a
+different topology (e.g. to start using a cluster of route reflectors to
+increase scaling), configure the appropriate peers before disabling the mesh
+to ensure continuity of service.
```
tomdee: So the node-to-node mesh is in additional to any explicitly configured peerings? It might be worth making that more explicit in the doc?

tomdee:
:thumbs:

PR #304, opened 2015-07-10
Add --detach=false option, to run calico-node in foreground

spikecurtis:
This addition allows the Calico service to be more easily integrated with init systems, which start a process and monitor it.  In this case, calicoctl is that process, and shares fate with the calico-node container.

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -98,7 +98,7 @@ def node(arguments):
                    detach=detach)
 
 
-def node_start(node_image, log_dir, ip="", ip6="", as_num=None, detach=True):
+def node_start(node_image, log_dir, ip, ip6, as_num, detach):
     """
     Create the calico-node container and establish Calico networking on this
     host.
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 """
 Usage:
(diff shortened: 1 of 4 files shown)


calico_containers/calico_ctl/node.py
```diff
@@ -84,14 +88,17 @@ def node(arguments):
...
                    node_image=arguments['--node-image'],
                    log_dir=arguments.get("--log-dir"),
                    ip6=arguments.get("--ip6"),
-                   as_num=arguments.get("--as"))
+                   as_num=arguments.get("--as"),
+                   detach=detach)
 
 
-def node_start(node_image, log_dir, ip="", ip6="", as_num=None):
+def node_start(node_image, log_dir, ip="", ip6="", as_num=None, detach=True):
```
tomdee: I don't think you should be defaulting the "detach" argument here.

calico_containers/nose.cfg
```diff
@@ -1,4 +1,4 @@
 [nosetests]
 with-coverage=1
 cover-erase=1
-cover-package=pycalico
+cover-package=pycalico,calico_ctl
```
tomdee: @TrimBiggs FYI

tomdee:
One minor comment. Otherwise, looks great.

calico_containers/calico_ctl/node.py
spikecurtis: probably shouldn't be defaulting any args, given how this is used.  Any objection?

PR #306, opened 2015-07-10
Remove warning about sh.command logger

tomdee:
Is this actually needed?

I get messages like this when running tests
```
nosetests calico_containers/tests/st/test_multi_host.py -sv --nologcapture --with-timer
Run a mainline multi-host test. ... No handlers could be found for logger "sh.command"
```


djosborne:
I do think its needed. I find it suspicious that the message only appears on the mainline test and not the others. I think its an initialization problem - its being called too early. But I believe removing it will bring us back to the log spam we were seeing before. Have you tested this?

tomdee:
Yes, tested and not a problem when using the Makefile. If manually running tests with nosetests, users need to use --no-log-capture

PR #311, opened 2015-07-10
Profile cannot be removed if it has endpoints

TrimBiggs:
If a profile is associated with any number of endpoints, prevent the profile from being removed.

### DIFF
--- a/calico_containers/calico_ctl/profile.py
+++ b/calico_containers/calico_ctl/profile.py
@@ -48,7 +48,7 @@ Description:
 Options:
   --detailed        Show additional information.
   --no-check        Remove a profile without checking if there are endpoints
-                    associated with the profile.git
+                    associated with the profile.
   --at=<POSITION>   Specify the position in the chain where the rule should
                     be placed. Default: append at end.
 
@@ -153,7 +153,7 @@ def profile_add(profile_name):
         print "Created profile %s" % profile_name
 
 
-def profile_remove(profile_name, nocheck=None):
+def profile_remove(profile_name, nocheck):
     """
     Remove a profile as long as it does not contain any endpoints.
     Allow user to explicitly remove the profile if desired.
@@ -163,9 +163,17 @@ def profile_remove(profile_name, nocheck=None):
     """
     # Check if the profile exists.
     if client.profile_exists(profile_name):
-        members = client.get_profile_members(profile_name)
-        # Check that the nocheck flag was used or the profile has no endpoints
-        if nocheck or not members:
+        rm_profile = False
+        # Check that the nocheck flag was used
+        if nocheck:
+            rm_profile = True
+        else:
+            # Check if the the profile has endpoints associated with it
+            members = client.get_profile_members(profile_name)
+            if not members:
+                rm_profile = True
+        # Remove the profile if criteria was met
+        if rm_profile:
             client.remove_profile(profile_name)
             print "Deleted profile %s" % profile_name
         else:
--- a/calico_containers/calico_ctl/profile.py
+++ b/calico_containers/calico_ctl/profile.py
@@ -178,7 +178,7 @@ def profile_remove(profile_name, nocheck):
             print "Deleted profile %s" % profile_name
         else:
             # Members must exist if this branch is reached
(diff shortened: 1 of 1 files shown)


calico_containers/calico_ctl/profile.py
```diff
@@ -151,13 +151,20 @@ def profile_add(profile_name):
...
+    if client.profile_exists(profile_name):
+        members = client.get_profile_members(profile_name)
+        if not members:
+            client.remove_profile(profile_name)
+            print "Deleted profile %s" % profile_name
+        else:
+            print "Cannot remove profile - profile contains endpoints."
     else:
-        print "Deleted profile %s" % profile_name
+        print "Couldn't find profile with name %s" % profile_name
```
robbrockbank: Stick to common text output.  I believe we currently just display "Profile xxx not found" ... see examples below.

TrimBiggs: This was just copied from above, but I'll modify it.

calico_containers/calico_ctl/profile.py
```diff
@@ -46,6 +47,8 @@
 
 Options:
   --detailed        Show additional information.
+  --no-check        Remove a profile without checking if there are endpoints
+                    associated with the profile.git
```
robbrockbank: Typo: profile.git

calico_containers/calico_ctl/profile.py
```diff
@@ -130,7 +133,7 @@ def profile(arguments):
     elif arguments.get("add"):
         profile_add(arguments.get("<PROFILE>"))
     elif arguments.get("remove"):
-        profile_remove(arguments.get("<PROFILE>"))
+        profile_remove(arguments.get("<PROFILE>"), arguments.get("--no-check"))
```
robbrockbank: since nocheck is actually a keyword argument, it is generally better to be explicit by using a keyword argument in the invocation.

Alternatively, since we are always passing in nocheck, make it a positional arg instead and drop the keyword.

calico_containers/calico_ctl/profile.py
```diff
@@ -150,14 +153,27 @@ def profile_add(profile_name):
...
+    """
+    Remove a profile as long as it does not contain any endpoints.
+    Allow user to explicitly remove the profile if desired.
+    :param profile_name: The name of the profile to remove.
+    :param nocheck: Flag saying to remove profile regardless of endpoints.
+    :return: None.
+    """
+    # Check if the profile exists.
+    if client.profile_exists(profile_name):
+        members = client.get_profile_members(profile_name)
```
robbrockbank: IIRC, getting the profile members is a fairly intensive operation - it has to do a full dump of the endpoints and then check if the profile is in the endpoint.

I don't think we should do that if the nocheck flag is set.

TrimBiggs: how about something like this:
if nocheck:
            rm_profile = True
        else:
            members = client.get_profile_members(profile_name)
            if not members:
                rm_profile = True
        if rm_profile:
            client.remove_profile(profile_name)
            print "Deleted profile %s" % profile_name

calico_containers/calico_ctl/profile.py
```diff
@@ -150,14 +153,35 @@ def profile_add(profile_name):
...
+            members = client.get_profile_members(profile_name)
+            if not members:
+                rm_profile = True
+        # Remove the profile if criteria was met
+        if rm_profile:
+            client.remove_profile(profile_name)
+            print "Deleted profile %s" % profile_name
+        else:
+            # Members must exist if this branch is reached
+            print "Cannot remove profile - profile contains endpoints.\n" + \
```
robbrockbank: Ah, just realised - not sure about this wording.  Maybe run by Tom too...

IMHO, a profile doesn't really contain endpoints, a profile is referenced by one or more endpoints.

robbrockbank:
LGTM - One more comment about the wording ... get a second opinion and if necessary just tweak the wording accordingly.

calico_containers/calico_ctl/profile.py
TrimBiggs: I've changed this to:
"Cannot remove profile - profile in use by endpoint(s).\n"

PR #312, opened 2015-07-13
Simplify and fix show endpoint summary processing

robbrockbank:


calico_containers/calico_ctl/endpoint.py
```diff
@@ -138,44 +138,21 @@ def endpoint_show(hostname, orchestrator_id, workload_id, endpoint_id,
...
-
-            # Add up how many endpoints each workload on this host has
-            num_endpoints = 0
-            for workload, endpoints in iter(table_dict[host].items()):
-                num_endpoints += endpoints
-
+        # For each host/orchestrator, keep track of unique workload IDs and
+        # total number of endpoints.  For simplicity we just store the working
+        # data in a list [set([workload IDs...], int(running total # eps)]
+        host_orch = {}
```
tomdee: I think expanding some of these short variable names would make this code more readable

calico_containers/calico_ctl/endpoint.py
```diff
@@ -138,44 +138,21 @@ def endpoint_show(hostname, orchestrator_id, workload_id, endpoint_id,
...
+        for (hn, oi), data in host_orch.iteritems():
             # Add the results to this table
-            new_row = [endpoint.hostname,
-                       endpoint.orchestrator_id,
-                       num_workloads,
-                       num_endpoints]
-
-            x.add_row(new_row)
+            x.add_row([hn, oi, len(data[0]), data[1]])
+
```
tomdee: Does this mean you can get rid of vividict?

calico_containers/calico_ctl/endpoint.py
```diff
@@ -138,44 +138,21 @@ def endpoint_show(hostname, orchestrator_id, workload_id, endpoint_id,
...
-            num_workloads = len(table_dict[host])
-
-            # Add up how many endpoints each workload on this host has
-            num_endpoints = 0
-            for workload, endpoints in iter(table_dict[host].items()):
-                num_endpoints += endpoints
-
+        # For each host/orchestrator, keep track of unique workload IDs and
+        # total number of endpoints.  For simplicity we just store the working
+        # data in a list [set([workload IDs...], int(running total # eps)]
```
tomdee: This isn't really a list though, it's a tuple - won't it always have two values? So a named tuple would probably result in clearer code here...

calico_containers/calico_ctl/endpoint.py
robbrockbank: good spot!  - yes it means we can.

calico_containers/calico_ctl/endpoint.py
```diff
@@ -138,44 +138,21 @@ def endpoint_show(hostname, orchestrator_id, workload_id, endpoint_id,
...
+        host_orch = {}
+        for ep in endpoints:
+            key = (ep.hostname, ep.orchestrator_id)
+            data = host_orch.setdefault(key, [set(), 0])
```
tomdee: data is too generic - can you think of a better name?

tomdee:
Looks good but I think the code could be clearer. Can I take another look after you address the comments.

calico_containers/calico_ctl/endpoint.py
robbrockbank: i could use a tuple, but they are immutable, so I'd have to create a new tuple to increment the count... or wrap it in a holder class.  Didn't feel like it was worth it being only a few lines of code.

robbrockbank:
As discussed with Tom, beefed up general comment about summary processing.  Merging.

Issue #296, opened 2015-07-09
uncaught exception when running calicoctl when docker daemon is not started


djosborne:
The error is thrown by checksystem (which is used in the node command) when a docker daemon is not running on the host system:
```
$ sudo calicoctl node
Unexpected error executing command.

Traceback (most recent call last):
  File "<string>", line 177, in <module>
  File "/code/build/calicoctl/out00-PYZ.pyz/calico_ctl.node", line 91, in node
  File "/code/build/calicoctl/out00-PYZ.pyz/calico_ctl.node", line 111, in node_start
  File "/code/build/calicoctl/out00-PYZ.pyz/calico_ctl.checksystem", line 115, in check_system
  File "/code/build/calicoctl/out00-PYZ.pyz/docker.client", line 1110, in version
  File "/code/build/calicoctl/out00-PYZ.pyz/docker.client", line 106, in _get
  File "/code/build/calicoctl/out00-PYZ.pyz/requests.sessions", line 477, in get
  File "/code/build/calicoctl/out00-PYZ.pyz/requests.sessions", line 465, in request
  File "/code/build/calicoctl/out00-PYZ.pyz/requests.sessions", line 573, in send
  File "/code/build/calicoctl/out00-PYZ.pyz/requests.adapters", line 415, in send
ConnectionError: ('Connection aborted.', error(2, 'No such file or directory'))
```

Caused by an uncaught exception here:
```
    try:
        info = docker_client.version()
    except docker.errors.APIError:
```

djosborne:
We may also want to add a check to checksystem which reports the status of the docker daemon. Andd possibly run `service docker start` if the `--fix` flag is provided

tomdee:
I'm OK with adding a check to `checksystem`. I'm not OK with calling service docker start with --fix.

TrimBiggs:
I'll take a look at this.

TrimBiggs:
Closed with PR #314

PR #314, opened 2015-07-13
Catch exception with docker not running

TrimBiggs:
Resolve Issue #296


robbrockbank:
LGTM, but maybe remove text about "not found" - just leave as "Docker daemon not running."

Issue #316, opened 2015-07-14
calicoctl rule 'to' clause doesn't work following a 'from tag'.


matthewdupre:
For example:

`calicoctl profile PROFILE1 rule add inbound allow udp from 192.168.0.1 to 192.168.0.2` works, but
`calicoctl profile PROFILE1 rule add inbound allow udp from tag PROFILE1 to 192.168.0.2` doesn't.

There isn't really any workaround for this, because the 'from' has to precede the 'to' (and someone might want to pass two tags anyway.

matthewdupre:
(By works, I mean the command is accepted (no output), and by doesn't I mean rejected by printing the usage statement.)

matthewdupre:
I've tried this using the `rule update < json` command with the above "doesn't work" rule, and that works fine.  The `rule show` command also works.

TrimBiggs:
Here's the command that the below input should be matching:
`calicoctl profile PROFILE1 rule add inbound allow udp from tag PROFILE1 to 192.168.0.2`

```
calicoctl profile <PROFILE> rule add (inbound|outbound) [--at=<POSITION>]
    (allow|deny) [(
      (tcp|udp) [(from [(ports <SRCPORTS>)] [(tag <SRCTAG>)] [<SRCCIDR>])]
                [(to   [(ports <DSTPORTS>)] [(tag <DSTTAG>)] [<DSTCIDR>])] |
      icmp [(type <ICMPTYPE> [(code <ICMPCODE>)])]
           [(from [(tag <SRCTAG>)] [<SRCCIDR>])]
           [(to   [(tag <DSTTAG>)] [<DSTCIDR>])] |
      [(from [(tag <SRCTAG>)] [<SRCCIDR>])]
      [(to   [(tag <DSTTAG>)] [<DSTCIDR>])]
    )]
```

I think docopt is interpretting the "to" in the 2nd command as the <SRCCIDR>, then sees the 192.168.0.2 and cannot match it to any of its possible input criteria.

PR #321, opened 2015-07-15
Fix profile rule add without CIDRs

TrimBiggs:
Fixes Issue #316

### DIFF
--- a/calico_containers/calico_ctl/profile.py
+++ b/calico_containers/calico_ctl/profile.py
@@ -20,23 +20,23 @@ Usage:
   calicoctl profile <PROFILE> tag (add|remove) <TAG>
   calicoctl profile <PROFILE> rule add (inbound|outbound) [--at=<POSITION>]
     (allow|deny) [(
-      (tcp|udp) [(from [(ports <SRCPORTS>)] [(tag <SRCTAG>)] [<SRCCIDR>])]
-                [(to   [(ports <DSTPORTS>)] [(tag <DSTTAG>)] [<DSTCIDR>])] |
+      (tcp|udp) [(from [(ports <SRCPORTS>)] [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
+                [(to   [(ports <DSTPORTS>)] [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])] |
       icmp [(type <ICMPTYPE> [(code <ICMPCODE>)])]
-           [(from [(tag <SRCTAG>)] [<SRCCIDR>])]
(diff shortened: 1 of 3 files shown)


calico_containers/calico_ctl/profile.py
```diff
@@ -20,8 +20,8 @@
   calicoctl profile <PROFILE> tag (add|remove) <TAG>
   calicoctl profile <PROFILE> rule add (inbound|outbound) [--at=<POSITION>]
     (allow|deny) [(
-      (tcp|udp) [(from [(ports <SRCPORTS>)] [(tag <SRCTAG>)] [<SRCCIDR>])]
-                [(to   [(ports <DSTPORTS>)] [(tag <DSTTAG>)] [<DSTCIDR>])] |
+      (tcp|udp) [(from [(ports <SRCPORTS>)] [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
```
robbrockbank: Unless I am mistaken, this feels like a bug in the doc opt parser rather than in our definitions.  Out of interest did you have a look to see if doc opt could be fixed?

Anyways... assuming we go with this interface change (which you should just run by the team), we should add  the cidr to the other similar branches (e.g. icmp, <no ip type> and the profile rule remove command).  It seems extremely odd to just have it on the tcp|udp branch.  I'd also be surprised if we don't hit the same error for those other branches.  Did you try that out?

We have the ability to show the rules in pretty format - this should match the format used to enter them on the command line - so there is a missing code change there too.  I think it is done in the Rules class in the datastore_datatypes class.

TrimBiggs: I've updated each `<SRCCIRD>, <DSTCIDR>` reference to have the `cidr` keyword now, and I've added the `cidr` keyword to the pretty print function.

I've briefly looked at doc opt and can't say for sure how easy it would be to modify for this.  I imagine it wouldn't be too difficult.

There was some other discussion I had with @spikecurtis about having our own parsing function that checks for the keywords and values types of the parameters passed in.  
This would be useful to us to avoid these issues and also to be able to allow different combinations of each rule (such as allowing ports by themselves, but not allowing a tag to be programmed without a cidr, etc.)  Spike was unsure of what Felix actually should allow, which is another thing to look into.

I think this current fix is usable and works in the meantime.  It may be worth talking about a bit more to decide if this should be the fix or if it should change now/be put on the backlog to improve later.

spikecurtis:
We should have a regression test for this issue as well.

calico_containers/calico_ctl/profile.py
djosborne: +1 to our own rule parsing eventually

PR #326, opened 2015-07-15
Move BGP configuration to its own sub-tree under calico

robbrockbank:
There is a follow on part to this - to update the RR image with the new paths.  That PR will follow.

### DIFF
--- a/calico_containers/pycalico/datastore.py
+++ b/calico_containers/pycalico/datastore.py
@@ -27,7 +27,7 @@ from pycalico.datastore_errors import DataStoreError, \
 ETCD_AUTHORITY_DEFAULT = "127.0.0.1:4001"
 ETCD_AUTHORITY_ENV = "ETCD_AUTHORITY"
 
-# etcd paths for Calico
+# etcd paths for Calico workloads, endpoints and IPAM.
 CALICO_V_PATH = "/calico/v1"
 CONFIG_PATH = CALICO_V_PATH + "/config/"
 CONFIG_IF_PREF_PATH = CONFIG_PATH + "InterfacePrefix"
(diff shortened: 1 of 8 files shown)


calico_containers/pycalico/datastore.py
```diff
@@ -56,9 +65,12 @@
 cali123456789ab.
 """
 
-# The default node AS number
+# The default node AS number.  Note that this value is also hard-coded into
```
tomdee: I thought you were changing the hardcoded part?

robbrockbank: I did - but you still have to allow for the situation that ensure_global_config_exists() has not yet been called ...

calico_containers/tests/unit/datastore_test.py
```diff
@@ -52,8 +50,17 @@
                                           "1234/"
 TEST_CONT_PATH = CALICO_V_PATH + "/host/TEST_HOST/workload/docker/1234/"
 CONFIG_PATH = CALICO_V_PATH + "/config/"
-BGP_NODE_DEF_AS_PATH = CALICO_V_PATH + "/config/bgp_as"
-BGP_NODE_MESH_PATH = CALICO_V_PATH + "/config/bgp_node_mesh"
+
+BGP_V_PATH = "/calico/bgp/v1"
```
tomdee: I'm sure there's a good reason for it, but why do we have to duplicate all these constants here?

docs/etcdStructure.md
```diff
@@ -2,51 +2,59 @@
...
+ 	      |--global
+ 	      |  |--as_num    # the default BGP AS number for the nodes
+ 	      |  |--node_mesh # JSON node-to-node mesh configuration (see below)
+	      |  |--peer_v4   # Global IPv4 BGP peers (all nodes peer with)
+	      |  |  `--<BGP peer IPv4 address>  # JSON BGP peer configuration (see below)
+	      |  `--peer_v6   # Global IPv6 BGP peers (all nodes peer with)
+	      |     `--<BGP peer IPv6 address>  # JSON BGP peer configuration (see below)
+	      `--host
+	         `--<hostname>  # one for each Docker host in the cluster
+	            |--ip_addr_v4 # the IP address BIRD listens on
```
tomdee: I think felix ipip might rely on the old path here - bgp_peer_v4

calico_containers/tests/unit/datastore_test.py
robbrockbank: So we aren't simply importing the same constants in from the datastore ... just serves as an extra sanity check.

docs/etcdStructure.md
tomdee: Yep - https://github.com/Metaswitch/calico/blob/cbb92d2be88dc026ba6efcb4a238d949a02c1eea/calico/felix/fetcd.py#L61

robbrockbank: oh that's just irritating!

calico_containers/pycalico/datastore.py
tomdee: So it doesn't need to be int he bird template right? Why did you add a comment to say it does? I think I must be missing something.

tomdee:
Just the one felix issue - otherwise LGTM. Please squash before you merge.

PR #332, opened 2015-07-16
Added unit tests for calicoctl utils functions

L-MA:
Also made changes to check_ip_version function to handle exception and facilitate testing

### DIFF
--- a/calico_containers/tests/unit/calicoctl_test.py
+++ b/calico_containers/tests/unit/calicoctl_test.py
@@ -15,6 +15,8 @@
 import unittest
 from StringIO import StringIO
 from mock import patch, Mock
+from nose_parameterized import parameterized
+from netaddr import IPAddress, IPNetwork
 from calico_ctl.bgp import *
 from calico_ctl.bgp import validate_arguments as bgp_validate_arguments
 from calico_ctl.endpoint import validate_arguments as ep_validate_arguments
@@ -22,36 +24,30 @@ from calico_ctl.node import validate_arguments as node_validate_arguments
 from calico_ctl.pool import validate_arguments as pool_validate_arguments
 from calico_ctl.profile import validate_arguments as profile_validate_arguments
 from calico_ctl.container import validate_arguments as container_validate_arguments
+from calico_ctl.utils import validate_cidr, validate_ip, validate_characters
 from pycalico.datastore_datatypes import BGPPeer
 
 
 class TestBgp(unittest.TestCase):
 
-    @patch('sys.exit', autospec=True)
-    def test_validate_arguments(self, m_sys_exit):
+    @parameterized.expand([
+        ({'<PEER_IP>':'127.a.0.1'}, True),
+        ({'<PEER_IP>':'aa:bb::zz'}, True),
+        ({'<AS_NUM>':9}, False),
+        ({'<AS_NUM>':'9'}, False),
+        ({'<AS_NUM>':'nine'}, True),
+        ({'show':1, '--ipv4':1}, False)
+    ])
+    def test_validate_arguments(self, case, sys_exit_called):
         """
         Test validate_arguments for calicoctl bgp command
         """
-        # Set up arguments
-        cases = (
-            ({'<PEER_IP>':'127.a.0.1'}, False),
-            ({'<PEER_IP>':'aa:bb::zz'}, False),
-            ({'<AS_NUM>':9}, True),
-            ({'<AS_NUM>':'9'}, True),
-            ({'<AS_NUM>':'nine'}, False),
-            ({'show':1, '--ipv4':1}, True)
-        )
-
-        # Call method under test for each test case
-        # Assert that method exits on bad input
-        for case, is_valid in cases:
-            print "Testing case %s ..." % case
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
             bgp_validate_arguments(case)
-            if is_valid:
-                assert not m_sys_exit.called
-            else:
-                assert m_sys_exit.called
-            m_sys_exit.reset_mock()
+
+            # Assert that method exits on bad input
+            self.assertEqual(m_sys_exit.called, sys_exit_called)
 
     @patch('calico_ctl.bgp.BGPPeer', autospec=True)
     @patch('calico_ctl.bgp.client', autospec=True)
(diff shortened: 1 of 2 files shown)


calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            ('1.2.3.0/24', 'v4', IPAddress, False, 4),
+            ('1.2.3.0/24', 'v4', IPNetwork, True, 5),
+            ('aa:bb::ff', 'v4', IPAddress, False, 6),
+            ('aa:bb::ff', 'v6', IPAddress, True, 7)
+        )
+
+        # Call method under test for each test case
+        # Assert that method exits on bad input
+        for case in cases:
+            (ip, version, cls, is_valid, case_num) = case
```
TrimBiggs: `cls` is a keyword in python.  I should have noticed this in your `check_ip_version` function, but I think this should be renamed to something like ip_cls in here and in the function declaration for `check_ip_version` for clarity

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        """
+        # Set up arguments
+        cases = (
+            ('1.2.3.4', 'v4', IPAddress, True, 1),
+            ('1.2.3.4', 'v6', IPAddress, False, 2),
+            ('1.2.3.4', 'v4', IPNetwork, True, 3),
+            ('1.2.3.0/24', 'v4', IPAddress, False, 4),
+            ('1.2.3.0/24', 'v4', IPNetwork, True, 5),
+            ('aa:bb::ff', 'v4', IPAddress, False, 6),
+            ('aa:bb::ff', 'v6', IPAddress, True, 7)
```
TrimBiggs: These use cases look good.  I think it would be useful to test a couple more cases too:
- full IPv6 address - 1111:2222:3333:4444:5555:6666:7777:8888
- number below 4294967295 (highest possibly ipv4 address as a number)
- number above 4294967296 (parsed as an ipv6 address, since it's larger than all ipv4 addresses)

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+
+        # Call method under test for each test case
+        # Assert that method exits on bad input
+        for case in cases:
+            (ip, version, cls, is_valid, case_num) = case
+            print "Testing case %s" % case_num
+            check_ip_version(ip, version, cls)
+
+            if is_valid:
+                assert not m_sys_exit.called
```
TrimBiggs: Having asserts in here being directly called will throw a different error than we would want for the UTs.
You should use `self.assertTrue` or `self.assertFalse` calls for these two cases

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        """
+        Test validate_cidr function in calico_ctl utils
+        """
+        # Set up arguments
+        cases = (
+            ('127.a.0.1', False, 1),
+            ('aa:bb::zz', False, 2),
+            ('1.2.3.4', True, 3),
+            ('1.2.3.0/24', True, 4),
+            ('aa:bb::ff', True, 5)
```
TrimBiggs: I'd add the values here that I mentioned for the `check_ip_version`  test as well

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        Test validate_ip function in calico_ctl utils
+        """
+        # Set up arguments
+        cases = (
+            ('1.2.3.4', 'v4', True, 1),
+            ('1.2.3.4', 'v6', False, 2),
+            ('1.2.3.4', 'v4', True, 3),
+            ('1.2.3.0/24', 'v4', False, 4),
+            ('aa:bb::ff', 'v4', False, 6),
+            ('aa:bb::ff', 'v6', True, 7)
```
TrimBiggs: I'd add those test values here too

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        """
+        Test validate_ip function in calico_ctl utils
+        """
+        # Set up arguments
+        cases = (
+            ('1.2.3.4', 'v4', True, 1),
+            ('1.2.3.4', 'v6', False, 2),
+            ('1.2.3.4', 'v4', True, 3),
+            ('1.2.3.0/24', 'v4', False, 4),
+            ('aa:bb::ff', 'v4', False, 6),
```
TrimBiggs: 1, 2, 3, 4, 6, 7

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            ('1.2.3.4', 'v4', IPNetwork, True, 3),
+            ('1.2.3.0/24', 'v4', IPAddress, False, 4),
+            ('1.2.3.0/24', 'v4', IPNetwork, True, 5),
+            ('aa:bb::ff', 'v4', IPAddress, False, 6),
+            ('aa:bb::ff', 'v6', IPAddress, True, 7)
+        )
+
+        # Call method under test for each test case
+        # Assert that method exits on bad input
+        for case in cases:
```
TrimBiggs: You can actually merge this step with the step below.  Just replace the `case` here with the tuple definition and it will understand what you mean.  This goes for the other tests too (i'll note them below)

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            ('aa:bb::zz', False, 2),
+            ('1.2.3.4', True, 3),
+            ('1.2.3.0/24', True, 4),
+            ('aa:bb::ff', True, 5)
+        )
+
+        # Call method under test for each test case
+        # Assert that method returns expected output
+        for case in cases:
+            (cidr, is_valid, case_num) = case
```
TrimBiggs: You can merge this line with the line above.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            ('1.2.3.4', 'v4', True, 3),
+            ('1.2.3.0/24', 'v4', False, 4),
+            ('aa:bb::ff', 'v4', False, 6),
+            ('aa:bb::ff', 'v6', True, 7)
+        )
+
+        # Call method under test for each test case
+        # Assert that method returns expected output
+        for case in cases:
+            (ip, version, is_valid, case_num) = case
```
TrimBiggs: You can merge this line with the line above.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            ('^', False, 12),
+            ('&', False, 13),
+            ('*', False, 14),
+            ('()', False, 15)
+        )
+
+        # Call method under test for each test case
+        # Assert that method returns expected output
+        for case in cases:
+            (input_string, is_valid, case_num) = case
```
TrimBiggs: You can merge this line with the line above.

calico_containers/calico_ctl/utils.py
```diff
@@ -48,15 +48,20 @@ def check_ip_version(ip, version, cls):
...
     try:
         parsed = cls(ip)
     except AddrFormatError:
         print "%s is not a valid IP address." % ip
         sys.exit(1)
-    if "v%d" % parsed.version != version:
+    except ValueError:
+        print "IPAddress does not support netmasks or subnet prefixes."
+        sys.exit(1)
+    if parsed and "v%d" % parsed.version != version:
```
TrimBiggs: Nice catches here.

calico_containers/calico_ctl/utils.py
```diff
@@ -48,15 +48,20 @@ def check_ip_version(ip, version, cls):
     :return: The parsed object of type "type"
     """
     assert version in ("v4", "v6")
+    parsed = None
     try:
         parsed = cls(ip)
```
TrimBiggs: I mentioned in a comment in the tests that I should have caught this before, but I think you should change the name of the `cls` because `cls` is used a keyword in python.  Something like ip_cls or similar would be good.

calico_containers/tests/unit/calicoctl_test.py
TrimBiggs: I'd probably rename the `is_valid` var name to be something like `expected` or something similar be a bit more clear that this is the result you are looking for.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        )
+
+        # Call method under test for each test case
+        # Assert that method exits on bad input
+        for case in cases:
+            (ip, version, cls, is_valid, case_num) = case
+            print "Testing case %s" % case_num
+            check_ip_version(ip, version, cls)
+
+            if is_valid:
```
TrimBiggs: I'd probably rename the `is_valid` var name to be more clear that it represents what you expect as output from the function.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        # Call method under test for each test case
+        # Assert that method returns expected output
+        for case in cases:
+            (input_string, is_valid, case_num) = case
+            print "Testing case %s" % case_num
+
+            test_rtn = validate_characters(input_string)
+
+            # Assert
+            self.assertEqual(is_valid, test_rtn)
```
TrimBiggs: I'd probably rename the `is_valid` var name to be more clear that it represents what you expect as output from the function.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+        # Call method under test for each test case
+        # Assert that method returns expected output
+        for case in cases:
+            (ip, version, is_valid, case_num) = case
+            print "Testing case %s" % case_num
+
+            test_rtn = validate_ip(ip, version)
+
+            # Assert
+            self.assertEqual(is_valid, test_rtn)
```
TrimBiggs: I'd probably rename the `is_valid` var name to be more clear that it represents what you expect as output from the function.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -23,6 +23,7 @@
 from calico_ctl.pool import validate_arguments as pool_validate_arguments
 from calico_ctl.profile import validate_arguments as profile_validate_arguments
 from calico_ctl.container import validate_arguments as container_validate_arguments
+from calico_ctl.utils import *
```
caseydavenport: I like to avoid using * imports unless you really need to (e.g you're importing 100 things from this file).  Even then, it is preferred to do something like

```
from calico_ctl import utils
```

This prevents possible clashes between imports.

calico_containers/tests/unit/calicoctl_test.py
L-MA: Cannot not create test for number above 4294967296 for ipv6. It seems like there isn't support for integers/integer strings for ipv6

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -305,3 +306,118 @@ def test_validate_arguments(self, m_sys_exit):
...
+            else:
+                assert m_sys_exit.called
+            m_sys_exit.reset_mock()
+
+    def test_validate_cidr(self):
+        """
+        Test validate_cidr function in calico_ctl utils
+        """
+        # Set up arguments
+        cases = (
```
caseydavenport: Another way to do this is to use the @params decorator provided in [nose2](http://nose2.readthedocs.org/en/latest/params.html)

TrimBiggs: We're using nose, not nose2 :unamused:, which doesn't support *@params*

calico_containers/calico_ctl/utils.py
```diff
@@ -39,7 +39,7 @@
     print "Missing command: %s" % e.message
 
 
-def check_ip_version(ip, version, cls):
+def check_ip_version(ip, version, ip_cls):
```
TrimBiggs: Make sure you update the name of the param in the docstring too!

TrimBiggs:
LGTM with one comment.

calico_containers/tests/unit/calicoctl_test.py
robbrockbank: Yup no nose2 unfortunately.  Maybe worth us porting similar function since I had to use the same trick in one of my tests...   If you feel like adding in an @param style decorator function let me know :-)

caseydavenport: Why are we not using nose2?

TrimBiggs:
You can also remove the `test_arg_parsing.py` file at this point since it is all covered under the `test_validate_argument` functions.

calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -22,36 +24,33 @@
...
-        # Assert that method exits on bad input
-        for case, is_valid in cases:
-            print "Testing case %s ..." % case
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
             bgp_validate_arguments(case)
-            if is_valid:
-                assert not m_sys_exit.called
+
+            # Assert that method exits on bad input
```
robbrockbank: expected_result is an odd name since bgp_validate_arguments doesn't return a result.

Perhaps if it was sys_exit_called, you could simplify the following as:

self.assertEquals(m_sys_exit.called, sys_exit_called)

robbrockbank:
LGTM

PR #334, opened 2015-07-16
Fixes #257 More readable etcd DB dump in diags bundle.

### LINKED ISSUE #257: calicoctl diags should record the etcd db in a more readable format
Right now it's just a flat dump of the `<class 'etcd.EtcdResult'>`, but without any line breaks between objects.

Try a `PrettyPrint` of the object structure at minimum, since it's just `dicts` and `lists`

spikecurtis:


spikecurtis:
Note that there is a bug in `python-etcd` that causes `get_subtree()` to return leaf nodes twice, so they appear twice in the output.  It creates a minor readability issue, but I think we should address it upstream rather than roll our own tree iterator.

Symmetric:
LGTM

PR #335, opened 2015-07-16
Kube integration doc

alexhersh:


docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
+# Integrating the Calico-Kubernetes Plugin
+
+This guide will take you through all of the configuration hooks needed to integrate the Calico-Kubernetes plugin into your cloud platform.
```
Symmetric: Let's s/Calico-Kubernetes plugin/Calico network plugin/ throughout.

Symmetric: How about "This guide will describe the configuration required to use the Calico network plugin in your Kubernetes deployment" since they aren't all hooks.

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Setting Up Your Environment 
+   The Calico-Kubernetes plugin looks for three environment variables. If one is not set, it will assume a default value. For custom networks, you must configure these variables in the _kubelet_ environment, not the _shell_ environment.
```
Symmetric: Replace "For custom networks..." with "If you need to override the defaults, you must set these variables in the environment of the kubelet process."

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
+# Integrating the Calico-Kubernetes Plugin
+
+This guide will take you through all of the configuration hooks needed to integrate the Calico-Kubernetes plugin into your cloud platform.
+
+## Setting Up Your Environment 
+   The Calico-Kubernetes plugin looks for three environment variables. If one is not set, it will assume a default value. For custom networks, you must configure these variables in the _kubelet_ environment, not the _shell_ environment. 
+
+* #####ETCD_AUTHORITY
+   By default, the Calico-Kubernetes plugin will assume that the ETCD Authority is located at the `Master IP` of your cluster at `port 6666`. Setting the `ETCD_AUTHORITY` variable in your environment will direct Calico to the correct IP if your cluster is set up differently.
```
caseydavenport: "correct IP" -> "correct IP and port"

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+   The Calico network plugin looks for three environment variables. If one is not set, it will assume a default value. If you need to override the defaults, you must set these variables in the environment of the _kubelet_ process. 
+
+* #####ETCD_AUTHORITY
+   By default, the Calico network plugin will assume that the ETCD Authority is located at `<MASTER_IP>:6666`. Setting the `ETCD_AUTHORITY` variable in your environment will direct Calico to the correct IP if your cluster is set up differently.
+
+* #####CALICOCTL_PATH
+   This plugin requires access to the `calicoctl` binary. If your binary is not located at `/usr/bin/calicoctl`, you can set the `CALICOCTL_PATH` environment variable to the correct path.
+
+* #####KUBE_API_ROOT
+   The `KUBE_API_ROOT` environment variable specifies where the Kubernetes API resources are located, defaulting to the `<MASTER_IP>/api/v1/`
```
caseydavenport: Key thing to mention here - its not just the MASTER_IP, its the MASTER_IP and port 8080.

8080 is the default insecure port on the API server, which means this needs to be changed if authentication is in use.

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Configuring Nodes
+
+#### Creating a Calico Node with the Network Plugin
+
+* #####Automatic Install
+
+   With our latest distribtution of Calico, we have included a `--kubernetes` flag to the `calicoctl node` command that will automatically install the Calico-Kubernetes Plugin as you spin up a Calico Node.
```
caseydavenport: No need to say with out latest.  If anything, you can say "as of 0.5.1, calicoctl node supports the --kubernetes flag"

Otherwise, we'll just have to update this line at the next release.

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+   ```
+   sudo ETCD_AUTHORITY=<MASTER_IP> calicoctl node --ip=<NODE_IP> --kubernetes
+   ```
+   >_Note in this example, we embedded the ETCD_AUTHORITY environment config_
```
caseydavenport: ETCD_AUTHORITY being included in this command does not replace the need for it to be used in the kubelet process environment.

The ETCD_AUTHORITY specified here is only valid for the duration of the command.

caseydavenport: <ETCD_IP>:<ETCD_PORT>

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+
+* #####Manual Install
+
+   Alternatively, you can download the [latest release](https://github.com/Metaswitch/calico-docker/releases/latest) of the plugin binary directly from our Github Repo.
+   To apply a plugin to the node, you can use the `--plugin-dir=<PLUGIN_DIR>` option of the `calicoctl node` command
```
caseydavenport: The plugin-dir option actually doesn't apply for the kubernetes case - it is only for docker plugins.

The plugin will automatically be installed in /usr/libexec/blahblahblah on most distros, but will be installed in /etc/kubelet-plugins on CoreOS.  This is not configurable.

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,38 @@
...
+
+#### Configuring Kubelet Services
+   On each of your nodes, you will need to verify that your kubelet service config files (`/etc/systemd/kube-kubelet.service` by default) include a reference to the calico networking plugin. Look for/add this line to the config:
```
caseydavenport: This is assuming that the Kubelet is managed by systemd (which it usually is, but is not necessarily true).

Probably good to re-word this along the lines of:
 "Once the plugin is installed, you'll need to tell the Kubelet to use Calico for networking.  This is done using the `--network_plugin=calico` option when starting the Kubelet.

If you're using systemd to manage the Kubelet, you can add the option to the systemd unit file (default location is /etc/systemd/system/) and restart the Kubelet."

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,37 @@
...
+* #####KUBE_API_ROOT
+   The `KUBE_API_ROOT` environment variable specifies where the Kubernetes API resources are located, defaulting to the `<MASTER_IP>:8080/api/v1/`
+
+## Configuring Nodes
+
+#### Creating a Calico Node with the Network Plugin
+
+* #####Automatic Install
+
+   As of Calico v0.5.1, we have included a `--kubernetes` flag to the `calicoctl node` command that will automatically install the Calico-Kubernetes Plugin as you spin up a Calico Node.
```
caseydavenport: Calico-Kubernetes Plugin -> Calico network plugin

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,37 @@
...
+   sudo ETCD_AUTHORITY=<ETCD_IP>:<ETCD_PORT> calicoctl node --ip=<NODE_IP> --kubernetes
+   ```
+   >_Note in this example, we set the ETCD_AUTHORITY environment config for the duration of the command_
+
+* #####Manual Install
+
+   Alternatively, you can download the [latest release](https://github.com/Metaswitch/calico-docker/releases/latest) of the plugin binary directly from our Github Repo.
+
+#### Configuring Kubelet Services
+   On each of your nodes, you will need to verify that your kubelet service config files (`/etc/systemd/kube-kubelet.service` by default) include a reference to the calico networking plugin. Look for/add this line to the config:
```
caseydavenport: Not sure where my comment went, but to summarize:

I think this should be more generic, since not all users will be using systemd to start the kubelet (though most probably will).

Maybe say something like:
"Once the Calico network plugin is installed, you'll need to configure the Kubelet to use Calico for networking.  This can be done using the `--network_plugin=calcio` option when starting the Kubelet.

If using systemd ..."

caseydavenport:
A couple more comments, but otherwise LGTM

caseydavenport:
Missing a link to this doc.  Could you add a link to the kubernetes README.md to this?

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,37 @@
+# Integrating the Calico Network Plugin with Kubernetes
+
+This guide will describe the configuration required to use the Calico network plugin in your Kubernetes deployment.
+
+## Setting Up Your Environment 
+   The Calico network plugin looks for three environment variables. If one is not set, it will assume a default value. If you need to override the defaults, you must set these variables in the environment of the _kubelet_ process. 
+
+* #####ETCD_AUTHORITY
+   By default, the Calico network plugin will assume that the ETCD Authority is located at `<MASTER_IP>:6666`. Setting the `ETCD_AUTHORITY` variable in your environment will direct Calico to the correct IP if your cluster is set up differently.
```
Symmetric: "ETCD Authority" => "etcd datastore" (the phrase "Authority" isn't a well-known identifier)

docs/kubernetes/KubernetesIntegration.md
```diff
@@ -0,0 +1,37 @@
...
+
+* #####CALICOCTL_PATH
+   This plugin requires access to the `calicoctl` binary. If your binary is not located at `/usr/bin/calicoctl`, you can set the `CALICOCTL_PATH` environment variable to the correct path.
```
Symmetric: "you can set" => "set"

Symmetric:
A couple minor edit suggestions, otherwise LGTM.

PR #336, opened 2015-07-17
Use urllib to download the Kubernetes plugin

caseydavenport:


calico_containers/calico_ctl/node.py
```diff
@@ -498,12 +499,14 @@ def install_kubernetes(kubernetes_plugin_dir):
...
+        urllib.urlretrieve(KUBERNETES_BINARY_URL, kubernetes_binary_path)
         st = os.stat(kubernetes_binary_path)
         executable_permissions = st.st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH
         os.chmod(kubernetes_binary_path, executable_permissions)
-    except sh.ErrorReturnCode_8:
-        print "ERROR: Couldn't download the kubernetes binary"
+    except IOError:
+        print "ERROR: Couldn't download the Kubernetes plugin"
+        sys.exit(1)
+    except OSError:
```
Symmetric: Maybe split into two trys so that each except only covers the statement that you expect to raise it (e.g. OSError only for the chmod).

Symmetric:
One suggestion for improvement, but up to you whether you want to implement it. Otherwise LGTM

caseydavenport:
Merged via command line

PR #338, opened 2015-07-17
Made calico_ctl.node.get_host_ips fail-safe

djosborne:
Switched get_host_ips to use subprocess instead of sh since sh was returning blank responses on fedora and subprocess works cross platform. Also updated get_host_ips so it swallows exceptions, allowing 'calicoctl node' to continue with the node creation even if the IP check fails

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -41,6 +41,7 @@ import sys
 import os
 import stat
 import sh
+from subprocess import check_output, CalledProcessError
 import docker
 import netaddr
 import socket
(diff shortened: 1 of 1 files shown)


caseydavenport:
Did you talk to @alexhersh about the Python library he was playing with on Friday?  From what I recall, it appears to return you the system's IP addresses in nicely formatted python objects - might be worth trying out.

Not sure what it uses under the covers, but it could make this code obsolete.

djosborne:
@tomdee Can you take a look at this? I'd prefer to implement this minimum impact fix which unblocks fedora before considering adding a new library.

tomdee:
LGTM

PR #342, opened 2015-07-20
Changed all string references of v4/v6 to integer representation. 

L-MA:
Removed check_ip_version utils function.

### DIFF
--- a/calico_containers/calico_ctl/container.py
+++ b/calico_containers/calico_ctl/container.py
@@ -23,7 +23,6 @@ from pycalico import netns
 from utils import hostname, ORCHESTRATOR_ID
 from utils import client
 from utils import enforce_root
-from utils import check_ip_version
 from utils import get_container_ipv_from_arguments
 from utils import docker_client
 from utils import print_paragraph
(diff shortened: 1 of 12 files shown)


calico_containers/calico_ctl/node.py
```diff
@@ -328,11 +327,11 @@ def node_bgppeer_add(ip, version, as_num):
...
     :param ip: The address to add
-    :param version: v4 or v6
+    :param version: 4 or 6
     :param as_num: The peer AS Number.
     :return: None
     """
-    address = check_ip_version(ip, version, IPAddress)
+    address = IPAddress(ip)
     peer = BGPPeer(address, as_num)
     client.add_bgp_peer(version, peer, hostname=hostname)
```
TrimBiggs: These functions will actually have to be updated also.  `datestore.py` contains a lot of references to "v4"/"v6".  These should be changed too.  This might be my fault, I don't think I was very clear on this.

calico_containers/calico_ctl/container.py
```diff
@@ -230,12 +229,12 @@ def container_ip_add(container_name, ip, version, interface):
 
     :param container_name: The name of the container.
     :param ip: The IP to add
-    :param version: The IP version ("v4" or "v6")
+    :param version: The IP version (4 or 6)
```
TrimBiggs: `version` isn't used in this function anymore.

calico_containers/calico_ctl/container.py
```diff
@@ -302,12 +301,12 @@ def container_ip_remove(container_name, ip, version, interface):
...
```
TrimBiggs: `version` isn't used in this function anymore

robbrockbank:
LGTM

PR #347, opened 2015-07-21
Move libn veth to netns

robbrockbank:
Need to write UTs still.

### DIFF
--- a/calico_containers/calico_ctl/container.py
+++ b/calico_containers/calico_ctl/container.py
@@ -13,6 +13,8 @@ Options:
                            [default: eth1]
 """
 import sys
+import uuid
+
 import docker.errors
 from requests.exceptions import ConnectionError
 from urllib3.exceptions import MaxRetryError
@@ -20,6 +22,7 @@ from subprocess import CalledProcessError
 from netaddr import IPAddress, IPNetwork
 
 from pycalico import netns
+from pycalico.datastore_datatypes import Endpoint
 from utils import hostname, ORCHESTRATOR_ID
 from utils import client
 from utils import enforce_root
@@ -111,6 +114,7 @@ def container_add(container_name, ip, interface):
 
     :param container_name: The name or ID of the container.
     :param ip: An IPAddress object with the desired IP to assign.
+    :param interface: The name of the interface in the container.
     """
     # The netns manipulations must be done as root.
     enforce_root()
@@ -157,25 +161,39 @@ def container_add(container_name, ip, interface):
         print "IP address is already assigned in pool %s " % pool
         sys.exit(1)
 
-    # Actually configure the netns. Defaults to eth1 since eth0 could
-    # already be in use (e.g. by the Docker bridge)
+    # Get the next hop for the IP address.
+    next_hop = next_hops[ip.version]
+
+    network = IPNetwork(IPAddress(ip))
+    ep = Endpoint(hostname=hostname,
+                  orchestrator_id=ORCHESTRATOR_ID,
+                  workload_id=container_id,
+                  endpoint_id=uuid.uuid1().hex,
+                  state="active",
+                  mac=None)
+    if network.version == 4:
+        ep.ipv4_nets.add(network)
+        ep.ipv4_gateway = next_hop
+    else:
+        ep.ipv6_nets.add(network)
+        ep.ipv6_gateway = next_hop
+
+    # Create the veth, move into the container namespace, add the IP and
+    # set up the default routes.
     pid = info["State"]["Pid"]
-    endpoint = netns.set_up_endpoint(ip=ip,
-                                     hostname=hostname,
-                                     orchestrator_id=ORCHESTRATOR_ID,
-                                     workload_id=container_id,
-                                     cpid=pid,
-                                     next_hop_ips=next_hops,
-                                     veth_name=interface,
-                                     proc_alias="/proc")
+    netns.create_veth(ep.name, ep.temp_interface_name)
+    netns.move_veth_into_ns(pid, ep.temp_interface_name, interface)
+    netns.add_ip_to_ns_veth(pid, ip, interface)
+    netns.add_ns_default_route(pid, next_hop, interface)
 
-    # Register the endpoint
-    client.set_endpoint(endpoint)
+    # Grab the MAC assigned to the veth in the namespace.
+    ep.mac = netns.get_ns_veth_mac(pid, interface)
 
-    print "IP %s added to %s" % (ip, container_name)
+    # Register the endpoint with Felix.
+    client.set_endpoint(ep)
 
     # Let the caller know what endpoint was created.
-    return endpoint
+    return ep
 
 
 def container_remove(container_name):
@@ -215,7 +233,7 @@ def container_remove(container_name):
                 client.unassign_address(pool, ip)
 
     # Remove the endpoint
-    netns.remove_endpoint(endpoint.endpoint_id)
+    netns.remove_veth(endpoint.name)
 
     # Remove the container from the datastore.
     client.remove_workload(hostname, ORCHESTRATOR_ID, workload_id)
@@ -223,6 +241,8 @@ def container_remove(container_name):
     print "Removed Calico interface from %s" % container_name
 
 
+# TODO: If container created with IPv4 and then add IPv6 address, do we set up the
+# default route for IPv6 correctly (code read suggests not).
 def container_ip_add(container_name, ip, interface):
     """
     Add an IP address to an existing Calico networked container.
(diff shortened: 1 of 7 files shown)


tomdee:
Is your initial PR comment above wrong?

calico_containers/libnetwork_plugin/docker_plugin.py
```diff
@@ -347,31 +347,31 @@ def backout_ip_assignments(cnm_ep):
...
-               timeout=IP_CMD_TIMEOUT)
+def create_veth(endpoint):
+    """
+    Create the veth, and configure the MAC address (since libnetwork does not
+    do this).
+    :param endpoint: The Endpoint being configured.
+    """
+    netns.create_veth(endpoint)
 
     # Set the mac as libnetwork doesn't do this for us.
```
tomdee: I don't think it makes sense _not_ to be this.

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -19,11 +21,21 @@
...
 
+from pycalico.util import generate_cali_interface_name
+
+
 VETH_NAME = "eth1"
-"""The name to give to the veth in the target container's namespace. Default
-to eth1 because eth0 could be in use"""
+"""
+The name to give to the veth in the target container's namespace. Default
+to eth1 because eth0 could be in use.
```
tomdee: This name is only used when netns.py is used to move the veth into the namespace and _not_ when libnetwork is used.

calico_containers/pycalico/netns.py
```diff
@@ -105,13 +93,10 @@ def remove_ip_from_interface(container_pid, ip, interface_name,
...
-                         "addr": ip,
-                         "device": interface_name},
-                      shell=True)
+        ns.check_call(["ip", "-%s" % ip.version, "addr", "del",
+                       "%s/%s" % (ip, PREFIX_LEN[ip.version]),
+                       "dev", "%(device)s" % interface_name],
+                      timeout=IP_CMD_TIMEOUT)
 
 
 def set_up_endpoint(ip, hostname, orchestrator_id, workload_id, cpid, next_hop_ips,
```
tomdee: This method is wrong and it would be nice to clean it up.

It shouldn't be creating an Endpoint object - the caller should be doing that. The netns code shouldn't depend on the Endpoint class.

calico_containers/pycalico/netns.py
```diff
@@ -141,76 +126,111 @@ def set_up_endpoint(ip, hostname, orchestrator_id, workload_id, cpid, next_hop_i
...
-    if network.version == 4:
-        ep.ipv4_nets.add(network)
-        ep.ipv4_gateway = next_hop
-    else:
-        ep.ipv6_nets.add(network)
-        ep.ipv6_gateway = next_hop
     return ep
 
 
+def remove_endpoint(endpoint):
```
tomdee: Why?

calico_containers/pycalico/netns.py
```diff
@@ -284,24 +304,44 @@ def __exit__(self, _type, _value, _traceback):
             _log.exception("Failed to remove link: %s", self.nsn_dir)
         return False
 
-    def check_call(self, command, shell=False):
+    def check_call(self, command, shell=False, timeout=IP_CMD_TIMEOUT):
```
tomdee: Get rid of all the shell stuff here.

calico_containers/pycalico/netns.py
```diff
@@ -284,24 +304,44 @@ def __exit__(self, _type, _value, _traceback):
...
+        :param timeout: Command timeout in seconds.
         """
         _log.debug("Run command: %s", command)
-        check_call("ip netns exec %s %s" % (self.name, command), shell=shell)
+        check_call(self._get_nets_command(command),
+                   shell=shell,
+                   timeout=timeout)
 
-    def check_output(self, command, shell=False):
+    def check_output(self, command, shell=False, timeout=IP_CMD_TIMEOUT):
```
tomdee: Get rid of the timeout param and always use the fixed timeout below

tomdee:
I've finished looking at this for now. Assign it back to me when you want me to take another look

calico_containers/calico_ctl/container.py
```diff
@@ -223,6 +242,8 @@ def container_remove(container_name):
     print "Removed Calico interface from %s" % container_name
 
 
+# TODO: If container created with IPv4 and then add IPv6 address, do we set up the
```
tomdee: Hopefully Tim will find out as part of his testing...

calico_containers/libnetwork_plugin/docker_plugin.py
```diff
@@ -347,31 +347,29 @@ def backout_ip_assignments(cnm_ep):
...
-    if rc != 0:
-        app.logger.warn("Failed to delete veth %s", ep.name)
+    netns.set_veth_mac(endpoint.temp_interface_name, FIXED_MAC)
+
+
+def remove_veth(endpoint):
+    """
+    Best effort removal of veth, logging if removal fails.
+    :param endpoint: The Endpoint tied to the veth.
+    :return:
```
tomdee: Delete this :return: line if we don't return anything

calico_containers/pycalico/datastore_datatypes.py
```diff
@@ -204,7 +209,6 @@ def to_json(self):
         json_dict = {"state": self.state,
                      "name": self.name,
                      "mac": self.mac,
-                     "container:if_name": self.if_name,
```
tomdee: You're getting rid of storing the interface name in the endpoint? Isn't that going to break things for some people?

calico_containers/calico_ctl/container.py
robbrockbank: @TrimBiggs see above.

calico_containers/pycalico/netns.py
```diff
@@ -61,191 +60,153 @@ def setup_logging(logfile, level=logging.INFO):
...
     """
-    iface = IF_PREFIX + ep_id[:11]
-    call("ip link delete %s" % iface, shell=True)
+    # The veth removal is best effort. If it fails then just log.
+    check_call(['ip', 'link', 'del', veth_name_host], timeout=IP_CMD_TIMEOUT)
 
 
-def add_ip_to_interface(container_pid, ip, interface_name,
-                        proc_alias=PROC_ALIAS):
+#TODO Do we need to make proc_alias configurable anymore?
```
tomdee: No, I think not. We only support running this from the root namespace.

calico_containers/pycalico/datastore_datatypes.py
robbrockbank: I don't think so - but perhaps I'm wrong.

This used to be used in the reinstating endpoint which was a feature of powerstrip... we don't reference that any more.

calico_containers/calico_ctl/container.py
TrimBiggs: Looks like it works with a quick test in single host environment.

calico_containers/pycalico/netns.py
```diff
@@ -61,191 +60,153 @@ def setup_logging(logfile, level=logging.INFO):
...
     """
-    with NamedNamespace(container_pid, proc=proc_alias) as ns:
-        ns.check_call("ip -%(version)s addr add %(addr)s/%(len)s "
-                      "dev %(device)s" %
-                        {"version": ip.version,
-                         "len": PREFIX_LEN[ip.version],
-                         "addr": ip,
-                         "device": interface_name},
-                      shell=True)
+    # Configure the container end.  To do this we create a temporary link from
```
tomdee: As per the TODO and comment above, does that mean we can get rid of this NamedNamespace thing?

calico_containers/pycalico/datastore_datatypes.py
tomdee: Yep, makes sense.

calico_containers/calico_ctl/container.py
tomdee: Don't you need to update/remove the comment based on what Tim has said?

calico_containers/pycalico/netns.py
```diff
@@ -61,201 +51,153 @@ def setup_logging(logfile, level=logging.INFO):
...
-    ep_id = uuid.uuid1().hex
-
-    iface = generate_cali_interface_name(IF_PREFIX, ep_id)
-    iface_tmp = generate_cali_interface_name("tmp", ep_id)
-
-    # Provision the networking.  We create a temporary link from the proc
-    # alias to the /var/run/netns to provide a named namespace.  If we don't
-    # do this, when run from the calico-node container the PID of the
+    # Configure the container end.  To do this we create a temporary link from
+    # the proc alias to the /var/run/netns to provide a named namespace.  If we
```
tomdee: Er... What's a proc alias?

tomdee:
OK, just one comment tweak then LGTM

calico_containers/pycalico/netns.py
robbrockbank: Oh poop I'd removed that entire comment in a previous pull request.  Don't think this comment should even be here.  I'm removing.

PR #353, opened 2015-07-21
We don't allow the IP to be passed in so remove it

tomdee:


robbrockbank:
LGTM.   (once it builds)

PR #355, opened 2015-07-21
make fast-st runs python version of calicoctl

tomdee:


Makefile
```diff
@@ -89,8 +89,11 @@ st: binary calico_containers/busybox.tar calico_containers/calico-node.tar run-e
 	dist/calicoctl checksystem --fix
 	nosetests $(ST_TO_RUN) -sv --nologcapture --with-timer
 
-fast-st: binary calico_containers/busybox.tar calico_containers/calico-node.tar run-etcd run-consul
-	nosetests $(ST_TO_RUN) -sv --nologcapture --with-timer -a '!slow'
+fast-st: calico_containers/busybox.tar calico_containers/calico-node.tar run-etcd run-consul
+	# This runs the tests by calling python directory without using the
```
robbrockbank: should we be running checksystem --fix here too?

robbrockbank:
LGTM

Makefile
tomdee: I don't think so. This is a make target for developers. You only need to run --fix once per reboot, so we shouldn't slow down every fast test run with it.

PR #359, opened 2015-07-22
Missed instance of profile bug

L-MA:


calico_containers/calico_ctl/profile.py
```diff
@@ -105,7 +105,7 @@ def validate_arguments(arguments):
     # Validate IPs
     cidr_ok = True
     for arg in ["<SRCCIDR>", "<DSTCIDR>"]:
-        if arguments.get(arg):
+        if arguments.get(arg) is not None:
```
TrimBiggs: This isn't necessary, but it is fine to keep in.

calico_containers/calico_ctl/endpoint.py
```diff
@@ -67,8 +67,8 @@ def validate_arguments(arguments):
 
     # Validate Profiles
     profile_ok = True
-    if "<PROFILES>" in arguments:
-        profiles = arguments.get("<PROFILES>")
+    profiles = arguments.get("<PROFILES>")
+    if profiles is not None:
```
TrimBiggs: This could be `if profiles`, but the way that it is now works fine too since the value of `<PROFILES>` will be set to `None` if it is not specified, which will be interpreted as `False` by python.

TrimBiggs:
LGTM

PR #362, opened 2015-07-22
Install script for running calicoctl natively

TrimBiggs:
Add make command for installing dependencies
Make sure st/fast-st install deps as well


tomdee:
You've added a dependency on developers using a debian based linux. That's not acceptable. We should discuss before you rewrite this.

PR #364, opened 2015-07-23
Network independent IP retrieval

alexhersh:
Previous implementation of IP retrieval pings Google's DNS Server, this uses legacy RegEx formulas to  parse `ip addr` output

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -56,6 +56,7 @@ from utils import print_paragraph
 from pycalico.datastore_datatypes import BGPPeer
 from pycalico.datastore import (ETCD_AUTHORITY_ENV,
                                 ETCD_AUTHORITY_DEFAULT)
+from pycalico.util import get_host_ips
 from checksystem import check_system
 from netaddr import IPAddress
 from prettytable import PrettyTable
@@ -190,7 +191,7 @@ def node_start(node_image, log_dir, ip, ip6, as_num, detach, kubernetes):
 
     # Get IP address of host, if none was specified
     if not ip:
-        ips = get_host_ips(4)
+        ips = get_host_ips(exclude=["docker0"])
         try:
             ip = ips.pop()
         except IndexError:
@@ -365,31 +366,6 @@ def node_bgppeer_show(version):
         print "No IPv%s BGP Peers defined for this node.\n" % version
 
 
-def get_host_ips(version):
-    """
-    Gets all IP addresses assigned to this host.
-
-    This function is fail-safe and will return an empty array instead of
-    raising any exceptions.
-
-    :param version: Desired version of IP addresses. Can be 4 or 6
-    :return: List of string representations of IP Addresses.
-    """
-    ip_addrs = []
-    addr_lines_raw = check_output(["ip", "-o", "-%d" % (version), "addr"]).strip().split("\n")
-    for addr_line_raw in addr_lines_raw:
-        # Each addr_line_raw is a single line from the ip addr command
-        values = addr_line_raw.split()
-        if 'lo' not in values and 'docker0' not in values:
-            try:
-                # Extract the IP, ensure its valid
-                ip_addrs.append(str(netaddr.IPNetwork(values[3]).ip))
-            except (OSError, CalledProcessError, IndexError, netaddr.AddrFormatError):
-                # Ignore any encountered errors to maintain fail-safe
-                pass
-    return ip_addrs
-
-
 def warn_if_unknown_ip(ip, ip6):
     """
     Prints a warning message if the IP addresses are not assigned to interfaces
@@ -399,11 +375,11 @@ def warn_if_unknown_ip(ip, ip6):
     :param ip6: IPv6 address which should be present on the host.
     :return: None
     """
-    if ip not in get_host_ips(4):
+    if ip and ip not in get_host_ips(version=4, exclude=["docker0"]):
         print "WARNING: Could not confirm that the provided IPv4 address is assigned" \
               " to this host."
 
-    if ip6 and ip6 not in get_host_ips(6):
+    if ip6 and ip6 not in get_host_ips(version=6, exclude=["docker0"]):
         print "WARNING: Could not confirm that the provided IPv6 address is assigned" \
               " to this host."
 
--- a/calico_containers/integrations/kubernetes/calico_kubernetes.py
+++ b/calico_containers/integrations/kubernetes/calico_kubernetes.py
@@ -16,7 +16,7 @@ print("Using ETCD_AUTHORITY=%s" % os.environ[ETCD_AUTHORITY_ENV])
 
 from calico_ctl.container import container_add
 from pycalico.datastore import IF_PREFIX
-from pycalico.util import generate_cali_interface_name
+from pycalico.util import generate_cali_interface_name, get_host_ips
 
 CALICOCTL_PATH = os.environ.get('CALICOCTL_PATH', '/usr/bin/calicoctl')
 print("Using CALICOCTL_PATH=%s" % CALICOCTL_PATH)
@@ -98,20 +98,22 @@ class NetworkPlugin(object):
         return ep
 
     def _get_node_ip(self):
-        """Determine the IP for the host node.
-
-        This hits a well-known IP (the Google public DNS).
+        """ 
+        Determine the IP for the host node.
         """
-        # We'd like to use the k8s API here, but it doesn't work in the Vagrant
-        # case since the node's name is set to its IP (so we have no way of
-        # getting from hostname=>IP).
-        # TODO: do this more reliably by parsing 'ip
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(('8.8.8.8', 80))
-        ip = s.getsockname()[0]
-        print('Node IP: %s' % ip)
-        s.close()
-        return ip
+        # Compile list of addresses on network, return the first entry.
+        # Try IPv4 and IPv6.
+        addrs = get_host_ips(version=4) or get_host_ips(version=6)
+
+        try:
+            addr = addrs[0]
+            print('Using IP Address %s' % (addr))
+            return addr
+        except IndexError:
+            # If both get_host_ips return empty lists, print message and exit.
+            print('No Valid IP Address Found for Host - cannot configure networking for pod %s' % (self.pod_name))
+            sys.exit(1)
+
 
     def _read_docker_ip(self):
         """Get the IP for the pod's infra container."""
--- a/calico_containers/pycalico/tests/test_util.py
+++ b/calico_containers/pycalico/tests/test_util.py
@@ -1,0 +1,124 @@
+# Copyright 2015 Metaswitch Networks
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+from mock import patch, Mock
+from subprocess import CalledProcessError, check_output
+from pycalico.util import get_host_ips
+
+MOCK_IP_ADDR = \
+"""
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default 
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
+    link/ether 08:00:27:73:c8:d0 brd ff:ff:ff:ff:ff:ff
+    inet 172.24.114.18/24 brd 172.24.114.255 scope global eth0
+       valid_lft forever preferred_lft forever
+    inet6 2620:104:4008:69:8d7c:499f:2f04:9e55/64 scope global temporary dynamic 
+       valid_lft 603690sec preferred_lft 84690sec
+    inet6 2620:104:4008:69:a00:27ff:fe73:c8d0/64 scope global dynamic 
+       valid_lft 604698sec preferred_lft 86298sec
+    inet6 fe80::a00:27ff:fe73:c8d0/64 scope link 
+       valid_lft forever preferred_lft forever
+3: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default 
+    link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
+    inet 172.17.42.1/24 scope global docker0
+       valid_lft forever preferred_lft forever
+    inet6 fe80::188f:d6ff:fe1f:1482/64 scope link 
+       valid_lft forever preferred_lft forever
+"""
+
+MOCK_IP_ADDR_LOOPBACK = \
+"""
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+"""
+
+
+class TestUtil(unittest.TestCase):
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_standard(self, m_check_output):
+    # Test IPv4
+    m_check_output.return_value = MOCK_IP_ADDR
+    addrs = get_host_ips(version=4)
+    m_check_output.assert_called_once_with(["ip", "-4", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['172.24.114.18/24', '172.17.42.1/24'])
+
+    # Test IPv6
+    addrs = get_host_ips(version=6)
+    m_check_output.assert_called_once_with(["ip", "-6", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['2620:104:4008:69:8d7c:499f:2f04:9e55/64',
+                  '2620:104:4008:69:a00:27ff:fe73:c8d0/64',
+                  'fe80::a00:27ff:fe73:c8d0/64',
+                  'fe80::188f:d6ff:fe1f:1482/64'])
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_loopback_only(self, m_check_output):
+    # Test Loopback
+    m_check_output.return_value = MOCK_IP_ADDR_LOOPBACK
+    addrs = get_host_ips(version=4)
+    m_check_output.assert_called_once_with(["ip", "-4", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, [])
+
+    addrs = get_host_ips(version=6)
+    m_check_output.assert_called_once_with(["ip", "-6", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, [])
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_exclude(self, m_check_output):
+    # Exclude "docker0"
+    m_check_output.return_value = MOCK_IP_ADDR
+    addrs = get_host_ips(version=4, exclude=["docker0"])
+    m_check_output.assert_called_once_with(["ip", "-4", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['172.24.114.18/24'])
+
+    addrs = get_host_ips(version=6, exclude=["docker0"])
+    m_check_output.assert_called_once_with(["ip", "-6", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['2620:104:4008:69:8d7c:499f:2f04:9e55/64',
+                  '2620:104:4008:69:a00:27ff:fe73:c8d0/64',
+                  'fe80::a00:27ff:fe73:c8d0/64'])
+
+    # Exclude empty list
+    addrs = get_host_ips(version=4, exclude=[""])
+    m_check_output.assert_called_once_with(["ip", "-4", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['172.24.114.18/24', '172.17.42.1/24'])
+
+    addrs = get_host_ips(version=6, exclude=[""])
+    m_check_output.assert_called_once_with(["ip", "-6", "addr"])
+    m_check_output.reset_mock()
+    self.assertEquals(addrs, ['2620:104:4008:69:8d7c:499f:2f04:9e55/64',
+                  '2620:104:4008:69:a00:27ff:fe73:c8d0/64',
+                  'fe80::a00:27ff:fe73:c8d0/64',
+                  'fe80::188f:d6ff:fe1f:1482/64'])
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_fail_check_output(self, m_check_output):
+    m_check_output.side_effect = CalledProcessError(returncode=1, cmd=check_output(["ip", "-4", "addr"]))
+    with self.assertRaises(SystemExit):
+        addrs = get_host_ips(version=4)
--- a/calico_containers/pycalico/util.py
+++ b/calico_containers/pycalico/util.py
@@ -1,3 +1,20 @@
+import sys
+import re
+from netaddr import IPNetwork, AddrFormatError
+from subprocess import check_output, CalledProcessError
+
+"""
+Compile Regexes
+"""
+# Splits into groups that start w/ no whitespace and contain all lines below that start w/ whitespace
+INTERFACE_SPLIT_RE = re.compile(r'(\d+:.*(?:\n\s+.*)+)')
+# Grabs interface name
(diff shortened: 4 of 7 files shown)


calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+    :return: List of string representations of IP Addresses.
+    """
+
+    ip_addrs = []
+
+    # Call `ip addr`
+    if interface is None:
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr"])    
+    else:    # Add argument to `ip addr` command if you are selecting an interface
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr", "show", "%s" % (interface)])
```
djosborne: You can combine both paths of this if statement if you change it to:
`ip_addr_output = check_output(["ip", "-%d" % (version), "addr", "show", interface)`

And make the function's default `interface=""` instead of `interface=None`

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+
+    # separate interface blocks from ip addr output and iterate
+    for iface_block in INTERFACE_SPLIT.findall(ip_addr_output):
+        # exclude certain interfaces
+        if GET_IFACE.search(iface_block).group(1) not in exclude :
+            # iterate through Addresses on interface
+            for address in GET_IP.findall(iface_block):
+                # eliminate non addresses and append
+                try:
+                    ip_addrs.append(address)
```
djosborne: This doesn't need to be in a try/except block anymore since the `append` method doesn't throw any of these. However, You need to exception handle where appropriate. I believe `check_output` throws the `CalledProcessError` if the command doesn't exist, so you should catch that up there.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+
+    # Select Regex for IPv6 or IPv4
+    GET_IP= GET_IPV4 if version==4 else GET_IPV6
+
+    # separate interface blocks from ip addr output and iterate
+    for iface_block in INTERFACE_SPLIT.findall(ip_addr_output):
+        # exclude certain interfaces
+        if GET_IFACE.search(iface_block).group(1) not in exclude :
+            # iterate through Addresses on interface
+            for address in GET_IP.findall(iface_block):
```
djosborne: I don't think dynamically setting GET_IP is very straightforward. It'd be more obvious what's going on if you branched an if statement down here:
```
if version == 4:
    addresses = GET_IPV4.findall(iface_block)
elif verison == 6:
    addresses = GET_IPV6.findall(iface_block)
```

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +96,19 @@ def _configure_interface(self):
         return ep
 
     def _get_node_ip(self):
-        """Determine the IP for the host node.
-
-        This hits a well-known IP (the Google public DNS).
+        """ Determine the IP for the host node.
```
caseydavenport: Docstring here is a bit weirdly formatted now.  

Also no need to say "Does not use public internet" - that should be the default assumption :)

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +96,19 @@ def _configure_interface(self):
...
-        # case since the node's name is set to its IP (so we have no way of
-        # getting from hostname=>IP).
-        # TODO: do this more reliably by parsing 'ip
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(('8.8.8.8', 80))
-        ip = s.getsockname()[0]
-        s.close()
-        return ip
+        # prefer IPv4, but will check IPv6
+        ipv4_addresses = get_host_ips(version=4,exclude=["lo","docker0"])
```
caseydavenport: Make sure you're sticking appropriate spaces after your commas.

e.g.

```
ipv4_addresses = get_host_ips(version=4, exclude=["lo", "docker0"])
```

calico_containers/pycalico/util.py
```diff
@@ -1,3 +1,19 @@
+from netaddr import IPNetwork, AddrFormatError
+from subprocess import check_output, CalledProcessError
+import re
+
+"""
+Compile Regex's
```
caseydavenport: Typo: Regex's -> Regexes

calico_containers/pycalico/util.py
```diff
@@ -1,3 +1,19 @@
...
+"""
+# splits into groups that start w/ no whitespace and contain all lines below that start w/ whitespace
```
caseydavenport: As a general rule, start all comments with a capital letter, end each one with punctuation, make them complete sentences.

calico_containers/pycalico/util.py
```diff
@@ -1,3 +1,19 @@
...
+INTERFACE_SPLIT = re.compile(r'(\d+:.*(?:\n\s+.*)+)')
+# grab interface name
```
caseydavenport: grab -> Grabs

calico_containers/pycalico/util.py
caseydavenport: What is the use case for providing the `interface=` argument?

Is there a reason we need this argument now?  It seems to me that it is never used.  If this is so, I think we should not include this argument.  It means less code to write, less code to maintain, less code to test.

If we ever need this functionality in the future, then we can add it at that point.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr"])    
+    else:    # Add argument to `ip addr` command if you are selecting an interface
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr", "show", "%s" % (interface)])
+
+    # Select Regex for IPv6 or IPv4
+    GET_IP= GET_IPV4 if version==4 else GET_IPV6
+
+    # separate interface blocks from ip addr output and iterate
+    for iface_block in INTERFACE_SPLIT.findall(ip_addr_output):
+        # exclude certain interfaces
```
caseydavenport: General style point: stick newlines before comments.

calico_containers/pycalico/util.py
caseydavenport: I don't think this is too confusing, and does cut down on a few lines of code.  I'd lean towards keeping this.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+    ip_addrs = []
+
+    # Call `ip addr`
+    if interface is None:
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr"])    
+    else:    # Add argument to `ip addr` command if you are selecting an interface
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr", "show", "%s" % (interface)])
+
+    # Select Regex for IPv6 or IPv4
+    GET_IP= GET_IPV4 if version==4 else GET_IPV6
```
caseydavenport: Stick a space after GET_IP and before the equals

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+
+    # separate interface blocks from ip addr output and iterate
+    for iface_block in INTERFACE_SPLIT.findall(ip_addr_output):
+        # exclude certain interfaces
+        if GET_IFACE.search(iface_block).group(1) not in exclude :
+            # iterate through Addresses on interface
+            for address in GET_IP.findall(iface_block):
+                # eliminate non addresses and append
```
caseydavenport: What does it mean when you say "eliminate non addresses" - looks like the code following it just appends.

calico_containers/pycalico/util.py
alexhersh: adding an empty input into `check_output` (as `""`) broke the command for me. Maybe I missed something. Will check again. 

I will remove the `interface` input.

calico_containers/pycalico/util.py
alexhersh: I meant to add a check to verify that it was a valid IP address. That check fell through the cracks. Mainly I need the `AddrFormatError` check

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +96,19 @@ def _configure_interface(self):
         return ep
 
     def _get_node_ip(self):
-        """Determine the IP for the host node.
-
-        This hits a well-known IP (the Google public DNS).
+        """ 
+            Determine the IP for the host node.
```
djosborne: indentation is off here

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +96,19 @@ def _configure_interface(self):
...
-        # case since the node's name is set to its IP (so we have no way of
-        # getting from hostname=>IP).
-        # TODO: do this more reliably by parsing 'ip
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(('8.8.8.8', 80))
-        ip = s.getsockname()[0]
-        s.close()
-        return ip
+        # Compile list of IPv4 addresses on network, return the first entry
+        ipv4_addresses = get_host_ips(version=4)
```
djosborne: I think we want to ignore the loopback address here:
```
ipv4_addresses = get_host_ips(version=4, exclude=["lo"])
```

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,40 @@ def generate_cali_interface_name(prefix, ep_id):
     if len(prefix) > 4:
         raise ValueError('Prefix must be 4 characters or less.')
     return prefix + ep_id[:11]
+
+
+def get_host_ips(version=4, exclude=[]):
+    """
+    Gets all IP addresses assigned to this host.
+
+    This function is fail-safe and will return an empty array instead of
```
caseydavenport: You'll want to include that this function doesn't return loopback addresses in the docstring.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,40 @@ def generate_cali_interface_name(prefix, ep_id):
...
+        # Exclude certain interfaces.
+        match = IFACE_RE.match(iface_block)
+        if match and (match.group(1) not in exclude) :
+            # Iterate through Addresses on interface.
+            for address in IP_RE.findall(iface_block):
+                # Eliminate loopback and non addresses and append.
+                try:
+                    if not IPNetwork(address).is_loopback():
+                        ip_addrs.append(address)
+                except (AddrFormatError):
```
caseydavenport: Do we ever expect this to happen?  Seems like if we're hitting this except block, then there is something wrong with our regex.

We should either include logging here so we can see when something is going wrong, or remove the except block and put faith in our regex.  I'd lean towards removing the try/except.

calico_containers/integrations/kubernetes/calico_kubernetes.py
caseydavenport: Don't indent the text inside docstrings.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +96,19 @@ def _configure_interface(self):
...
+        
+        # If no IPv4 addresses exist, check IPv6, return the first entry
+        if not ipv4_addresses:
+            return get_host_ips(version=6)[0]
```
djosborne: It's a bit dangerous to blindly access an index of a list when we can't guarentee that there's a value there. We probably want to wrap this in a `try/except IndexError` block.

However, the kubernetes plugin doesn't have the best error handling currently - I'm not sure what the best course of action would be in the except statement. Maybe `print("Couldn't determine the host's IP")` and then `sys.exit(1)`, just like we do in calico-docker? I'll defer to @caseydavenport for how we might want to handle this.

calico_containers/integrations/kubernetes/calico_kubernetes.py
djosborne: Unless of course that function already does ignore loopbacks, which it does appear to. So ignore this comment.

calico_containers/pycalico/util.py
alexhersh: It is mainly to protect against oddities in `ip addr` output. This is a remnant of my previous work (using a plugin not RegExes), but I suppose the RegEx shouldn't pull in false addresses. Unless someone throws a 999.999.999.999/99 address in there....

calico_containers/integrations/kubernetes/calico_kubernetes.py
caseydavenport: There's no guarantee that get_host_ips(version=6) will return any addresses here, so just blindly grabbing index 0 might raise an IndexError.

Might want to try something like this:

```
# Get the IPs for this node.  Try getting IPv4 addresses, if there are none use IPv6 addresses.
addrs = get_host_ips(version=4) or get_host_ips(version=6)

if not addrs:
    # No configured IP addresses - return None.
    return None

# An address exists.
return addrs[0]
```

Another options might be 
```
try:
    # Try to get an address
    return addrs[0]
except IndexError:
    # No addresses found.
    return None
```

caseydavenport: Right - I missed where this code was.

Returning None is probably not the right thing to do, since the calling code is just going to fall over.
Probably best to include a good error log and sys.exit(1)

calico_containers/pycalico/util.py
caseydavenport: ip addr should be trustworthy (until proven otherwise).

Nothing should be able to configure an invalid IP address into the Linux IP table.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +95,24 @@ def _configure_interface(self):
...
+        try:
+            addr = addrs[0]
+            print('Using IP Address %s' % (addr))
+            return addr
+        # If both get_host_ips return empty lists, print msg and return None
+        except IndexError:
+            print('No Valid IP Address Found for Host')
+            return None
+
+        return None
```
caseydavenport: This return is redundant - It won't ever be hit because of the try / except above.

Also, I think we shouldn't be returning None from this method at all.  If we can't find an IP address, we should log a clear error and sys.exit

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +95,24 @@ def _configure_interface(self):
...
-        return ip
+        # Compile list of addresses on network, return the first entry.
+        # Try IPv4 and IPv6.
+        addrs = get_host_ips(version=4) or get_host_ips(version=6)
+
+        try:
+            addr = addrs[0]
+            print('Using IP Address %s' % (addr))
+            return addr
+        # If both get_host_ips return empty lists, print msg and return None
```
caseydavenport: This comment should be inside the except clause, rather than before it.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +26,39 @@ def generate_cali_interface_name(prefix, ep_id):
...
+    :param exclude: list of interfaces (strings) to ignore (ex. ["lo","docker0"])
+    :return: List of string representations of IP Addresses.
+    """
+    ip_addrs = []
+
+    # Select Regex for IPv6 or IPv4.
+    IP_RE = IPV4_RE if version==4 else IPV6_RE
+
+    # Call `ip addr`.
+    ip_addr_output = check_output(["ip", "-%d" % (version), "addr"])
```
caseydavenport: Probably want to wrap this in a try/except which catches CalledProcessError.

calico_containers/calico_ctl/node.py
```diff
@@ -399,11 +375,11 @@ def warn_if_unknown_ip(ip, ip6):
     :param ip6: IPv6 address which should be present on the host.
     :return: None
     """
-    if ip not in get_host_ips(4):
+    if ip not in get_host_ips(exclude=["docker0"]):
```
caseydavenport: These if statements should be symmetric.

Either:
1) They are both gated on the definition of ip/ip6, or
2) Neither are gated on the definition of ip/ip6, and the check is performed for both of them always.

If option 2, then you should clean up the if statements to be a bit more clear.

e.g.
```
if ip6 and (ip6 not in get_host_ips(version=6, exclude=["docker0"])):
  ....
```

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -95,19 +95,22 @@ def _configure_interface(self):
...
+        # Try IPv4 and IPv6.
+        addrs = get_host_ips(version=4) or get_host_ips(version=6)
+
+        try:
+            addr = addrs[0]
+            print('Using IP Address %s' % (addr))
+            return addr
+        except IndexError:
+            # If both get_host_ips return empty lists, print message and exit.
+            print('No Valid IP Address Found for Host')
```
caseydavenport: Worth saying something in this log like "No Valid host IP address found - cannot configure networking for pod %s"

calico_containers/pycalico/util.py
```diff
@@ -10,3 +27,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+    """
+    ip_addrs = []
+
+    # Select Regex for IPv6 or IPv4.
+    IP_RE = IPV4_RE if version is 4 else IPV6_RE
+
+    # Call `ip addr`.
+    try:
+        ip_addr_output = check_output(["ip", "-%d" % (version), "addr"])
+    except CalledProcessError:
```
Symmetric: I think this can hit an OSError too.

calico_containers/pycalico/util.py
```diff
@@ -10,3 +27,43 @@ def generate_cali_interface_name(prefix, ep_id):
...
+        print "Call to 'ip addr' Failed"
+        sys.exit(1)
+
+    # Separate interface blocks from ip addr output and iterate.
+    for iface_block in INTERFACE_SPLIT_RE.findall(ip_addr_output):
+        # Exclude certain interfaces.
+        match = IFACE_RE.match(iface_block)
+        if match and (match.group(1) not in exclude):
```
Symmetric: superfluous brackets?

alexhersh: Do you mean the `match` line? The paren are there for clarity not function. removable if desired.

Symmetric: PEP8 considers superfluous brackets to be an error. Should be highlighted in Pycharm? (LMK if you need more detail on this)

caseydavenport: Agreed that this particular IF statement probably doesn't need the parens, however, PEP8 does say the following -"Some other good reasons to ignore a particular guideline: When applying the guideline would make the code less readable, even for someone who is used to reading code that follows this PEP."

Good to remember that blindly following PEP is not always the best option.

djosborne:
Just a note - you're definitely going to want to squash this before you merge it. Anyone on the team can help you when you're ready to do that

calico_containers/pycalico/util.py
```diff
@@ -10,3 +27,43 @@ def generate_cali_interface_name(prefix, ep_id):
     if len(prefix) > 4:
         raise ValueError('Prefix must be 4 characters or less.')
     return prefix + ep_id[:11]
+
+
+def get_host_ips(version=4, exclude=[]):
```
caseydavenport: Didn't think of this before - this function definitely needs to be unit tested.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+from mock import patch, Mock, call, ANY
+from calico_ctl.node import _attach_and_stream
```
caseydavenport: Could you clean up the imports? Only import what you need.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+from docker import Client
+import sys
+import signal
+import os
+from pycalico.util import get_host_ips
+
+mock_ip_addr = \
```
caseydavenport: Its typical to make global definitions like this in all caps - makes it easier to spot.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_standard(self, m_check_output):
+    # Test IPv4
+    m_check_output.return_value = mock_ip_addr
+    addrs = get_host_ips(version=4)
+    self.assertEquals(addrs, ['172.24.114.18/24', '172.17.42.1/24'])
+
+    # Test IPv6
+    m_check_output.return_value = mock_ip_addr
```
caseydavenport: You don't need to set the return_value twice - you can get rid of this line.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_loopback_only(self, m_check_output):
+    # Test Loopback
+    m_check_output.return_value = mock_ip_addr_loopback
+    addrs = get_host_ips(version=4)
+    self.assertEquals(addrs, [])
+
+    addrs = get_host_ips(version=6)
+    self.assertEquals(addrs, [])
+
```
caseydavenport: Only one newline between methods.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
+# Copyright 2015 Metaswitch Networks
```
caseydavenport: Missing test case when check_output() throws a CalledProcessError or OSError

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+class TestUtil(unittest.TestCase):
+
+  @patch("pycalico.util.check_output", autospec=True)
+  def test_get_host_ips_standard(self, m_check_output):
+    # Test IPv4
+    m_check_output.return_value = mock_ip_addr
+    addrs = get_host_ips(version=4)
+    self.assertEquals(addrs, ['172.24.114.18/24', '172.17.42.1/24'])
```
caseydavenport: For each of these, you should be asserting that check_output is called with the expected arguments.

e.g.
```
m_check_output.assert_called_once_with(["ip", "-4", "addr"])
```

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,111 @@
...
+
+    m_check_output.return_value = mock_ip_addr
+    addrs = get_host_ips(version=6, exclude=["docker0"])
+    self.assertEquals(addrs, ['2620:104:4008:69:8d7c:499f:2f04:9e55/64',
+                  '2620:104:4008:69:a00:27ff:fe73:c8d0/64',
+                  'fe80::a00:27ff:fe73:c8d0/64'])
+
+    # Exclude empty list
+    m_check_output.return_value = mock_ip_addr
+    addrs = get_host_ips(version=4, exclude=[""])
```
caseydavenport: An empty list is just [ ].  What you've got here is a list with a single empty string as an entry.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,116 @@
...
+
+    addrs = get_host_ips(version=6, exclude=["docker0"])
+    m_check_output.assert_called_with(["ip", "-6", "addr"])
+    self.assertEquals(addrs, ['2620:104:4008:69:8d7c:499f:2f04:9e55/64',
+                  '2620:104:4008:69:a00:27ff:fe73:c8d0/64',
+                  'fe80::a00:27ff:fe73:c8d0/64'])
+
+    # Exclude empty list
+    addrs = get_host_ips(version=4, exclude=[""])
+    m_check_output.assert_called_with(["ip", "-4", "addr"])
```
caseydavenport: Since you are calling get_host_ips multiple times here, this assert is not really meaningful (since we've already asserted that m_check_output is called with this argument).

A better pattern to use when checking calls is the following:

```
m_check_output.assert_called_once_with( ... )
m_check_output.reset_mock()
```
This asserts that the mock was called with the correct args once and only once (if it is called twice, it will fail).  reset_mock() resets the call counters on the mock so that subsequent assert_called_once_with() calls will only see the latest call.

Another way to do this would be to use assert_has_calls(), which allows you to specify the complete list of calls.

caseydavenport: I'd go so far as to say you should probably never use assert_called_with, and instead use assert_called_once_with.  Its a stronger check.

calico_containers/pycalico/tests/test_util.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default 
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host 
+       valid_lft forever preferred_lft forever
+2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
+    link/ether 08:00:27:73:c8:d0 brd ff:ff:ff:ff:ff:ff
+    inet 172.24.114.18/24 brd 172.24.114.255 scope global eth0
```
djosborne: Not sure if I'm just being paranoid but - it might not be the best idea to post your computer's IP on github

caseydavenport: That is his private network address. (In the 172.24.0.0/16 sub)

"Addresses in the private space are not allocated to any specific organization and anyone may use these addresses without approval from a regional Internet registry. However, IP packets addressed from them cannot be transmitted through the public Internet."

https://en.wikipedia.org/wiki/Private_network

caseydavenport:
LGTM - will merge once semaphore passes.

PR #366, opened 2015-07-23
Fixes #363 calicoctl and Powerstrip adapter use of alternative socket

### LINKED ISSUE #363: [Powerstrip] `calicoctl node` fails if alternate unix socket in use
Failure seems to be in the checksystem module

spikecurtis:


tomdee:
LGTM

PR #369, opened 2015-07-23
Convert felix-invalid "/" characters into "_" characters

djosborne:
Replace forward-slash character with underscore to make kubernetes labels valid in Felix ip-tables rules. Also removed replacement of `-` characteres since Kubernetes now accepts them as valid characters.

In Kubernetes, label names are passed to Felix which use them as rule-names (for human readability). However, Kubernetes accepts the `/` character, which Felix will error on. Therefore, we need to encode the `/` character. 

This PR is a temporary fix which converts `/` to `_`, but will result in collisions if an identical chain-name is passed in where the character is replaced. For example, if Kubernetes creates `custom/rule` and then creates `custom_rule`, there will be an IP Tables error in calico. (Opening a seperate issue to come up with a better encoding)

### DIFF
--- a/calico_containers/integrations/kubernetes/calico_kubernetes.py
+++ b/calico_containers/integrations/kubernetes/calico_kubernetes.py
@@ -191,7 +191,7 @@ class NetworkPlugin(object):
 
         for pod in pods:
             print('Processing pod %s' % pod)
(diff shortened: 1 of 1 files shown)


caseydavenport:
LGTM

PR #371, opened 2015-07-24
Build Binary for Kube Plugin in Makefile

alexhersh:


djosborne:
@tomdee Can we get your approval on this? This makes calico_kubernetes an artifact of the circleci builds.

tomdee:
I would prefer that this _didn't_ go in. Not everybody wants a kubernetes binary! We should be splitting the kubernetes stuff out of this repo ASAP and then you can do whatever you want.

PR #375, opened 2015-07-24
Split out IPAM and BGP configuration in templates to allow more speci…

robbrockbank:
Split out BIRD templates into separate IPAM and BGP templates to allow more specific etcd watch prefixes ... to reduce confd churn.


tomdee:
LGTM - merging

PR #376, opened 2015-07-24
Fixes #346 Warns if ipip kernel module is not loaded.

### LINKED ISSUE #346: Warn if ipip module not loaded
I spent the morning trying to network two boot2docker VMs provisioned by docker-machine running the experimental branch of docker. Everything seemed to install ok, but my containers couldn't talk to each other or be pinged on their calico addresses, even within the same host.

After some investigation, it seems to be because boot2docker does not include the ipip kernel module. It took me all morning to figure this out - I wonder if a check could be added to calico to warn if the module is not present? Perhaps on running `calicoctl checksystem`?

Perhaps this issue should be moved to the main calico project, but I was following the instructions for the plug-in, and I'm not sure what belongs to what...

spikecurtis:


tomdee:
LGTM

PR #378, opened 2015-07-24
Adjusted regex expression for IP Address

djosborne:
Python regex expression was grouping the subnet with the IP Address. Adjusted the parenthesis to confirm that a subnet is present but only return the IP address


caseydavenport:
LGTM

PR #379, opened 2015-07-24
Fix path to calicoctl

djosborne:
The new vagrant scripts place calicoctl in `/usr/bin` so the instructions should assume its in the $PATH instead of the current directory


tomdee:
This is being handled elsewhere so closing. @robbrockbank FYI

PR #386, opened 2015-07-28
Converted calico_ctl into its own package.

djosborne:
Now, when calico-docker repo is installed with pip, in addition to getting a pycalico package, they will get a calico_ctl package. This is a current requirement by the kubernetes plugin, and will most likely be used by other packages in the future.


djosborne:
Discussed with the Team and decided against this for repo-structure reasons

PR #388, opened 2015-07-28
Switch build container to debian

Symmetric:
Our calicoctl binary depends on libc, and so we will bake in
a dependency on a newer distro by using something modern like
Ubuntu 14.04.

Switch to debian 7, which is new enough to meet our requirements,
while adding support for older distros.

(debian 7 comes with libc version 2.13-38+deb7u6).


tomdee:
I like it! Merging.

PR #392, opened 2015-07-28
Split up calicoctl_test into separate files based on class

L-MA:


robbrockbank:
Files have changed too much - closing this PR.

We can re-implement as necessary.

PR #396, opened 2015-07-29
Policy-only mode for calico/node

caseydavenport:


tomdee:
Looks OK, but I would ditch setting a value for POLICY_ONLY_CALICO

We shouldn't care about the value - if it's set, then that should be enough to trigger the behaviour.

caseydavenport:
@tomdee - I've ditched the value for POLICY_ONLY_CALICO.

tomdee:
OK, LGTM. No plans to add a calicoctl option for this?

caseydavenport:
Wasn't planning on it in the short term, but I suppose it would be easy enough to add.  Might be worth throwing in..

lxpollitt:
No, I vote this is an internal feature for the moment.  We've some more work to do before deciding how best to package it beyond the initial POC/demo stuff we are using it for now.

PR #401, opened 2015-07-31
Calico Policy through Kubernetes Annotations

alexhersh:
WIP

### DIFF
--- a/calico_containers/integrations/kubernetes/calico_kubernetes.py
+++ b/calico_containers/integrations/kubernetes/calico_kubernetes.py
@@ -4,6 +4,7 @@ import os
 import sys
 from subprocess import check_output, CalledProcessError, check_call
 import requests
+from urllib import quote
 import sh
 
 # Append to existing env, to avoid losing PATH etc.
@@ -160,12 +161,12 @@ class NetworkPlugin(object):
 
         Currently assumes one pod with each name.
         """
-        print('Configuring Pod Profile')
         profile_name = self.pod_name
+        print('Configuring Pod Profile %s' % profile_name)
         self.calicoctl('profile', 'add', profile_name)
         pod = self._get_pod_config()
 
-        self._apply_rules(profile_name)
+        self._apply_rules(profile_name, pod)
 
         self._apply_tags(profile_name, pod)
 
@@ -247,7 +248,7 @@ class NetworkPlugin(object):
         auth_data = json.loads(json_string)
         return auth_data['BearerToken']
 
-    def _generate_rules(self):
+    def _generate_rules(self, pod):
         """
         Generate the Profile rules that have been specified on the Pod's ports.
 
@@ -257,17 +258,42 @@ class NetworkPlugin(object):
 
         :return list() rules: the rules to be added to the Profile.
         """
+
+        namespace, ns_tag = self._get_namespace_and_tag(pod)
+
         inbound_rules = [
             {
-                'action': 'allow',
-            },
+                "action": "allow",
+                "src_tag": ns_tag
+            }
         ]
 
         outbound_rules = [
             {
-                'action': 'allow',
-            },
+                "action": "allow"
+            }
         ]
+
+        print('Getting Policy Rules from Annotation of pod %s' % pod)
+
+        annotations = self._get_metadata(pod, 'annotations')
+
+        if 'allowFrom' in annotations.keys():
+            inbound_rules = []
+            rules = json.loads(annotations['allowFrom'])
+            for rule in rules:
+                rule['action'] = 'allow'
+                rule = self._translate_rule(rule, namespace)
+                inbound_rules.append(rule)
+
+        if 'allowTo' in annotations.keys():
+            outbound_rules = []
+            rules = json.loads(annotations['allowTo'])
+            for rule in rules:
+                rule['action'] = 'allow'
+                rule = self._translate_rule(rule, namespace)
+                outbound_rules.append(rule)
+
         return inbound_rules, outbound_rules
 
     def _generate_profile_json(self, profile_name, rules):
@@ -289,10 +315,10 @@ class NetworkPlugin(object):
             'outbound_rules': outbound,
         }
         profile_json = json.dumps(profile, indent=2)
-        print('Final profile "%s": %s' % (profile_name, profile_json))
+        print('Final profile "%s":\n%s' % (profile_name, profile_json))
         return profile_json
 
-    def _apply_rules(self, profile_name):
+    def _apply_rules(self, profile_name, pod):
         """
         Generate a new profile with the default 'allow all' rules.
 
@@ -300,7 +326,7 @@ class NetworkPlugin(object):
         :type profile_name: string
         :return:
         """
-        rules = self._generate_rules()
+        rules = self._generate_rules(pod)
         profile_json = self._generate_profile_json(profile_name, rules)
 
         # Pipe the Profile JSON into the calicoctl command to update the rule.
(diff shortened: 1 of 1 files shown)


calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -260,14 +261,26 @@ def _generate_rules(self):
...
+            }
         ]
+
+        print('Getting Policy Rules from Annotation of pod %s' % pod)
+
+        annotations = self._get_annotations(pod)
+        namespace = self._get_namespace(pod)
+
+        if 'allowFrom' in annotations.keys():
+            inbound_rules = eval(annotations['allowFrom'])
```
caseydavenport: I wouldn't use the eval() builtin to load json - it could do other things that we don't want it to (e.g. someone could embed python in their annotation and we'd start executing it.)

Take a look at https://docs.python.org/2/library/json.html

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -326,9 +339,18 @@ def _apply_tags(self, profile_name, pod):
             print('No labels found in pod %s' % pod)
             return
 
+        # Grab namespace and create a tag if it exists.
+        namespace = self._get_namespace(pod)
+        if namespace:
+            tag = self._label_to_tag('namespace', namespace, None)
```
caseydavenport: I'd define 'namespace' as NAMESPACE_PREFIX up at the top.  Will make this code easier to understand.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -336,6 +358,97 @@ def _apply_tags(self, profile_name, pod):
                 print('Could not create tag %s.\n%s' % (tag, e))
         print('Finished applying tags.')
 
+    def _get_annotations(self, pod):
```
caseydavenport: _get_annotations and _get_namespace look highly duplicated.

We could probably do with a generic "_get_pod_metadata" method which allows you to specify which bit of metadata you'd like.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -336,6 +358,97 @@ def _apply_tags(self, profile_name, pod):
...
+        translation_dictionary = {
+            'srcPorts': 'src_ports',
+            'dstPorts': 'dst_ports',
+            'srcNet': 'src_net',
+            'dstNet': 'dst_net',
+            'icmpType': 'icmp_type'
+        }
+
+        # Kubernetes and Calico use different syntax, use a dictionary to translate
+        for kube_key, calico_key in translation_dictionary.iteritems():
```
caseydavenport: This is not the safest way to do this conversion, as it doesn't handle keys which are not in the translation dict.

I'd create a new dictionary, iterate through the given kubernetes rule, and for each key in the translation dictionary add an entry in the new calico dictionary.  This allows us to catch rules that we don't support, and provide a suitable warning / error message.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -336,6 +358,97 @@ def _apply_tags(self, profile_name, pod):
...
+    def _translate_rule(self, rule, namespace):
+        """
+        Given a JSON dict rule from Kubernetes Annotations, 
+        output a JSON dict that is calicoctl profile compliant
+        :param rule: dict of keys relating to policy rules
+        :param namespace: string indicating namespace for pod, None if not available
+        :return: translated dict of keys relating to policy rules for Calico profile
+        :rtype: JSON dict
+        """
+        calico_rule = rule
```
caseydavenport: This doesn't actually make a copy of rule - they two variables will use the same underlying dictionary.

calico_containers/integrations/kubernetes/calico_kubernetes.py
alexhersh: I think thats what I meant to do... committing now.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -4,6 +4,8 @@
 import sys
 from subprocess import check_output, CalledProcessError, check_call
 import requests
+from urllib import quote
+from json import loads as load_json
```
caseydavenport: I wouldn't mask the name of a module unless you need to - it makes the code harder to follow.  I'd just import json.

json.loads( ) is what people will expect when reading this code.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -260,14 +262,26 @@ def _generate_rules(self):
...
+            }
         ]
+
+        print('Getting Policy Rules from Annotation of pod %s' % pod)
+
+        annotations = self._get_metadata(pod, 'annotations')
+        namespace = self._get_metadata(pod, 'namespace')
+
+        if 'allowFrom' in annotations.keys():
+            inbound_rules = load_json(annotations['allowFrom'])
```
caseydavenport: I wouldn't load directly to inbound_rules.  There might be problems with the loaded rules.  We should load the rules into their own list, translate each one, and if it succeeds translation then add it to the list.

The other thing to think about is what do we do if a rule is not valid? Do we carry on and try to do our best?  Its probably not a security concern (yet), since by default traffic will be dropped, and each rule is a whitelist entry.  But we should be logging appropriate, human readable errors when an allowFrom rule is invalid.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -319,23 +333,118 @@ def _apply_tags(self, profile_name, pod):
...
+                if 'src_tag' not in calico_rule.keys():
+                    tag = self._label_to_tag(k, v, namespace)
+                    calico_rule['src_tag'] = tag
+                else:
+                    print "More than one label specified for policy rule. Multi-tag selection not supported in Calico"
+                    sys.exit(1)
+
+            kube_rule.pop('labels')
+
+        if kube_rule :
```
caseydavenport: Stick a comment here explaining that since we pop the entries from kube_rule, any remaining entries must be un-parseable.

calico_containers/integrations/kubernetes/calico_kubernetes.py
```diff
@@ -319,23 +333,113 @@ def _apply_tags(self, profile_name, pod):
...
-            print('No labels found in pod %s' % pod)
-            return
-
-        for k, v in labels.iteritems():
-            tag = '%s_%s' % (k, v)
-            tag = tag.replace('/', '_')
-            print('Adding tag ' + tag)
+
+        # Grab namespace and create a tag if it exists.
+        NAMESPACE_PREFIX = self._get_metadata(pod, 'namespace')
```
caseydavenport: Not quite what I meant when I said you should use NAMESPACE_PREFIX.

I was thinking of the 'namespace' string as a constant which indicates a namespace.

e.g.
```
tag = self._label_to_tag(NAMESPACE_PREFIX, namespace, None)
```

Looking at this closer, I'm not sure that self._label_to_tag is the correct method to be using to generate the namespace tag.  It works, since it is similar function, but the abstraction seems wrong, since we're not actually creating a tag from a label, but instead from a namespace and a defined prefix.


Maybe something like this?
```
....
ns_tag = _escape_string("%s=%s" % (NAMESPACE_PREFIX, namespace))
....

def _label_to_tag(self, label_key, label_value, namespace):
    ....
    tag = "%s/%s=%s" % (namespace, label_key, label_value)
    return self._escape_string(tag)

def _escape_string(self, string):
    """ Escapes the given string for use in Calcio profiles / tags """
   ....

```

calico_containers/integrations/kubernetes/calico_kubernetes.py
alexhersh: calico has some pretty strict rule requirements. It will complain if you try to add an invalid rule.

alexhersh:
Moved to calico-kubernetes

PR #405, opened 2015-08-03
Container ut

robbrockbank:
Originally Walkers branch - now rebased off master.

### DIFF
--- a/calico_containers/tests/unit/calicoctl_test.py
+++ b/calico_containers/tests/unit/calicoctl_test.py
@@ -15,9 +15,13 @@
 import unittest
 from requests import Response
 from StringIO import StringIO
+
+from docker.errors import APIError
 from mock import patch, Mock, call
 from nose_parameterized import parameterized
-from docker.errors import APIError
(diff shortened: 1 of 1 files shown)


calico_containers/tests/unit/calicoctl_test.py
```diff
@@ -25,11 +29,12 @@
 from calico_ctl.node import validate_arguments as node_validate_arguments
 from calico_ctl.pool import validate_arguments as pool_validate_arguments
 from calico_ctl.profile import validate_arguments as profile_validate_arguments
```
TrimBiggs: All of these imports should probably just be called like `<module>.validate_arguments` now, rather than importing each  one individually in addition to the module and naming it pretty much the same thing.

robbrockbank: Not planning to change as we should address this as part of the split-out into separate files.  Once split out into individual files we can just import validate_arguments without conflict.

robbrockbank:
Combining into single PR

PR #406, opened 2015-08-03
Add calicoctl diags ut

robbrockbank:


robbrockbank:
Combining into single PR.

PR #408, opened 2015-08-03
Add calicoctl checksystem UTs

robbrockbank:


robbrockbank:
Combining into single PR.

PR #413, opened 2015-08-03
Updated doc hostname references to ip address equivalents

TrimBiggs:
Replace instances of calico-01 and calico-02 when referencing hostnames with 172.17.8.101 and 172.17.8.102, respectively.

### DIFF
--- a/docs/getting-started/GCE.md
+++ b/docs/getting-started/GCE.md
@@ -63,7 +63,9 @@ gcloud compute instances create \
   --metadata-from-file user-data=<PATH_TO_CLOUD_CONFIG>/user-data-first
 ```
 
-For the second server run:
+Open your `user-data-others` file and replace the instances of `172.17.8.101` with the private IP address of the `calico-01` server you just created.  You can find this in the output of the previous command.
+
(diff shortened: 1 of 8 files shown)


tomdee:
LGTM - @robbrockbank can you take a look too?

docs/getting-started/GCE.md
```diff
@@ -63,7 +63,9 @@ gcloud compute instances create \
   --metadata-from-file user-data=<PATH_TO_CLOUD_CONFIG>/user-data-first
 ```
 
-For the second server run:
+Open your `user-data-others` file and replace the instances of `172.17.8.101` with this private IP address of the `calico-01` server you just created.
```
robbrockbank: with this -> with the

Also, we don't tell the user how to determine the private IP address do we?  We explicitly tell them for AWS and DigitalOcean.

TrimBiggs: Correct, but I'm not sure how in GCE yet.  I can update this when I test it later.

robbrockbank:
LGTM.

PR #417, opened 2015-08-04
Fix a few typos and formatting bugs in the docs

alexwlchan:


docs/getting-started/DigitalOcean.md
```diff
@@ -120,7 +120,7 @@ sudo iptables -A PREROUTING -t nat -i eth0 -p tcp --dport 80 -j DNAT  --to 192.1
 ```
 
 You should now be able to access the containers using the public IP address of your DigitalOcean host on port 80 by 
-visiting http://<host public ip>:80 or running:
+visiting http://`<host public ip>`:80 or running:
```
robbrockbank: I find just quoting the <host..> looks a bit odd when viewed, and we don't hightlight the <host...> in the curl below.

I'd either leave unchanged, or quote the entire http://blah

alexwlchan: I was adopting the style used in one of the other Getting Started docs. As it currently stands, those angle brackets are swallowed as an HTML tag. Feel free to change, but there's at least one other doc (not in this commit) where that style is used.

robbrockbank: Yeah I'll change to quote the entire http string - from a glance at other docs we don't just quote the angle-bracket parameter.  If we do that elsewhere I'll fix up when I see it.

robbrockbank:
Thanks Alex,  I made a couple of minor tweaks and have now merged.

PR #421, opened 2015-08-05
Fix container ip add command parameters

### LINKED ISSUE #420: function container_ip_add() takes exactly 3 arguments but 4 given
The definition of `container_ip_add()` [here](https://github.com/projectcalico/calico-docker/blob/master/calico_containers/calico_ctl/container.py#L252), only have three arguments
But when you calling it, you passed it 4 args, [here](https://github.com/projectcalico/calico-docker/blob/master/calico_containers/calico_ctl/container.py#L74)

TrimBiggs:
Fixes #420

### DIFF
--- a/calico_containers/calico_ctl/container.py
+++ b/calico_containers/calico_ctl/container.py
@@ -65,7 +65,6 @@ def container(arguments):
     """
     validate_arguments(arguments)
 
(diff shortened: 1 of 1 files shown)


robbrockbank:
LGTM.

PR #425, opened 2015-08-05
Fix broken UTs after container ip add bug fix

TrimBiggs:


robbrockbank:
Fine.  Check in once circle/ci is working.

PR #428, opened 2015-08-07
Add system tests for calicoctl pool command

TrimBiggs:


calico_containers/tests/st/utils/defaults.py
```diff
@@ -0,0 +1,23 @@
...
+
+# IPv4 Pool Adresses
+DEFAULT_IPV4_POOL = "192.168.0.0/16"
+IPV4_POOL_ADDR_1 =  "192.168.1.1"
+IPV4_POOL_ADDR_2 =  "192.168.1.2"
+
+# IPv6 Pool Addresses
+DEFAULT_IPV6_POOL = "fd80:24e2:f998:72d6::/64"
+IPV6_POOL_ADDR_1 = "fd80:24e2:f998:72d6::11"
+IPV6_POOL_ADDR_2 = "fd80:24e2:f998:72d6::22"
```
robbrockbank: Where are the ADDR_1 and ADDR_2 values used?

Should the default POOLs be taken from libcalico somewhere?

TrimBiggs: Whoops, ADDR_1 and ADDR_2 aren't used here, I copied them from another test I wrote, meant to get rid of them.

There isn't a `defaults` or `constants` file for the calico pools.  These pools are general also, not just libnetwork.  

I decided to put these default pool addresses in an ST file in case we decide to use more ST constants that are unrelated to the rest of the repo (as the ADDR_1 and ADDR_2 values would be).  I would also be on board with saving the IP as a global wherever the default pool is defined in calico-docker, though I think have a defaults file for the STs is also useful.

robbrockbank: i said libcalico not libnetwork.  are these default values just hardcoded somewhere then?

TrimBiggs:
Right libcalico, apologies.  They are just hardcoded, yes, in `calico_ctl/node.py`.  I've updated my PR to have the values defined as values in `node.py`, at the very least, and removed the defaults.py since it is unnecessary right now.

calico_containers/calico_ctl/node.py
```diff
@@ -62,8 +62,10 @@
 from utils import validate_ip
 from checksystem import check_system
 
-DEFAULT_IPV4_POOL = IPPool("192.168.0.0/16")
-DEFAULT_IPV6_POOL = IPPool("fd80:24e2:f998:72d6::/64")
+DEFAULT_IPV4_POOL_IP = "192.168.0.0/16"
```
robbrockbank: strictly, this is not an IP, it's a CIDR.

calico_containers/tests/st/test_pool.py
```diff
@@ -28,23 +31,60 @@
...
+        Assert container receives IP from new IPv4 pool.
         """
-        pass
-        # TODO Remove the default, create a new pool. Create a container and check
-        #  it gets the IP from the pool.
-        # [Needs to be libnetwork based since that uses IPAM]
+        with DockerHost('host', dind=False) as host:
+            # Remove default pool and add new pool
+            ipv4_pool = "10.0.1.0/24"
+            host.calicoctl("pool remove %s" % DEFAULT_IPV4_POOL_IP)
```
robbrockbank: could you not have just done str(DEFAULT_IPV4_POOL) ?

TrimBiggs: Yup, that works.

calico_containers/calico_ctl/node.py
TrimBiggs: Right.  I guess it's not so bad to use str() of the POOL globals, but I think it would be simpler to have the string constant as well.

TrimBiggs: I guess it's not used in "a lot" of places, but having a global would still be useful.

PR #433, opened 2015-08-11
Remove docopt completion since it no longer works

tomdee:


TrimBiggs:
LGTM, I think the semaphore issue is a transient failure. 

I kicked off a rebuild, but it should be fine.

PR #435, opened 2015-08-11
Create KubernetesPolicy.md

alexhersh:


docs/kubernetes/KubernetesIntegration.md
```diff
@@ -32,3 +32,6 @@ This guide will describe the configuration required to use the Calico network pl
 
 #### Configuring Kubelet Services
    On each of your nodes, you will need to configure the Kubelet to use the Calico Networking Plugin. This can be done by including the `--network_plugin=calico` option when starting the Kubelet. If you are using systemd to manage your services, you can add this line to the Kubelet config file (`/etc/systemd/` by default) and restart your Kubelets to begin using Calico.
+
+### Configuring Policy
+   See our doc on [Programming Kubernetes Policy](KubernetesPolicy.md) to start networking your nodes!
```
caseydavenport: "networking your nodes" --> "enforcing security policy on Kubernetes pods"

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
+# Programming Calico Policy in Kubernetes
+The [Calico-Kubernetes plugin](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md) will allow you to program Networking policy directly from your Kubernetes pod config.
```
caseydavenport: "Networking" should be lowercase.

This should probably read more like "Calico Kubernetes plugin allows you to specify networking policy on the Kubernetes API using pod annotations."

Side note: Don't stick the hyphen in Calico Kubernetes :)

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
+# Programming Calico Policy in Kubernetes
+The [Calico-Kubernetes plugin](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md) will allow you to program Networking policy directly from your Kubernetes pod config. 
+
+
+### The Basics
```
caseydavenport: This should be "Declaring Policy", and explain how a user can declare policy.

"The Calico Networking plugin will look for a `policy` key" --- I think this is too implementation oriented.  The goal here is to allow end-users to easily declare policy, so we shouldn't need to specify anything implementation specific. (For example, why should they care if annotations are passed over the networking plugin API, or if the plugin queries for the annotations?)

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
+# Programming Calico Policy in Kubernetes
+The [Calico-Kubernetes plugin](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md) will allow you to program Networking policy directly from your Kubernetes pod config. 
+
+
```
caseydavenport: I think we want a pre-requisites section that links to "how to install the plugin" and also touches on the BenTheElder iptables-based kube-proxy requirement.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
...
+### The Basics
+The Calico Networking plugin will look for a `policy` key in the annotations section of your pod's metadata. This `policy` key should map to a single string that outlines networking behavior semantically.
+
+Programming policy is as simple as saying what you want your pod to do!
```
caseydavenport: As fun as this line is,  I think we should kill it. :(

Yep.  I'm a fun killer...

docs/kubernetes/KubernetesPolicy.md
caseydavenport: The syntax bit can be moved up here as well.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
...
+...
+      policy: "allow from label name=backend; allow tcp to ports 4001,443"
+...
+```
+In it's current state, Calico supports whitelist oriented, inbound rules. This means that anything unspecified by the policy spec will be rejected at the destination. 
+
+Calico can specify any combination of label (multiple labels per rule are unsupported), protocol, source/dest ports, and source/dest net. The full syntax for creating compound rules is shown below [below](#Syntax).
+
+### Defaults
+With no specified policy, Calico will only allow traffic from within a pod's own namespace. This default rule will be overidden if any policy is programmed. If you wish to include rules that police the namespace, you can do so with the keyword `tag namespace_<NAMESPACE>`, as in `allow tcp from tag namespace_default`
```
caseydavenport: Eh, I don't think we want to be exposing the "allow from namespace" semantics to the world.  That is more of a fluke in terms of how we implemented the first version, rather than an approved and supported way of declaring policy.

We still need to design and propose the correct way of exposing services across namespaces.  For now, it is unsupported.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,62 @@
...
+...
+```
+
+To specify multiple rules, separate them by semicolons.
+```
+...
+      policy: "allow from label name=backend; allow tcp to ports 4001,443"
+...
+```
+In it's current state, Calico supports whitelist oriented, inbound rules. This means that anything unspecified by the policy spec will be rejected at the destination.
```
caseydavenport: Calico supports whitelist and blacklist oriented inbound and outbound rules.

The Calico Kubernetes Plugin supports whitelist oriented inbound rules.

"This means that any traffic not specified in a pod's policy is unauthorized, and unauthorized traffic will be dropped at the receiving Kubernetes node."

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
+# Programming Calico Policy in Kubernetes
+The [Calico Kubernetes plugin](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md) allows you to specify networking policy in the Kubernetes API using pod annotations. 
+
+## Prerequisites
+* A Kubernetes Deployment
+    - To implement service policy using recent builds of Kubernetes, make sure you have configured your kube-proxy using the `--legacy-userspace-proxy=false` option
+* v0.1.0+ of the Calico Kubernetes Plugin
```
caseydavenport: I'd link to the GitHub releases page here - its a bit confusing, since there are some hosted version of calico_kubernetes on calico-docker with version 0.4.* and 0.5.*, which are actually older than the 0.1.0 release on calico-kubernetes.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
...
+You can specify multiple rules by separating them with semicolons.
+```
+...
+      policy: "allow from label name=backend; allow tcp to ports 4001,443"
+...
+```
+In it's current state, the Calico Kubernetes Plugin supports whitelist oriented, inbound rules. This means that any traffic not specified in a pod's policy is unauthorized, and unauthorized traffic will be dropped at the receiving Kubernetes node.
+
+### Defaults
+With no specified policy, Calico will only allow traffic from within a pod's own namespace. This default rule will be overidden if any policy is programmed. The only exception to this are services within the `kube-system` namespace. These services are universally accessed by all namespaces and will allow all traffic.
```
caseydavenport: Careful with the use of the work "services", since these mean a specific thing in Kubernetes.  Any pods within the kube-system namespace are freely accessible.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
...
+...
+```
+
+You can specify multiple rules by separating them with semicolons.
+```
+...
+      policy: "allow from label name=backend; allow tcp to ports 4001,443"
+...
+```
+In it's current state, the Calico Kubernetes Plugin supports whitelist oriented, inbound rules. This means that any traffic not specified in a pod's policy is unauthorized, and unauthorized traffic will be dropped at the receiving Kubernetes node.
```
alexwlchan: “it’s” shouldn’t have an apostrophe.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
+# Programming Calico Policy in Kubernetes
+The [Calico Kubernetes plugin](https://github.com/projectcalico/calico-docker/blob/master/docs/kubernetes/KubernetesIntegration.md) allows you to specify networking policy in the Kubernetes API using pod annotations. 
+
+## Prerequisites
+* A Kubernetes Deployment
+    - To implement service policy using recent builds of Kubernetes, make sure you have configured your kube-proxy using the `--legacy-userspace-proxy=false` option
```
caseydavenport: Stick punctuation at the end of these guys.

I'd also mention that the v1.0 release of Kubernetes doesn't include the correct kube-proxy, and that you'll need to build Kubernetes master commit #abcdef or later to get the required kube-proxy behavior.

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
...
+* [v0.1.0+](https://github.com/projectcalico/calico-kubernetes/releases) of the Calico Kubernetes Plugin
+    - For more information on how to integrate Calico into your Kubernetes Deployment, view our [integration doc](KubernetesIntegration.md)
+
+### Declaring Policy
+To enforce a policy rule to your pod, add a `policy` key to the annotations section of your pod's metadata. This `policy` key should map to a single string that outlines networking behavior semantically.
```
caseydavenport: "outlines networking behavior semantically" is a bit odd sounding.

"The policy should map to a single string which defines network behavior.  The syntax for this can be seen below"

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
...
+### Tags, Labels, and Namespaces
+For each policy profile, Calico will generate a tag for its namespace, pod name, and for each label pair.
+
+For example, the metadata
+```
+metadata:
+  name: pod1
+  namespace: app1
+  labels:
+    name: backend
```
caseydavenport: This label is a bit atypcial - I'd use "role: backend", rather than "name"

docs/kubernetes/KubernetesPolicy.md
```diff
@@ -0,0 +1,63 @@
...
+With no specified policy, Calico will only allow traffic from within a pod's own namespace. This default rule will be overidden if any policy is programmed. The only exception to this are resources within the `kube-system` namespace. These are universally accessed by all namespaces and will accept all traffic.
+
+### Tags, Labels, and Namespaces
+For each policy profile, Calico will generate a tag for its namespace, pod name, and for each label pair.
+
+For example, the metadata
+```
+metadata:
+  name: pod1
+  namespace: app1
```
caseydavenport: Namespaces in Kubernetes wouldn't usually be used per-application, but instead per-tenant or per-deployment.  I might call this "namespace: production" or the like.

caseydavenport:
LGTM

PR #449, opened 2015-08-14
Remove libnetwork.

TrimBiggs:
libnetwork now has its own repo.
Clean up STs to use default networking.


calico_containers/tests/st/bgp/test_route_reflector_cluster.py
```diff
@@ -32,28 +32,45 @@ def test_route_reflector_cluster(self):
...
+                                                     DEFAULT_IPV4_ADDR_3))
+
+            # Get the endpoint IDs for the containers
+            ep1 = host1.calicoctl("container %s endpoint-id show" % workload_host1)
+            ep2 = host2.calicoctl("container %s endpoint-id show" % workload_host2)
+            ep3 = host3.calicoctl("container %s endpoint-id show" % workload_host3)
+
+            # Now add the profiles
+            host1.calicoctl("endpoint %s profile set TEST_GROUP" % ep1)
+            host2.calicoctl("endpoint %s profile append TEST_GROUP" % ep2)
```
robbrockbank: append will work here, but it should probably just be `set`

The profile is set or appended to the endpoint, so none of the endpoints will initially have a profile.

robbrockbank: same for the other places.

robbrockbank:
once the tests pass - lgtm

PR #452, opened 2015-08-17
Added --libnetwork flag to calicoctl node

TrimBiggs:


calico_containers/calico_ctl/node.py
```diff
@@ -26,7 +26,7 @@
 Options:
   --force                   Stop the node process even if it has active endpoints.
   --node-image=<DOCKER_IMAGE_NAME>    Docker image to use for Calico's per-node
-                                      container [default: calico/node:latest]
+                                      container. Default is calico/node:latest.
```
robbrockbank: I would move the comment below up here so that we talk about the default being X, or with libnetwork Y.


Then the comment for --libnetwork can just be run calico node with the libnetwork plugin.

robbrockbank:
I assume there are some doc updates somewhere?

docs/getting-started/libnetwork/EnvironmentSetup.md
```diff
@@ -106,12 +106,12 @@ directory.
 
 ### Preload the Calico docker image (optional)
 
-You can optionally preload the Calico Docker image to avoid the delay when you 
-run `calicoctl node` the first time.  Select the same version of the Calico 
-Docker image as you selected above.  For example, to pull the latest released 
-version
+The libnetwork release of calico uses its own calico/node-libnetwork Docker
```
robbrockbank: Strictly Calico (with a cap C) is really the networking component (i.e. Felix), so there isn't a libnetwork release of Calico.  There is of calico/node ... 

Anyway, I don't think you need to compare the two different images here - it just adds confusion.

docs/getting-started/libnetwork/EnvironmentSetup.md
```diff
@@ -106,12 +106,11 @@ directory.
 
 ### Preload the Calico docker image (optional)
 
-You can optionally preload the Calico Docker image to avoid the delay when you 
-run `calicoctl node` the first time.  Select the same version of the Calico 
-Docker image as you selected above.  For example, to pull the latest released 
-version
+You can optionally preload this image to avoid the delay when you  run
```
robbrockbank: two spaces between you and run

robbrockbank:
LGTM.

 I think we need a backlog item to look into whether we support Docker 1.8 and if not to update our docs accordingly, or if we do, update our docs to handle the slight differences in docker ps output etc.

PR #453, opened 2015-08-17
Allow calicoctl container add to take namespace path

tomdee:
Just putting this up here to get a test run.

### DIFF
--- a/calico_containers/calico_ctl/container.py
+++ b/calico_containers/calico_ctl/container.py
@@ -14,6 +14,7 @@ Options:
   --interface=<INTERFACE>  The name to give to the interface in the container
                            [default: eth1]
 """
+import os
 import sys
 import uuid
 
@@ -28,7 +29,8 @@ from pycalico.datastore_datatypes import Endpoint
 
 from connectors import client
 from connectors import docker_client
-from utils import hostname, ORCHESTRATOR_ID
+from utils import hostname, DOCKER_ORCHESTRATOR_ID, NAMESPACE_ORCHESTRATOR_ID, \
+    escape_etcd
 from utils import enforce_root
 from utils import print_paragraph
 from utils import validate_ip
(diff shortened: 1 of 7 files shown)


calico_containers/calico_ctl/container.py
```diff
@@ -161,18 +185,7 @@ def container_add(container_name, ip, interface):
...
-    # Check the container is actually running.
-    if not info["State"]["Running"]:
-        print "%s is not currently running." % container_name
-        sys.exit(1)
-
-    # We can't set up Calico if the container shares the host namespace.
-    if info["HostConfig"]["NetworkMode"] == "host":
-        print "Can't add %s to Calico because it is " \
-              "running NetworkMode = host." % container_name
+              id
```
TrimBiggs: This could just go on the same line.

calico_containers/calico_ctl/container.py
```diff
@@ -134,23 +136,45 @@ def container(arguments):
         sys.exit(1)
 
 
-def container_add(container_name, ip, interface):
+def container_add(id, ip, interface):
```
TrimBiggs: I'd lean towards not naming this id, since id is a keyword in python.

calico_containers/calico_ctl/container.py
```diff
@@ -228,7 +240,7 @@ def container_add(container_name, ip, interface):
     return ep
 
 
-def container_remove(container_name):
+def container_remove(id):
```
TrimBiggs: Here also

calico_containers/calico_ctl/container.py
TrimBiggs: Same for other places

image/install.sh
```diff
@@ -13,11 +13,6 @@ apt-get install -qy \
         libffi-dev \
         libssl-dev
 
-# Install Felix and python-etcd from the Metaswitch github repos.
-pip install git+https://github.com/projectcalico/python-etcd.git
-pip install git+https://github.com/projectcalico/calico.git@0.28
```
TrimBiggs: you'll get a conflict here when rebasing

TrimBiggs:
Yup LGTM

calico_containers/calico_ctl/container.py
```diff
@@ -90,26 +92,26 @@ def container(arguments):
                     container_remove(arguments.get("<CONTAINER>"))
         if arguments.get("endpoint"):
             endpoint.endpoint_show(hostname,
-                                   ORCHESTRATOR_ID,
+                                   None,
```
robbrockbank: IIRC, the endpoint lookup does a recursive lookup based on the maximum number of complete path segments that can be constructed and then does a linear search to match the remaining segments.

Upshot, if you remove the orchestrator like you are doing here, we'll end up enumerating all of the workloads even though you're actually passing it in.

tomdee: OK, I'll fix up tomorrow.

PR #454, opened 2015-08-17
Add general configurability of BIRD and Felix

robbrockbank:
Need to write UTs - but IMO I wouldn't let this prevent us merging.

### DIFF
--- a/calico_containers/calico_ctl/config.py
+++ b/calico_containers/calico_ctl/config.py
@@ -53,10 +53,10 @@ from utils import hostname
 # Dictionaries providing look up between the configuration name, and a tuple
 # of (internal name, value regex string)
 FELIX_CONFIG_DATA = {
-    "logging": ("LogSeverityScreen", "none|debug|info|warning|error|critical")
+    "loglevel": ("LogSeverityScreen", "none|debug|info|warning|error|critical")
 }
 BGP_CONFIG_DATA = {
-    "logging": ("LogSeverityScreen", "none|debug|info")
+    "loglevel": ("loglevel", "none|debug|info")
 }
 
 def validate_arguments(arguments):
--- a/calico_containers/calico_ctl/config.py
+++ b/calico_containers/calico_ctl/config.py
@@ -1,0 +1,192 @@
+# Copyright 2015 Metaswitch Networks
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Usage:
+  calicoctl config felix <NAME> [<VALUE>|--remove] [--force]
+  calicoctl config bgp <NAME> [<VALUE>|--remove] [--force]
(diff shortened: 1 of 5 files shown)


calico_containers/calico_ctl/config.py
```diff
@@ -0,0 +1,191 @@
...
+from utils import print_paragraph
+from utils import hostname
+
+# Dictionaries providing look up between the configuration name, and a tuple
+# of (internal name, value regex string)
+FELIX_CONFIG_DATA = {
+    "logging": ("LogSeverityScreen", "none|debug|info|warning|error|critical")
+}
+BGP_CONFIG_DATA = {
+    "logging": ("LogSeverityScreen", "none|debug|info|warning|error|critical")
```
TrimBiggs: Should this just be `none|debug|info`?

robbrockbank: yup

calico_containers/calico_ctl/config.py
```diff
@@ -0,0 +1,192 @@
...
+            print_paragraph("The --force option is set.")
+            return
+        else:
+            print_paragraph("Use the --force option to override.")
+            sys.exit(1)
+
+    value = arguments.get("<VALUE>")
+    if value:
+        _, valid_values = config_data[name]
+        valid_values_re = re.compile(valid_values)
```
tomdee: Do you really need to use REs here? Why not just have a list of valid values and check if the user string is in that list?

calico_containers/calico_ctl/config.py
```diff
@@ -0,0 +1,192 @@
...
+from utils import print_paragraph
+from utils import hostname
+
+# Dictionaries providing look up between the configuration name, and a tuple
+# of (internal name, value regex string)
+FELIX_CONFIG_DATA = {
+    "logging": ("LogSeverityScreen", "none|debug|info|warning|error|critical")
+}
+BGP_CONFIG_DATA = {
+    "logging": ("LogSeverityScreen", "none|debug|info")
```
tomdee: Why are we using LogSeverityScreen for bird?

node_filesystem/templates/bird.cfg.template
```diff
@@ -3,6 +3,14 @@ include "bird_ipam.cfg";
 
 router id {{getenv "IP"}};
 
+{{define "LOGGING"}}
+{{$node_logging_key := printf "/host/%s/LogSeverityScreen" (getenv "HOSTNAME")}}{{if exists $node_logging_key}}{{$logging := getv $node_logging_key}}
+{{if eq $logging "debug"}}  debug all;{{else if ne $logging "none"}}  debug { states };{{end}}
```
tomdee: What is the debug {states} thing?

tomdee:
I've finished looking at this for now.

calico_containers/calico_ctl/config.py
robbrockbank: With the current options a simple list of values would suffice.  I did regexes because I thought it would be a little more extensible... say for example you need a value to be an IP address.  Perhaps should encapsulate so you can define your own validator, but not worth it just at the moment.

calico_containers/calico_ctl/config.py
robbrockbank: I'll change to loglevel.  I'll also change the nice name in all cases to be loglevel.

node_filesystem/templates/bird.cfg.template
robbrockbank: Just what we currently use for our "info" level logging.  Could perhaps use some more refining based on conversations with Shaun from his scale testing.

robbrockbank:
Just realised there is a doc change required for the logging doc - I'll do this later, but let's start the release first.

PR #462, opened 2015-08-21
Don't include cbr0 in IP autodetect

caseydavenport:
Fixes issue found [here](https://github.com/projectcalico/libcalico/issues/15)

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -202,7 +202,7 @@ def node_start(node_image, log_dir, ip, ip6, as_num, detach, kubernetes,
 
     # Get IP address of host, if none was specified
     if not ip:
-        ips = get_host_ips(exclude=["$docker0.*"])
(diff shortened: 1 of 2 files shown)


calico_containers/calico_ctl/node.py
```diff
@@ -202,7 +202,7 @@ def node_start(node_image, log_dir, ip, ip6, as_num, detach, kubernetes,
 
     # Get IP address of host, if none was specified
     if not ip:
-        ips = get_host_ips(exclude=["$docker0.*"])
+        ips = get_host_ips(exclude=["$docker0.*", "cbr0"])
```
TrimBiggs: This will exclude all interfaces that contain the string "cbr0".  To do just cbr0, use "^cbr0$".

I think the "$docker.*" should actually be "^docker.*", can you fix that too, while you're here?

TrimBiggs:
LGTM

PR #463, opened 2015-08-21
Add IPv6 demo to default networking

TrimBiggs:


robbrockbank:
Looks good.

Do we need to link to the IPv6 document from anywhere?  Otherwise, will anyone stumble across it?

TrimBiggs:
Yup, good point.  I'll update the Demonstration to point to it like with the libnetwork demo.

robbrockbank:
lgtm

PR #468, opened 2015-08-25
Forces integer notation in templates for AS numbers in the 4-byte range

cheuschober:
Prior to this, 4-byte integers would appear in scientific notation

### DIFF
--- a/node_filesystem/templates/bird.cfg.template
+++ b/node_filesystem/templates/bird.cfg.template
@@ -77,7 +77,7 @@ template bgp bgp_template {
 {{$nums := split $data.ip "."}}{{$id := join $nums "_"}}
 # For peer {{.Key}}
 protocol bgp Global_{{$id}} from bgp_template {
-  neighbor {{$data.ip}} as {{$data.as_num}};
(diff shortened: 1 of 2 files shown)


robbrockbank:
Hi @cheuschober,

Thanks for this fix.  We've actually gone ahead and made a slightly different change - rather than forcing integer notation in the template, we've updated libcalico (the common library used for manipulating calico data) to store the peer AS numbers as strings - which is what we were doing for the AS number of the node.

The reason behind this change is more future proofing with the idea being that we may allow dotted notation for the AS numbers.  It seems at first glance BIRD doesn't support that at the moment, so we'll just convert an dotted AS numbers to a single number.  We're making a change to calicoctl to allow dotted AS numbers to be specified as well (but we'll actually just convert the value before storing in the datastore).  If BIRD starts supporting dotted notation then we can store them as dotted notation in the datastore.

cheuschober:
Hi @robbrockbank,

I noticed some of the other commits as well. No worries -- we were looking for a quick fix for some big AS's in our pre-prod mesos/calico/docker environment but the string implementation is, I think, the better option. Will close as obsolete.

PR #469, opened 2015-08-25
Rework confd templates to remove churn (requires confd change)

robbrockbank:
I've cut a confd release based on the two branches Cory implemented - install updated to pull in this release.


node_filesystem/templates/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+These templates write out a TOML file (bird.toml and bird6.toml) that is used 
+to tell confd which set of the main BIRD templates to use.
+
+Based off the node_mesh parameter, the TOML file generated either points to the
+full-mesh config files, or the no-mesh config files.
+
+Once confd writes out the appropriate TOML file, confd is restarted to pick up
+the change to the bird.toml and bird6.toml files.  Since there are two sets
+of changes, a change to the node_mesh parameter will result in a double
```
tomdee: I don't understand this double restart of confd thing? Won't it restart confd then confd will restart bird?

Ah! Or do you mean one restart for ipv4 and one for ipv6?

tomdee:
LGTM

lxpollitt:
Please holds off meeting this until I have also LGTM. I want to understand this since it sounds like HLD that I have not reviewed.

tomdee:
@lxpollitt We chatted about this and I think @robbrockbank talked to you about it too.

The headline was
* This adds some complexity
* It's not strictly needed for scale testing (but would probably make lives easier)
* And has no effect/behavior change when using full mesh
* It makes scaling much better when using a RR.

I'm inclined to merge it, since we've gone to the effort of writing and testing it. WDYT?

lxpollitt:
OK, so long as we are 100% convinced that confd failing halfway through updating some of its own templates can never result in us getting stuck in a broken state.

tomdee:
@robbrockbank Can you comment on what Alex says above?

robbrockbank:
@lxpollitt @tomdee 

confd writes temporary files which it moves into place, so there is no risk of a half written confd toml file.  Worst case is that the toml file is either not present (because it failed before writing the first one), or that the toml file is out of date.  However, in both cases this is rectified once confd is started again.

PR #476, opened 2015-08-27
validate as_num argument to support 4-byte asn

L-MA:


calico_containers/calico_ctl/bgp.py
```diff
@@ -54,10 +54,16 @@ def validate_arguments(arguments):
                     validate_ip(arguments["<PEER_IP>"], 4) or \
                     validate_ip(arguments["<PEER_IP>"], 6)
     asnum_ok = True
-    if arguments.get("<AS_NUM>"):
+    if arguments.get("<AS_NUM>") or arguments.get("--as"):
```
TrimBiggs: The `--as` param isn't used by the bgp command.

TrimBiggs: A couple other references to remove in this file too.

calico_containers/calico_ctl/bgp.py
```diff
@@ -54,10 +54,16 @@ def validate_arguments(arguments):
...
         try:
-            asnum = int(arguments["<AS_NUM>"])
-            asnum_ok = 0 <= asnum <= 4294967295
+            asnum = str(arguments.get("<AS_NUM>") or arguments.get("--as"))
+            if "." in asnum:
+                left_asnum, right_asnum = asnum.split(".")
+                if not (0 <= (int(left_asnum) | int(right_asnum)) <= 65535):
```
TrimBiggs: The `|` here is functioning as a bitwise "or", but I think you mean to be testing that each value on its own is >= 0 and also <= 65535.

calico_containers/calico_ctl/bgp.py
```diff
@@ -54,10 +54,16 @@ def validate_arguments(arguments):
...
+                    raise ValueError
+                asnum_ok = 0 <= (65536*int(left_asnum)+int(right_asnum)) <= 4294967295
```
TrimBiggs: This is pretty much testing the same thing as you meant to test in the if statement above: `65535.65535 == 4294967295`

You can just set `asnum_ok = <if statement logic>`

calico_containers/calico_ctl/node.py
```diff
@@ -109,8 +109,14 @@ def validate_arguments(arguments):
...
     if arguments.get("<AS_NUM>") or arguments.get("--as"):
         try:
-            asnum = int(arguments["<AS_NUM>"] or arguments["--as"])
-            asnum_ok = 0 <= asnum <= 4294967295
+            asnum = str(arguments.get("<AS_NUM>") or arguments.get("--as"))
+            if "." in asnum:
+                left_asnum, right_asnum = asnum.split(".")
+                if not (0 <= (int(left_asnum) | int(right_asnum)) <= 65535):
+                    raise ValueError
+                asnum_ok = 0 <= (65536*int(left_asnum)+int(right_asnum)) <= 4294967295
```
TrimBiggs: Same changes here as mentioned for bgp.py

calico_containers/calico_ctl/bgp.py
```diff
@@ -56,8 +56,12 @@ def validate_arguments(arguments):
     asnum_ok = True
     if arguments.get("<AS_NUM>"):
         try:
-            asnum = int(arguments["<AS_NUM>"])
-            asnum_ok = 0 <= asnum <= 4294967295
+            asnum = str(arguments.get("<AS_NUM>"))
```
robbrockbank: Can this get return None?  In that case str(None) will return a string "None".

I don't think you want the str()

robbrockbank:
An AS number may be specified as an integer OR the dotted notation.  You need to handle both situations.

Also, you've repeated this code twice.  Need to put this in a utils class.

calico_containers/calico_ctl/node.py
```diff
@@ -109,8 +109,12 @@ def validate_arguments(arguments):
     asnum_ok = True
     if arguments.get("<AS_NUM>") or arguments.get("--as"):
         try:
-            asnum = int(arguments["<AS_NUM>"] or arguments["--as"])
-            asnum_ok = 0 <= asnum <= 4294967295
+            asnum = str(arguments.get("<AS_NUM>") or arguments.get("--as"))
```
robbrockbank: Maybe pull this outside so that you don't have to repeat the gets in the if statement immediately above.

calico_containers/calico_ctl/node.py
```diff
@@ -109,8 +109,12 @@ def validate_arguments(arguments):
...
+            if "." in asnum:
+                left_asnum, right_asnum = asnum.split(".")
+                asnum_ok = 0 <= (int(left_asnum) and int(right_asnum)) <= 65535
```
robbrockbank: This logic is not correct.

`int(left_asnum) and int(right_asnum)`

will return you int(right_asnum) if int(left_asnum) is non-zero, otherwise it'll return 0.  Which reduces your expression to:

asnum_ok = 0 <= int(right_asnum) <= 65535

robbrockbank:
One final thing - you should add an ST for this as well - i'm not 100% sure whether BIRD actually supports this format.  I believe it does - but need an ST to ensure that

robbrockbank:
Yup that looks good!  Just the STs to write now.

calico_containers/calico_ctl/utils.py
```diff
@@ -179,3 +179,22 @@ def validate_hostname_port(hostname_port):
         return False
     else:
         return 1 <= port <= 65535
+
+def validate_asn(asn):
```
TrimBiggs: Nice commonizing here.

calico_containers/calico_ctl/utils.py
```diff
@@ -179,3 +179,22 @@ def validate_hostname_port(hostname_port):
...
+    """
+    Validate the format of a 4-byte autonomous system number
```
TrimBiggs: This validates 2-byte or 4-byte as nums.

calico_containers/calico_ctl/utils.py
```diff
@@ -179,3 +179,22 @@ def validate_hostname_port(hostname_port):
...
+
+    :param asn: User input of asn
```
TrimBiggs: User input of AS number

calico_containers/calico_ctl/utils.py
```diff
@@ -179,3 +179,22 @@ def validate_hostname_port(hostname_port):
...
+    :return: Boolean: True if valid format, False if invalid format
+    """
+    try:
+        if "." in str(asn):
+            left_asnum, right_asnum = asn.split(".")
+            asn_ok = (0 <= int(left_asnum) <= 65535) and \
+                       (0 <= int(right_asnum) <= 65535)
```
TrimBiggs: Mmmmm much better. Align those parens though.

calico_containers/tests/st/utils/constants.py
```diff
@@ -17,4 +17,4 @@
 DEFAULT_IPV4_ADDR_2 = "192.168.1.2"
 DEFAULT_IPV4_ADDR_3 = "192.168.1.3"
 DEFAULT_IPV6_POOL_CIDR = "fd80:24e2:f998:72d6::/64"
-LARGE_AS_NUM = "4294967290"
+LARGE_AS_NUM = "4294967200"
```
robbrockbank: You should make this dotted notation - that'll run through your changes and check it works.

TrimBiggs:
LGTM

PR #478, opened 2015-08-28
Resilient diags

robbrockbank:


TrimBiggs:
Looks like a good change, but Circle is failing. Good to merge once it's fixed.

Also might be worth writing a diags ST at this point if we don't have one yet.

Issue #443, opened 2015-08-13
Move to use upstream confd


fasaxc:
This seems to be the root-cause of the etcd issues that I was hitting in my scale tests:

confd generates O(n^2) work in regenerating the BIRD config, which is expected.  However, it seems to always pick a single etcd cluster node to talk to, which results in overloading the chosen node.

I created an upstream bug to track https://github.com/kelseyhightower/confd/issues/319 but we may want to address this through using confd differently or abandoning it all together.

(@matthewdupre edit: "original title confd overloads etcd when bringing up a lot of hosts")

tomdee:
Adding blocked label as this is currently being treated as an upstream issue. We might be able to contribute a fix upstream or as Shaun suggests use something else.

Lukasa:
We are also considering other confd changes to improve this: see kelseyhightower/confd#326.

robbrockbank:
We have a projectcalico/confd release that addresses this issue.  I'll update our node image to pull in that release. 

Cory has created an upstream PR - so we should use the upstream release when it's available.

robbrockbank:
Fixed by #483 

Once this fix is upstream in confd, we should revert to using the standard confd.

fasaxc:
@robbrockbank @tomdee Looks like we now have the upstream fix: https://github.com/kelseyhightower/confd/issues/319

Re-opening as a reminder to switch to upstream confd, if we can.

tomdee:
Let take v0.12 as soon as it comes out - https://github.com/kelseyhightower/confd/releases

caseydavenport:
What do we need to do to close this out? Just pull in the latest upstream release here and test? https://github.com/projectcalico/calico-containers/blob/master/calico_node/build.sh#L30

Do we still have any outstanding patches in our fork that haven't been upstreamed?

caseydavenport:
@fasaxc - Assigned this guy to you for tracking / retest when the new release comes out. Looks like we just need to wait... hasn't been a release in a while.

pizzarabe:
I would like to see https://github.com/kelseyhightower/confd/pull/329 in calico node, because I can see a bunch of `closed request prematurely` using etcdv2. 
The fix is in the 0.12 version (using a new etcd lib)  which is still in alpha so I guess we will wait for the 0.12 ro be released.

fasaxc:
@heschlie did you work to do a k8s back-end for confd include revving to v0.12?

heschlie:
[...]

PR #483, opened 2015-09-08
Use projectcalico modified confd

robbrockbank:
Quick change to resolve issue #443 

Change basically uses an updated version of confd (in our own fork).  Shaun has tested the binary and testing indicates that it is working correctly.


TrimBiggs:
LGTM once tests pass.

PR #495, opened 2015-09-21
Change env var POLICY_ONLY_CALICO -> CALICO_NETWORKING

L-MA:
CALICO_NETWORKING defaults to true. When set to false, bird and confd are disabled.

The reason for the name change is to avoid confusion in the plugins. In the Kubernetes plugin we use an environment variable called CALICO_POLICY that indicates when the user wants to specify policy for pods. There should be a clear distinction between this environment variable and the calicoctl environment variable that disables bird.


caseydavenport:
This change also avoids confusion, since POLICY_ONLY mode is a bit of a misnomer, since we still do veth manipulation, route installation, etc.

caseydavenport:
Should be noted that this is a non-backwards compatible change.  I think its OK since it isn't a publicly advertised API, but I'll leave that up to you guys to decide.

tomdee:
LGTM

PR #498, opened 2015-09-24
Install a more recent version of pip/setuptools.

fasaxc:
python-setuptools is old enough to have a bug that balks on python-etcd's version number (0.4.1+calico.1).  This works around by getting an up-to-date pip and setuptools.

Had to install ca-certificates manually, they were being pulled in as an indirect dependency and they're required to run curl.


tomdee:
LGTM

PR #502, opened 2015-09-25
Fix documentation for KUBE_API_ROOT env variable

Symmetric:
The docs for the KUBE_API_ROOT were unclear; they did not specify that the transport (http/https) must be specified to construct a valid URI. The example of the default value was also incorrect.

Fix this by giving a correct example of the default value.

FIXES: https://github.com/projectcalico/calico-kubernetes/issues/44


caseydavenport:
LGTM

PR #504, opened 2015-09-28
Tell felix not to log to file.

### LINKED ISSUE #480: Multiple felix logfiles
I think the current felix.cfg results in multiple felix logfiles.  There's one generated from stdout/err in /var/log/calico/felix/current and a separate one that felix creates by default at /var/log/calico/felix.log.

tomdee:
Fixes #480


caseydavenport:
LGTM.

Presumably if `LogFilePath` gets overridden, we'll still log to two locations?

tomdee:
Yep

PR #506, opened 2015-09-29
Automatically assign IPs

tomdee:
When using "container add" or "container ip add" add the option to assign IP addresses automatically.

PR #508, opened 2015-09-30
Adding auto_assigning IP logic to containers

insequent:
This is almost exactly the same as #506, but with a few extra bits. We clearly had the same idea in mind! Great minds.

I'm still adding a few more tests, but since there's already another PR for the same type of change, I thought I would throw mine up for discussion. I don't expect the tests to pass yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ calico_containers/calico-node.tar: caliconode.created
 	docker save --output calico_containers/calico-node.tar calico/node
 
 st: binary calico_containers/busybox.tar calico_containers/routereflector.tar calico_containers/calico-node.tar run-etcd
-	dist/calicoctl checksystem --fix
+	sudo dist/calicoctl checksystem --fix
 	nosetests $(ST_TO_RUN) -sv --nologcapture --with-timer
 
 fast-st: calico_containers/busybox.tar calico_containers/routereflector.tar calico_containers/calico-node.tar run-etcd
--- a/calico_containers/calico_ctl/container.py
+++ b/calico_containers/calico_ctl/container.py
@@ -7,13 +7,20 @@ Usage:
   calicoctl container remove <CONTAINER>
 
 Description:
-  Add or remove containers to calico networking, manage their IP addresses and profiles.
+  Add or remove containers to Calico networking, manage their IP addresses and profiles.
   All these commands must be run on the host that contains the container.
 
 Options:
+  <IP>                     The IP address desired. If "ipv4", "ipv6", or a CIDR
+                           is given, then Calico will attempt to automatically
+                           assign an available IPv4 address, IPv6 address, or
+                           IP from any Pool matching the provided CIDR,
+                           respectively. NOTE: When a CIDR is passed, it must
+                           exactly match an existing Calico pool.
   --interface=<INTERFACE>  The name to give to the interface in the container
                            [default: eth1]
 """
+
 import os
 import sys
 import uuid
@@ -25,7 +32,7 @@ from subprocess import CalledProcessError
 from netaddr import IPAddress, IPNetwork
 from calico_ctl import endpoint
 from pycalico import netns
-from pycalico.datastore_datatypes import Endpoint
+from pycalico.datastore_datatypes import IPPool, Endpoint
 
 from connectors import client
 from connectors import docker_client
@@ -33,8 +40,26 @@ from utils import hostname, DOCKER_ORCHESTRATOR_ID, NAMESPACE_ORCHESTRATOR_ID, \
     escape_etcd
 from utils import enforce_root
 from utils import print_paragraph
+from utils import validate_cidr
 from utils import validate_ip
 
+
+def assign_any(v4_count, v6_count, pool=(None, None)):
+    """
+    Reserve <count> IP(s) from the datastore to be applied to a container
+
+    :param arguments: v4_count = Count of IPv4 addresses
+                      v6_count = Count of IPv6 addresses
+                      pool = tuple(<IPv4 cidr>, <IPv6 cidr>)
+    :return: tuple(list(IPv4 IPAddresses), list(IPv6 IPAddresses))
+    """
+    v4_list, v6_list = client.auto_assign_ips(v4_count, v6_count, None, {},
+                                              pool=pool)
+    if not any((v4_list, v6_list)):
+        sys.exit("Failed to allocate any IPs (requested {0} IPv4s and {1} IPv6s). "
+                 "Pools are likely exhausted.".format(v4_count, v6_count))
+
+    return (v4_list, v6_list)
 
 def validate_arguments(arguments):
     """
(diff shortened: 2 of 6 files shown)


tomdee:
Thanks for putting this up. It's more feature rich so I'd take it in preference to #506. 

Would you consider extending it to the `calicoctl container <CONTAINER> ip (add|remove)` commands too?

Before I can merge the PR, you'll need to sign a contributor agreement - see here http://docs.projectcalico.org/en/latest/contribute.html#contributor-agreements

insequent:
Hey @tomdee,

Thanks for the quick response! The change already extends to `calicoctl container <CONTAINER> ip (add|remove)`. You can see it with the `get_ip_and_pool` call within `def container_ip_add`. I ran the commands a few times in testing and it appears to work correctly. Sorry for the delay on the tests, as I'm having a bugger of a time getting `make test` to work locally. I should be able to finish it up by tomorrow however.

Also, although I haven't submitted code yet, I did already sign the contributor agreement. I have an email from Jon Flynn about it. Let me know if you need anything else from me on this front.

tomdee:
Sounds good! If you need help getting tests running I can offer support on our slack channel - sign up here https://calicousers-slackin.herokuapp.com

calico_containers/calicoctl.py
```diff
@@ -110,6 +110,6 @@
         print_paragraph(e.message)
         sys.exit(1)
     except BaseException as e:
-        print "Unexpected error executing command.\n"
+        print "{0}: {1}\n".format(type(e).__name__, e)
```
insequent: Added this more to clarify in troubleshooting. If this isn't desired, I can certainly take it out.

Makefile
```diff
@@ -78,7 +78,7 @@ calico_containers/calico-node.tar: caliconode.created
 	docker save --output calico_containers/calico-node.tar calico/node
 
 st: binary calico_containers/busybox.tar calico_containers/routereflector.tar calico_containers/calico-node.tar run-etcd
-	dist/calicoctl checksystem --fix
+	sudo dist/calicoctl checksystem --fix
```
insequent: `make` commands should not be forced to run as root, however this will cause the user to be prompted unless they have sudoers `NOPASSWD` or the like.

insequent:
Assuming this latest build passes the tests (the `st` tests still fail for me locally), this PR should be good to go.

insequent:
Rebased this morning to make sure we're not drifting.

tomdee:
A few comments but it's looking good. My main concern is about the skipped IPv6 tests.

@lxpollitt or @spikecurtis are you happy with this design? It's fine by me...

insequent:
@tomdee: As to IPv6 not working, I don't know. Unfortunately, I still cannot get `st` tests to work locally, but Semaphore's log showed that node was not configured for IPv6 (https://semaphoreci.com/calico/calico-docker--5/branches/pull-request-508/builds/10). I tried to get it to work by adding a `calicoctl checksystem --fix` to the node setup in tests, but it didn't seem to help. The other container tests didn't have IPv6 either.

calico_containers/calico_ctl/container.py
```diff
@@ -7,13 +7,20 @@
...
 
 Description:
-  Add or remove containers to calico networking, manage their IP addresses and profiles.
+  Add or remove containers to Calico networking, manage their IP addresses and profiles.
   All these commands must be run on the host that contains the container.
 
 Options:
+  <IP>                     The IP address desired. If "ipv4", "ipv6", or a CIDR
+                           is given, then Calico will attempt to automatically
+                           assign the first available IPv4 address, IPv6
```
spikecurtis: "the first available" -> "an available"

I don't want to imply anything concrete about allocation strategy.  It currently obeys host affinity, so will not, in general, pick the lowest address in a pool.  We may later implement other changes to allocation strategy.

calico_containers/calico_ctl/container.py
```diff
@@ -443,11 +477,44 @@ def container_ip_remove(container_id, ip, interface):
...
+            result = assign_any(1, 0)
+            ip = result[0][0]
+        elif '6' in ip:
+            result = assign_any(0, 1)
+            ip = result[1][0]
+        pool = get_pool_or_exit(ip)
+    elif ip is not None and '/' in ip:
+        requested_pool = IPNetwork(ip)
+        requested_version = requested_pool.version
+        pool = get_pool_or_exit(requested_pool.ip)
```
spikecurtis: We've already confirmed the pool exists in `validate_arguments()`.  You can just create the pool object directly
```
pool = IPPool(ip)
```

An `IPPool` subclasses `IPNetwork` so you can then do `pool.version` and the like.

This saves a few redundant reads of the datastore.

calico_containers/calico_ctl/container.py
```diff
@@ -206,11 +247,8 @@ def container_add(container_id, ip, interface):
         next_hops[ip.version]
     except KeyError:
         print "This node is not configured for IPv%d." % ip.version
-        sys.exit(1)
-
-    # Assign the IP
-    if not client.assign_address(pool, ip):
-        print "IP address is already assigned in pool %s " % pool
+        # Ignoring return as we don't care if the IP is already unallocated
```
spikecurtis: Can you expand this comment.  It wasn't immediately clear to me that `get_ip_and_pool()` actually does the IP allocation, so we need to `release_ips()` here to clean up.

Furthermore, I think we do care about the return value of `release_ips()`, at least from a debugging standpoint.  If `ip` is not allocated then something wrong and unexpected has happened.  We should print an error.

calico_containers/calico_ctl/container.py
insequent: Easy fix, will do.

spikecurtis:
I'm happy with the design and just made a few small suggestions in the code.

I do think we want more regression tests to go with this change.  I couldn't find code coverage output on CircleCI.  @tomdee aren't we supposed to get that from Circle, or am I thinking of a different repo?

calico_containers/calico_ctl/container.py
insequent: It's been a while since I wrote that comment, but I believe there were different branches involved at first where the IP may not have been allocated at this point. Seeing as all roads in `get_ip_and_pool()` do assign now, then certainly, a failed release_ips() here would be odd. I'll add a check.

calico_containers/calico_ctl/container.py
insequent: :+1: Good call

tomdee:
We have coverage on coveralls.io but I'm not sure if it's working on PRs
https://coveralls.io/github/projectcalico/calico-docker

plwhite:
I've also implemented something that does this, but I'm happy to ditch mine and replace with this (which on the face of it is more feature rich).

tomdee:
OK, I'm happy on the IPv6 tests. I forgot we didn't have them working. @robbrockbank was going to take a look at getting them working on Semaphore but that shouldn't block us merging this PR.

How can I help you get the STs running locally. What error output are you seeing?

tomdee:
So I think we're pretty much ready to merge this. THe final thing to check is the test coverage.

spikecurtis:
@tomdee I'm happy as long as test coverage looks good.

insequent:
Hey all. I'm back in the states, and thus have _much_ more reliable internet. Let me know whatever you need on this.

tomdee:
CircleCI has test coverage - which is higher than we've got on master at the moment. So merging.

@insequent I would love for you to be able to run the STs. I'm around all day today if I can offer you any help on Slack.

PR #517, opened 2015-10-08
Add route aggregation based on IPAM

plwhite:
This fix adds route aggregation. If a block is assigned to a host then : 

* a black hole route to that block is added to the host and exported through bird;
* any routes within that block are not advertised.

I've manually tested that it works. Note that it relies on https://github.com/projectcalico/libcalico/pull/30 which creates empty directories for IPAM to use.


spikecurtis:
LGTM, but it's failing on semaphore, and I can't immediately figure out why.  I'm not comfortable merging until we get to the bottom of it.

robbrockbank:
I rebased and this this has now been merged.  Closing PR.

PR #525, opened 2015-10-12
Clean up build and UT container

tomdee:
- Remove etcd (and curl since it was only needed to install etcd)
- Remove outdated comments
- Remove old requirements


robbrockbank:
LGTM

PR #526, opened 2015-10-12
Use upstream etcd

tomdee:


build_calicoctl/requirements.txt
```diff
@@ -14,5 +14,5 @@ netaddr==0.7.15
 flask
 gunicorn
 subprocess32
-git+https://github.com/projectcalico/python-etcd.git
+git+https://github.com/jplana/python-etcd.git
```
spikecurtis: why install from github instead of pypi?  e.g.
```
python-etcd>=0.4.2
```

calico_containers/requirements.txt
```diff
@@ -2,7 +2,7 @@ sh==1.11
 netaddr==0.7.15
 six==1.9.0
 subprocess32
-git+https://github.com/projectcalico/python-etcd.git
+git+https://github.com/jplana/python-etcd.git
```
spikecurtis: We don't need this; `calico` and `libcalico` have it as deps.  Now that we're not dependent on our own fork, just let pip sort it out.  We don't directly import it anyway.

build_calicoctl/requirements.txt
spikecurtis: Actually, I don't think we need this here either.  `calicoctl` shouldn't be directly importing it, so just let the `libcalico` dep sort it out.

spikecurtis:
LGTM! :shipit:

PR #534, opened 2015-10-15
Create ip6tables command object within try/except

alexhersh:
+ Fixes projectcalico/calico-docker#531
+ Command obj creation will cause check_system to fail silently without hitting our loud exception handlers.


tomdee:
Were you able to manually test this? If so how?

alexhersh:
I removed `/sbin/ip6tables` and then ran `calicoctl node`. Without this patch, I got your output. With it, calico-node started with the following warning:
```WARNING: Unable to detect the ip6_tables module. Load with `modprobe ip6_tables````

alexhersh:
Oh, it appears performing `modprobe ip6_tables` doesnt fix the issue if you remove the binary.

PR #542, opened 2015-10-16
Remove check_ip_forwarding from check_system

### LINKED ISSUE #536: Calicoctl node should set the ip forwarding sysctls everytime it starts
They are required for proper Calico operation, and it's a very fast operation.

It's also what Docker does.

Having these related to the checksystem command is wrong.

alexhersh:
+ Fixes #536
+ node_start now sets up ip forwarding with sysctl
+ Reconfigure UTs

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -67,6 +67,7 @@ from prettytable import PrettyTable
 from connectors import client
 from connectors import docker_client
 from utils import DOCKER_ORCHESTRATOR_ID
+from utils import sysctl
 from utils import hostname
 from utils import print_paragraph
 from utils import get_container_ipv_from_arguments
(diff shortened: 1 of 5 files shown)


tomdee:
You'll need to rebase - there are conflicts

calico_containers/calico_ctl/node.py
```diff
@@ -393,6 +397,22 @@ def _start_libnetwork_container(etcd_authority, libnetwork_image):
...
 
 
+def _setup_ip_forwarding():
+    """
+    Ensure that IP forwarding is enabled.
+    :return: None
+    """
+    # Enable IP forwarding since all compute hosts are vRouters.
+    # IPv4 forwarding should be enabled already by docker.
+    if "1" not in sysctl("net.ipv4.ip_forward"):
```
tomdee: I don't think there's any need to read this then write it. We should just write it every time.

And rather than writing using sysctl, let just use Python

with open('/proc/sys/net/ipv4/ip_forward', 'w') as f:
   f.write("1")

Then we can get rid of one more use of the sh module.

tomdee:
LGTM (if the tests pass)

PR #543, opened 2015-10-16
Check for docker API v1.21 if using libnetwork

### LINKED ISSUE #537: Calicoctl should check for the correct docker version when using the --libentwork flag
--libnetwork requires Docker 1.9

alexhersh:
+ Fix #537
+ Add the checksystem --libnetwork flag
+ Fix UTs for checksystem/node

### DIFF
--- a/calico_containers/calico_ctl/checksystem.py
+++ b/calico_containers/calico_ctl/checksystem.py
@@ -68,9 +68,11 @@ def check_system(fix=False, quit_if_error=False, libnetwork=False):
     # modprobe and sysctl require root privileges.
     enforce_root()
 
-    system_ok = (_check_kernel_modules(fix) and
-                 _check_ip_forwarding(fix) and
-                 _check_docker_version(libnetwork))
+    kernel_ok = _check_kernel_modules(fix)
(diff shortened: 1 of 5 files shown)


calico_containers/calico_ctl/checksystem.py
```diff
@@ -66,7 +70,7 @@ def check_system(fix=False, quit_if_error=False):
 
     system_ok = (_check_kernel_modules(fix) and
                  _check_ip_forwarding(fix) and
-                 _check_docker_version())
+                 _check_docker_version(libnetwork))
```
TrimBiggs: I thought about this a bit more after you were having trouble hitting the `_check_docker_version` function when one of the previous functions in this statement returned `False`.

I don't like it.  If `_check_kernel_modules` returns False, this variable will immediately be set to false and the other two functions will not be called.  If your system is not configured correctly, you should get full output for everything that is wrong.

Though it's not as pretty, I think the function return values should be stored as separate variables, then checked here.

alexhersh: fixed :+1:

tomdee:
LGTM

PR #556, opened 2015-10-22
Reduce the size of the calico/node image (299.3MB -> 78.12MB)

tomdee:
- Move to using Alpine as the base (instead of ubuntu)
- Move "node" files to be in their own directory.
- Greatly simplify the build (78 additions and 511 deletions.)


Makefile
```diff
@@ -111,7 +111,7 @@ clean:
 	-rm -f calico_containers/calico-node.tar
 	-rm -f calico_containers/routereflector.tar
 	-docker rm -f calico-build
-	-docker rm -f calico-node
+	-docker rm -f calico_node
```
robbrockbank: The Docker image hasn't changed name has it?  Shouldn't this still be "calico-node"?

robbrockbank:
One comment - once resolved LGTM.

PR #563, opened 2015-10-23
Use $HOSTNAME instead of socket.gethostname

djosborne:


calico_containers/calico_ctl/container.py
```diff
@@ -133,7 +134,7 @@ def container(arguments):
         elif arguments.get("endpoint"):
             orchestrator_id, workload_id = \
                                   lookup_workload(arguments.get("<CONTAINER>"))
-            endpoint.endpoint_show(hostname,
+            endpoint.endpoint_show(get_hostname(),
```
djosborne: Is calling a function in your function call considered bad practice?

robbrockbank: Not if you're using it just the once.  However, you are calling this function several times in a row - which is generally less efficient than storing the result in a local and using that same result multiple times.

calico_containers/calico_ctl/utils.py
```diff
@@ -26,7 +25,6 @@
 DOCKER_ORCHESTRATOR_ID = "docker"
 NAMESPACE_ORCHESTRATOR_ID = "namespace"
 REQUIRED_MODULES = ["xt_set", "ip6_tables"]
-hostname = socket.gethostname()
```
robbrockbank: Is there any reason you couldn't have just changed this one line to:

```hostname = pycalico.util.get_hostname()```

or is there a reason we have to dynamically get the hostname everytime?  I assume it's not really going to work properly if the hostname changes under our feer, so I'd have thought setting up the hostname at start of day would still be sufficient.

robbrockbank:
Dan, I'm closing this in preference to:

https://github.com/projectcalico/calico-docker/pull/589

As discussed, I think we can make a much simpler change and set up the hostname in utils at start of day.  Have a peek and see if you agree.

PR #566, opened 2015-10-23
Add FAQ question about reserved IP addresses

TrimBiggs:


docs/FAQ.md
```diff
@@ -57,7 +57,15 @@ node on which each container is hosted.
 ```
 Where `<CIDR>` is the CIDR of your IP pool, for example `192.168.0.0/16`.
 
-Remember: the security profile for the container will need to allow traffic to the internet as well. You can read about how to configure security profiles in the [Advanced Network Policy](AdvancedNetworkPolicy.md) guide.
+Remember: the security profile for the container will need to allow traffic to the 
+internet as well. You can read about how to configure security profiles in the 
+[Advanced Network Policy](AdvancedNetworkPolicy.md) guide.
+
+## Can Calico containers use any IP address within a pool, even reserved address?
```
spikecurtis: reserved address -> subnet network/broadcast addresses

docs/FAQ.md
```diff
@@ -57,7 +57,15 @@ node on which each container is hosted.
...
+
+Yes!  Calico is fully routed, so all IP address within a Calico pool are usable as 
+private IP addresses to assign to a workload.  This means that commonly reserved IP
```
spikecurtis: "This means addresses commonly reserved in a L2 subnet, such as IPv4 addresses..."

docs/FAQ.md
spikecurtis: Need one more # to fit with rest of formatting.

PR #572, opened 2015-10-28
Removed outdated mesos files

djosborne:
These mesos files were checked in 7 months ago and are no longer relevant.


tomdee:
LGTM

PR #574, opened 2015-10-28
Master docs master code

robbrockbank:


robbrockbank:
Tom, not complete.  Just wanted to get your opinion on the banner at the top of the page.  Perhaps remove the icon as nice formatting is a little bit of a pain in MD unless you start getting into CSS which is not always supported.

-  I'd planned to add this to every MD document, where the link to the latest release will always take you to the top level README (since we'd have to worry about pages only existing at certain versions otherwise).

WDYT?

I'd like to get the docs changes in as one PR, and then I'll have another PR for implementing the release scripts.

tomdee:
Can't we just link to the latest release using this https://help.github.com/articles/linking-to-releases/

tomdee:
I don't like the warning.png in the root dir

robbrockbank:
I tried that.  I believe the linking to latest release just links to the release page, rather than the "latest" branch of source code - which is what we'd need

Yeah I'll move the warning.png - it was never my intention to keep it there.

docs/Building.md
```diff
@@ -96,10 +96,6 @@ The full suite of system tests takes approximately 30 minutes to complete.  If y
 Docker, and are regularly running the system tests, a faster subset of system tests may be run which covers most
 mainline function.
 
-Both ST suites utilize Docker's [libnetwork network driver](https://github.com/docker/libnetwork), available in the
```
tomdee: I prefer not to have different changes in a single PR. I've just deleted this block in a different commit

tomdee:
Yep, approach looks good to me.

Can we get HTTPS on the projectcalico.org URLs? If not, why not?

tomdee:
All looks good to me. Merge away!

PR #578, opened 2015-10-29
Add check for nonexistant if when adding IP

TrimBiggs:


robbrockbank:
LGTM once the tests pass (I assume this is because the libcalico change is not yet in).

TrimBiggs:
Required libcalico fix: https://github.com/projectcalico/libcalico/pull/45

Issue #558, opened 2015-10-22
`Calicoctl endpoint <ep_id> (append|set|remove) profile` contains undocumented parameters


TrimBiggs:
This came up after raising issue #557 

I imagine this command used to be merged with the `endpoint <ENDPOINT_ID> profile show command`, which would explain this and why the `<PROFILES>` param is optional here.

Something like this:
```
calicoctl endpoint <ENDPOINT_ID> profile (append|remove|set|show) [--host=<HOSTNAME>] 
    [--orchestrator=<ORCHESTRATOR_ID>] [--workload=<WORKLOAD_ID>]  [<PROFILES>...]
```

but this command should look like this:
```
calicoctl endpoint <ENDPOINT_ID> profile (append|remove|set) <PROFILES>...
```

If the endpoint is required, we should not need any additional information about the hostname/orchestrator_id/workload_id to know which endpoint we are working with.

robbrockbank:
It is correct that the endpoint ID is generally sufficient to identify an endpoint.  However, performing an endpoint lookup based purely on the endpoint ID requires a linear search through the etcd datastore to locate the endpoint.

Whilst this is OK for testing by hand, it does not make sense in a large deployment where there will be an orchestration layer.  In this case you can specify all of the IDs for faster lookup (no linear scan required).

PR #590, opened 2015-11-03
Added docs for endpoint commands passing in additional lookup params

TrimBiggs:


TrimBiggs:
Fixes #558

docs/calicoctl/endpoint.md
```diff
@@ -162,6 +162,11 @@ calicoctl endpoint <ENDPOINT_ID> profile (append|remove|set) [--host=<HOSTNAME>]
     <PROFILES>: One or more profiles to append, remove, or set on the endpoint.
 ```
 
+This command perform the desired action even if the `<ENDPOINT_ID>` is specified
```
robbrockbank: I'd word it so you can use identical text in both cases, so i'd remove the details about what the command does:

"The <ENDPOINT ID> uniquely identifies the endpoint.  Although it is not necessary to include the optional <HOSTNAME>, <ORCHESTRATOR_ID> and <WORKLOAD_ID> identifiers, this command executes faster and with reduced load on the etcd datastore when all of the identifiers are specified together.

robbrockbank: But i'm not too fussed either way.

Wording is a little wrong as it stands though...   "This command perform"   missing s?

docs/calicoctl/endpoint.md
```diff
@@ -162,6 +162,11 @@ calicoctl endpoint <ENDPOINT_ID> profile (append|remove|set) [--host=<HOSTNAME>]
     <PROFILES>: One or more profiles to append, remove, or set on the endpoint.
 ```
 
+This command perform the desired action even if the `<ENDPOINT_ID>` is specified 
+without the `<HOSTNAME>`, `<ORCHESTRATOR_ID>`, or `<WORKLOAD_ID>`.  However, 
+the lookup to retrieve the correct endpoint will execute faster if any or all 
+of these parameters are passed in as well.
```
robbrockbank: I'd just say "all" not "any"

PR #594, opened 2015-11-04
Run `calicoctl node` without launching docker

### LINKED ISSUE #564: It should be possible to set up a node without starting calico (felix/bird/confd)
Operators may want to choose their own way of deploying the `calico-node` components i.e. felix, bird and confd. We currently only support using a docker container, but we do have native packages in our openstack repos. In addition to running the components natively, operators may want to use other container formats such as rkt.

The expected way that we'll support this is to add a new `calicoctl node --runtime=<RUNTIME>` command. It can take the following values
* docker - which is the current and default behaviour
* none - which doesn't run felix/bird/confd at all.

It future we may add other options.

djosborne:
Base implementation to fix #564 . Adds the `--runtime` option with possible values of `docker` or `none`. This PR also patches some bugs encountered by `checksystem` and `status` which both assume a connection to the docker-daemon is available. 

Additional design is needed to provide more robust checksystem and status for dockerless systems, but I propose we raise separate issues in calico-docker to address improving these.

### DIFF
--- a/calico_containers/calico_ctl/node.py
+++ b/calico_containers/calico_ctl/node.py
@@ -14,7 +14,8 @@
 """
 Usage:
   calicoctl node [--ip=<IP>] [--ip6=<IP6>] [--node-image=<DOCKER_IMAGE_NAME>]
-    [--as=<AS_NUM>] [--log-dir=<LOG_DIR>] [--detach=<DETACH>] [--rkt]
+    [--runtime=<RUNTIME>] [--as=<AS_NUM>] [--log-dir=<LOG_DIR>]
+    [--detach=<DETACH>] [--rkt]
     [(--kubernetes [--kube-plugin-version=<KUBE_PLUGIN_VERSION])]
     [(--libnetwork [--libnetwork-image=<LIBNETWORK_IMAGE_NAME>])]
   calicoctl node stop [--force]
@@ -35,6 +36,11 @@ Options:
                             "false" to run in the foreground.  When using
                             libnetwork, this may not be set to "false".
                             [default: true]
+  --runtime=<RUNTIME>       Specify how Calico services should be
+                            launched.  When set to "docker", services will be
+                            launched via the calico-node container, whereas a
+                            value of "none" will not launch them at all.
+                            [default: docker]
   --log-dir=<LOG_DIR>       The directory for logs [default: /var/log/calico]
   --ip=<IP>                 The local management address to use.
   --ip6=<IP6>               The local IPv6 management address to use.
@@ -119,6 +125,7 @@ def validate_arguments(arguments):
     peer_ip_ok = arguments.get("<PEER_IP>") is None or \
                  validate_ip(arguments["<PEER_IP>"], 4) or \
                  validate_ip(arguments["<PEER_IP>"], 6)
+    runtime_ok = arguments.get("--runtime") in [None, "none", "docker"]
 
     asnum_ok = True
     asnum = arguments.get("<AS_NUM>") or arguments.get("--as")
@@ -145,10 +152,12 @@ def validate_arguments(arguments):
         print "Valid values for --detach are 'true' and 'false'"
     if not detach_libnetwork_ok:
         print "The only valid value for --detach is 'true' when using libnetwork"
+    if not runtime_ok:
+        print "Runtime must be 'docker' or 'none'."
 
     # Exit if not valid argument
     if not (ip_ok and ip6_ok and container_ip_ok and peer_ip_ok and asnum_ok
-            and detach_ok and detach_libnetwork_ok):
+            and detach_ok and detach_libnetwork_ok and runtime_ok):
         sys.exit(1)
 
 
@@ -190,6 +199,7 @@ def node(arguments):
                                 else arguments.get("--libnetwork-image")
         node_start(ip=arguments.get("--ip"),
                    node_image=arguments.get('--node-image'),
+                   runtime=arguments.get("--runtime"),
                    log_dir=arguments.get("--log-dir"),
                    ip6=arguments.get("--ip6"),
                    as_num=as_num,
@@ -199,7 +209,7 @@ def node(arguments):
                    libnetwork_image=libnetwork_image)
 
 
-def node_start(node_image, log_dir, ip, ip6, as_num, detach,
+def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
                kubernetes_version, rkt, libnetwork_image):
     """
     Create the calico-node container and establish Calico networking on this
@@ -230,7 +240,9 @@ def node_start(node_image, log_dir, ip, ip6, as_num, detach,
         pass
 
     # Print warnings for any known system issues before continuing
-    check_system(quit_if_error=False, libnetwork=libnetwork_image)
+    using_docker = True if runtime == 'docker' else False
+    check_system(quit_if_error=False, libnetwork=libnetwork_image,
+                 check_docker=using_docker)
 
     # We will always want to setup IP forwarding
     _setup_ip_forwarding()
@@ -297,10 +309,10 @@ def node_start(node_image, log_dir, ip, ip6, as_num, detach,
     etcd_authority_address, etcd_authority_port = etcd_authority.split(':')
     etcd_authority = '%s:%s' % (socket.gethostbyname(etcd_authority_address),
                                 etcd_authority_port)
-
-    _start_node_container(ip, ip6, etcd_authority, log_dir, node_image, detach)
-    if libnetwork_image:
-        _start_libnetwork_container(etcd_authority, libnetwork_image)
+    if runtime == 'docker':
+        _start_node_container(ip, ip6, etcd_authority, log_dir, node_image, detach)
+        if libnetwork_image:
+            _start_libnetwork_container(etcd_authority, libnetwork_image)
 
 
 def _start_node_container(ip, ip6, etcd_authority, log_dir, node_image, detach):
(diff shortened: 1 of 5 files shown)


robbrockbank:
I know this isn't up for review yet - but just a note:  Don't forget to also update the calicoctl documentation.  Ta!

calico_containers/calico_ctl/node.py
```diff
@@ -35,6 +36,8 @@
                             "false" to run in the foreground.  When using
                             libnetwork, this may not be set to "false".
                             [default: true]
+  --runtime=<RUNTIME>       Specify how calico-node processes should be
```
spikecurtis: "Specify how Calico services..."

I don't want to ossify around `calico-node` terminology.  Let's leave that term just to mean the Docker container image.

calico_containers/calico_ctl/node.py
```diff
@@ -35,6 +36,8 @@
...
+                            launched.  Can be: docker, none. [default: docker]
```
spikecurtis: We probably need some explanation of what "none" means.

calico_containers/calico_ctl/status.py
```diff
@@ -35,6 +36,13 @@ def status(arguments):
     this file's docstring with docopt
     :return: None
     """
+    try:
+        docker_client.version()
```
spikecurtis: Why not just wrap the call below in try except?

calico_containers/calico_ctl/node.py
```diff
@@ -536,24 +543,31 @@ def warn_if_hostname_conflict(ip):
...
-                print_paragraph("WARNING: Hostname '%s' is already in use "
-                                "with IP address %s. Calico requires each compute host to "
-                                "have a unique hostname. If this is your first time "
-                                "running 'calicoctl node' on this host, ensure that "
-                                "another host is not already using the "
-                                "same hostname." % (hostname, ip))
+    try:
+        if docker_client.containers(filters={'name': 'calico-node'}):
+            return
+    except IOError:
```
spikecurtis: I think we still want to do the hostname conflict warning, even if the user doesn't have Docker running.

calico_containers/calico_ctl/status.py
```diff
@@ -35,6 +36,13 @@ def status(arguments):
...
+    except ConnectionError:
```
spikecurtis: In this check you seem to be using "Is the Docker daemon running?" as a proxy for "Did the user start the node with `--runtime=none`?"  These aren't the same thing.  You could use `--runtime=none` even though Docker is running, for example. In that case, won't this function fail?

docs/calicoctl/node.md
```diff
@@ -112,6 +119,10 @@ When running the `calicoctl node` command with the `--libnetwork` plugin, the
 command starts a container using the `calico/node-libnetwork` Docker image in 
 addition to starting the `calico/node` Docker image.
 
+The `--runtime` flag can be used to prevent calico from launching the
+calico-node container, instead allowing you to run the core processes
+on baremetal.
```
spikecurtis: Bare metal is a bit misleading.  You might use this in a VM, for example.

"... run the core Calico services via some other means. For example, you might install directly on the host and execute them via a systemd unit."

docs/calicoctl/node.md
```diff
@@ -112,6 +119,10 @@ When running the `calicoctl node` command with the `--libnetwork` plugin, the
 command starts a container using the `calico/node-libnetwork` Docker image in 
 addition to starting the `calico/node` Docker image.
 
+The `--runtime` flag can be used to prevent calico from launching the
+calico-node container, instead allowing you to run the core processes
```
spikecurtis: "calico-node Docker container"

docs/calicoctl/node.md
```diff
@@ -112,6 +119,10 @@ When running the `calicoctl node` command with the `--libnetwork` plugin, the
 command starts a container using the `calico/node-libnetwork` Docker image in 
 addition to starting the `calico/node` Docker image.
 
+The `--runtime` flag can be used to prevent calico from launching the
```
spikecurtis: Calico

spikecurtis: `--runtime=none` setting

docs/calicoctl/node.md
```diff
@@ -71,7 +74,9 @@ Options:
 
 ### calicoctl node 
 
-This command starts a container using the `calico/node` Docker image. 
+This command performs two actions:
+1. Initialize this host for calico by setting first-time config in etcd.
+2. Start the `calico/node` Docker container.
```
spikecurtis: (Optional) Start Calico services.  By default (or via `--runtime=docker`) the command does this by downloading the `calico/node` Docker image and running it in a container.

docs/calicoctl/node.md
```diff
@@ -71,7 +74,9 @@ Options:
 
 ### calicoctl node 
 
-This command starts a container using the `calico/node` Docker image. 
+This command performs two actions:
+1. Initialize this host for calico by setting first-time config in etcd.
```
spikecurtis: Calico

calico_containers/calico_ctl/status.py
djosborne: There's a lot of docker calls throughout `status` which gather info. Wrapping the whole thing in a try/except means that If one of the initial calls passes fine but the later ones do not, the "Docker is not running" message wouldn't accurately capture the real problem.

calico_containers/calico_ctl/node.py
```diff
@@ -536,24 +548,33 @@ def warn_if_hostname_conflict(ip):
...
+                    # Calico-node isn't running on this host.
+                    # There may be another host using this hostname.
+                    print_paragraph("WARNING: Hostname '%s' is already in use "
+                            "with IP address %s. Calico requires each compute host to "
+                            "have a unique hostname. If this is your first time "
+                            "running 'calicoctl node' on this host, ensure that "
+                            "another host is not already using the "
+                            "same hostname." % (hostname, ip))
+            except IOError:
+                # Couldn't connect to docker to confirm calico-node is running.
```
spikecurtis: I don't think we want to drop a warning if they are using `--runtime=none` here, or at least not this warning.

djosborne: Ideally we'd detect if calico-felix is running on the system here, and if it is not, then drop the warning. Instead of designing a way to detect if Felix is running or not, I figured it'd be safer to always warn for now.

calico_containers/calico_ctl/status.py
```diff
@@ -35,29 +36,34 @@ def status(arguments):
...
+
+            if result is not None:
+                print "Running felix version %s" % result.group(1)
+
+            print "\nIPv4 BGP status"
+            pprint_bird_protocols(4)
+            print "IPv6 BGP status"
+            pprint_bird_protocols(6)
+    except ConnectionError:
+        print "Docker is not running"
```
spikecurtis: Can we make this message meaningful to a user that doesn't want to use Docker?  Ditto with other messages.  The messages are factually true, but with multiple runtimes it isn't clear what the user should expect to see.

I'm wondering if we should write some data to an OS file to make `calicoctl` stateful and more helpful.  It would help to know answers to questions like

 - has an admin run `calicoctl node` on this host before?
 - what runtime did they use the last time they ran it?

We could even log all the `calicoctl` commands, which would be super helpful for debugging issues in the field.

djosborne: While I do agree with these suggestions, I don't believe they fall within the scope of the issue this PR is trying to address. My goal in this PR is only to allow users to launch a dockerless node. Improvements to `checksystem` and `statuts` to provide useful information when docker has launched in this way are definitely the next step, but require significant thought into how calicoctl is structured. I think it'd be best if we start that work in a new PR after a bit of designing

spikecurtis: Sounds good.  Raise an issue so we don't lose track of it.

calico_containers/calico_ctl/node.py
spikecurtis: OK.  I'd forgotten that this only drops if they re-run and change the IP address, so it's not like we'll drop it every time if they put it in a startup script or whatever.

calico_containers/tests/unit/node_test.py
```diff
@@ -166,6 +166,98 @@ def test_validate_arguments(self, case, sys_exit_called):
...
     @patch('calico_ctl.node._find_or_pull_node_image', autospec=True)
     @patch('calico_ctl.node._attach_and_stream', autospec=True)
+    def test_node_dockerless_start(self, m_attach_and_stream,
+                                   m_find_or_pull_node_image, m_docker,
+                                   m_docker_client, m_client, m_install_plugin,
+                                   m_warn_if_hostname_conflict, m_warn_if_unknown_ip,
+                                   m_get_host_ips, m_setup_ip, m_check_system,
+                                   m_os_makedirs, m_os_path_exists):
+        """
+        Test that the node_start function behaves as expected when
```
spikecurtis: "behaves as expected" is pretty meaningless in this context.  How about "node_start function does not make Docker calls when runtime=None"

spikecurtis:
One minor nit, otherwise LGTM.

calico_containers/calico_ctl/node.py
```diff
@@ -35,6 +36,11 @@
                             "false" to run in the foreground.  When using
                             libnetwork, this may not be set to "false".
                             [default: true]
+  --runtime=<RUNTIME>       Specify how calico services should be
```
tomdee: Calico?

calico_containers/calico_ctl/node.py
```diff
@@ -35,6 +36,11 @@
...
+                            launched.  When set to "docker", services will be
+                            launched via the calico-node container, whereas a
+                            value of "none" will not launch them
```
tomdee: funny line break

calico_containers/calico_ctl/node.py
```diff
@@ -536,24 +548,33 @@ def warn_if_hostname_conflict(ip):
...
+                            "same hostname." % (hostname, ip))
+            except IOError:
+                # Couldn't connect to docker to confirm calico-node is running.
                 print_paragraph("WARNING: Hostname '%s' is already in use "
-                                "with IP address %s. Calico requires each compute host to "
-                                "have a unique hostname. If this is your first time "
-                                "running 'calicoctl node' on this host, ensure that "
-                                "another host is not already using the "
-                                "same hostname." % (hostname, ip))
+                            "with IP address %s. Calico requires each compute host to "
```
tomdee: Why did you change the alignment here?

calico_containers/calico_ctl/node.py
djosborne: I didn't want [default:all] on its own line, so I broke early....

calico_containers/calico_ctl/node.py
tomdee: You should stick this big text block in a variable.

tomdee:
A couple of little tidyups but otherwise LGTM

PR #600, opened 2015-11-06
Check etcd is up and has the correct version

### LINKED ISSUE #479: calicoctl checksystem should examine etcd version and warn
There are some versions of etcd that have known problems.  As part of check-system we should check the running version of etcd and drop a warning if it isn't the version we recommend.

This means we would need to introduce the concept of a recommended version, do our testing with it, and be proactive about testing the new versions.

tomdee:
Fixes #479

### DIFF
--- a/calico_containers/calico_ctl/checksystem.py
+++ b/calico_containers/calico_ctl/checksystem.py
@@ -26,12 +26,14 @@ import sys
 import re
 
 import docker
+from etcd import EtcdConnectionFailed
 from requests import ConnectionError
 from subprocess32 import check_output
 
-from utils import DOCKER_VERSION, DOCKER_LIBNETWORK_VERSION, REQUIRED_MODULES
+from utils import DOCKER_VERSION, DOCKER_LIBNETWORK_VERSION, REQUIRED_MODULES, \
+    ETCD_VERSION
 from utils import enforce_root
-from connectors import docker_client
+from connectors import docker_client, client, etcd_authority
 
 
 def checksystem(arguments):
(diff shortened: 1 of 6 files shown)


calico_containers/calico_ctl/checksystem.py
```diff
@@ -177,3 +181,36 @@ def _check_docker_version(libnetwork=False):
...
+    try:
+        detected_version_string = \
+            client.etcd_client.api_execute("/version", 'GET').data
+    except EtcdConnectionFailed:
+        print >> sys.stderr, "ERROR: Could not connect to etcd at %s" % \
+            etcd_authority
+        system_ok = False
+    else:
+        # Remove everything except numbers and dots from the string.
+        detected_version_string = re.sub(r'[^\d\.]', '',
```
robbrockbank: This isn't normal semantic versioning is it?

Normally you'd have x.y.z with a possible extra bit at the end separated by a hyphen (e.g.1.0.0-alpha.1).

In the example above the tuple would end up as (1, 0, 1) which is incorrect.

robbrockbank: Better perhaps to stick to normal version format and include that in the normalize_version ?

calico_containers/calico_ctl/node.py
```diff
@@ -241,8 +241,12 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
 
     # Print warnings for any known system issues before continuing
     using_docker = True if runtime == 'docker' else False
```
robbrockbank: not you i realise ...  but are we really using a ternary operator to convert a boolean to a boolean?

calico_containers/calico_ctl/checksystem.py
robbrockbank: Actually htere is a good article here:  http://semver.org/

robbrockbank: I'd be happy if you just use the   X.Y.Z and ignore the bit at the end and use that rather than trying to be clever with the bit at the end.

robbrockbank:
Actually - despite my comments last night.  I'm ok with this.

LGTM.

calico_containers/calico_ctl/checksystem.py
```diff
@@ -177,3 +181,36 @@ def _check_docker_version(libnetwork=False):
             system_ok = False
 
     return system_ok
+
+def _check_etcd_version():
```
spikecurtis: I'd like to see UT of this function that mocks out `api_execute` calls with some results we know occur in the field.

calico_containers/calico_ctl/checksystem.py
robbrockbank: Actually, the more I think about this I'm not sure I do like the removing all other characters.

With Semver, the version could be something like 1.2.3-x1 which would get transposed to 1.2.31 with this algorithm.

I think we probably do need to at least handle the - or + separating the X.Y.Z from the pre-release identifier.

PR #605, opened 2015-11-06
Safer etcd version checking

tomdee:


robbrockbank:
LGTM, but as Spike pointed out, a UT of the version function might be good ;-)

PR #607, opened 2015-11-06
Add a simple IPv6 mainline test

tomdee:
It doesn't work Docker-in-Docker - that requires Docker to do some IPv6 which requires additional config.

This works as long as an interface on the host has an autoconfigured link local address.

### DIFF
--- a/calico_containers/tests/st/utils/utils.py
+++ b/calico_containers/tests/st/utils/utils.py
@@ -51,6 +51,7 @@ def get_ip(v6=False):
             ips = get_host_ips(version)
             if ips:
                 ip = ips[0]
+
     return ip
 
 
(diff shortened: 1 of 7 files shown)


calico_containers/tests/st/utils/utils.py
```diff
@@ -20,26 +20,37 @@
...
     """
-    try:
-        ip = os.environ[LOCAL_IP_ENV]
-    except KeyError:
-        # No env variable set; try to auto detect.
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(("8.8.8.8", 80))
-        ip = s.getsockname()[0]
-        s.close()
+    env = LOCAL_IPv6_ENV if v6 else LOCAL_IP_ENV
```
robbrockbank: I assume there is a minimum IPv6 setup required on your dev box for this to work.  It would be nice to raise an appropriate exception if it appears your dev box is not set up for IPv6 - so that there is a clear user action.  WDYT?

robbrockbank:
One comment, otherwise LGTM.

PR #612, opened 2015-11-11
Change ST framework to make it more compatible

tomdee:


calico_containers/tests/st/utils/workload.py
```diff
@@ -79,12 +79,16 @@ def _get_ping_function(self, ip):
...
-            ping = "ping6"
+        # Default to "ping"
+        ping = "ping"
+
+        try:
+            version = IPAddress(ip).version
+            assert version in [4, 6]
+            if version == 6:
+                ping = "ping6"
+        except BaseException:
```
robbrockbank: why do you need to catch a BaseException here?

calico_containers/tests/st/utils/workload.py
```diff
@@ -79,12 +79,16 @@ def _get_ping_function(self, ip):
...
-        else:  # if version == 6:
-            ping = "ping6"
+        # Default to "ping"
+        ping = "ping"
+
+        try:
+            version = IPAddress(ip).version
+            assert version in [4, 6]
+            if version == 6:
+                ping = "ping6"
```
robbrockbank: `ping = "ping" if version ==4 else "ping6"` might be nicer than defining it and overwriting it

calico_containers/tests/st/utils/workload.py
tomdee: Because I want to ping a hostname.

robbrockbank:
Minor comments.  Not sure I prefer this approach of passing in an options string, but understand why it might be useful so LGTM if you have a preference to do it this way,

PR #616, opened 2015-11-12
Reorganize calico-node

djosborne:
Split calico-node into common and docker/runit-specific build directories 

Moved the directory structure around within the calico-node container to run confd/bird out of /etc/calico instead of the root filepath


Makefile
```diff
@@ -52,9 +52,10 @@ calico_test/.calico_test.created: $(TEST_CONTAINER_FILES)
 	cd calico_test && docker build -t calico/test:latest .
 	touch calico_test/.calico_test.created
 
-calico_node/.calico_node.created: $(NODE_CONTAINER_FILES)
-	cd calico_node && docker build -t calico/node:latest .
-	touch calico_node/.calico_node.created
+calico_node/docker_dist/.calico_node.created: $(NODE_CONTAINER_FILES)
+	cd calico_node && docker build -f docker_dist/Dockerfile -t calico/node:latest .
```
tomdee: I don't think this is compatible with Docker Hub. You can pass in a different path for a Dockerfile but that becomes the context root.

tomdee:
This rename is wrong
calico_node/filesystem/config/.gitkeep → calico_node/common/confd/config/.gitkeep

tomdee:
Can you explain why you're splitting into common and docker_dist?

djosborne:
The reason for the common / docker_dist split is really the reason for this whole PR: all calico-node distributions (docker and future baremetal / rpm) share the same confd templates, while also having distro-specific files

Makefile
djosborne: I thought the final argument for `docker build` (`.` ) would be the context root

PR #617, opened 2015-11-12
Bump kubernetes plugin version to v0.6.0

caseydavenport:


TrimBiggs:
LGTM, any idea what the failures are from?

calico_containers/calico_ctl/node.py
```diff
@@ -53,7 +53,7 @@
   --kubernetes              Download and install the Kubernetes plugin.
   --kube-plugin-version=<KUBE_PLUGIN_VERSION> Version of the Kubernetes plugin
                             to install when using the --kubernetes option.
-                            [default: v0.3.0]
+                            [default: v0.5.0]
```
robbrockbank: You'll need to update the node.md docs file too.

robbrockbank:
Probably need to rebase to get the tests working.

robbrockbank:
In the master code we pin calico/node to "latest".  Can we do the same for the kubernetes plug in?

IIRC, the plugin is an artifact from a CircleCI build.  If that is the case, then we do have a mechanism for pulling the latest artifacts from a particular branch of a CircleCI build.  If you think this would be useful, let's have a chat.

caseydavenport:
The plugin is just pulled from GitHub releases - do you know if there is a way to do latest on GitHub releases?

caseydavenport:
We don't really have a way in place to make sure these things are compatible. e.g if calico-kubernetes uses a different libcalico data-model than calico-docker, etc.  I don't think pinning it to `latest` will be any worse than it currently is, but it does scare me a tiny little bit...

PR #622, opened 2015-11-13
Log the command output

robbrockbank:


robbrockbank:
Tom, any objections to this?  I think it's quite nice having the output of the commands logged as well as the commands themselves.  I've tried to make it clear what is command ouput by indenting slightly.  Obviously this produces more output and I realize that you don't get a simple linear list of commands, but quite useful nonetheless.

tomdee:
What about using a # as a the separator (to allow copy-paste of commands)? Otherwise, LGTM

robbrockbank:
Yup that's great idea!  Will change to `#`

PR #623, opened 2015-11-13
Should not remove last character from line output

robbrockbank:


TrimBiggs:
LGTM
Do we know why it was originally written like this?

robbrockbank:
Yes I'd just written it and assumed the split included the new line.  Wasn't thinking and didn't notice in the output until just now.

PR #624, opened 2015-11-13
Add mtu configuration when creating veth, especially for ip-in-ip

### LINKED ISSUE #610: MTU within container not set to be equal to tunnel MTU
From titilambert:

Hello, I got a problem with MTU. MTU of container cali interface should be less or equal to the MTU of tunl0 interface. (I think the default is 1440)
Also, It should be cool, to be able to define the MTU when we create docker interface.
```
docker network create --driver=calico --opt nat-outgoing=true --opt ipip=true --mtu=1440 --subnet=192.168.1.0/24
```
Thanks !

TrimBiggs:
Fixes #610

### DIFF
--- a/calicoctl/calico_ctl/container.py
+++ b/calicoctl/calico_ctl/container.py
@@ -45,6 +45,11 @@ from utils import validate_cidr
 from utils import validate_ip
 
 
+# The default veth MTU for containers.
+DEFAULT_VETH_MTU = 1500
+DEFAULT_IPIP_VETH_MTU = 1440
+
+
 def assign_any(v4_count, v6_count, pool=(None, None)):
     """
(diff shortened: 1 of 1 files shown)


TrimBiggs:
This isn't how we should fix the issue.

dudymas:
Can there be more of a description to why this PR was closed? I don't see why this solution is bad.

robbrockbank:
@TrimBiggs Can you add some comments as to why this one was closed.  Ta.

TrimBiggs:
This was originally closed because the backend design had not been fleshed out, so this piece did not make sense to go in without any configuration change being made to the containers.  We are working on designing a solution to this problem now.

dudymas:
thanks for the feedback. I'm mostly affected by this once I start bridging edge nodes in Mantl.io

PR #628, opened 2015-11-16
Correct typo in restart policy

### LINKED ISSUE #626: calico docker container are not restarted automatically, when docker restart
I think we want to set the calico-node, calico-libnetwork container to be restart automatically.
With docker 1.9, there are not restarted auto when docker restart.

I found the settings in the code 

https://github.com/projectcalico/calico-docker/blob/master/calicoctl/calico_ctl/node.py

```python
    host_config = docker.utils.create_host_config(
        privileged=True,
        restart_policy={"Name": "Always"},
        network_mode="host",
        binds=binds)
```

Should it be "always" instead of "Always"? According to 
https://github.com/docker/docker/blob/13963957ebadc663915aed8c61451a93db41c27c/daemon/container.go#L509-L513

tomdee:
Fixes #626

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -370,7 +370,7 @@ def _start_node_container(ip, ip6, etcd_authority, log_dir, node_image, detach):
 
     host_config = docker.utils.create_host_config(
         privileged=True,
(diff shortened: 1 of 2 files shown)


robbrockbank:
LGTM once tests pass

PR #632, opened 2015-11-17
Ensure that version in the Makefile are updated so tests pass

tomdee:


robbrockbank:
Do we have versions of calico/test too?  Or should that always be latest?

If the latter, then LGTM.

PR #635, opened 2015-11-18
Move validation function to libcalico

L-MA:
This is part of https://github.com/projectcalico/libcalico/pull/53


L-MA:
With these changes, some of the Rule arguments are being validated twice: once in the calicoctl's `profile.py` and once in libcalico when the Rule is created. The functions used in validation live in libcalico `util.py`.

The reason validation is repeated is to ensure good error messages are being displayed when bad arguments are given to calicoctl.

robbrockbank:
LGTM

PR #644, opened 2015-11-20
Update etcd parms to limit allowable IPs

robbrockbank:
So I think including localhost in advertise client urls is not actually correct - although in our STs there won't be any side effect of it.

Also it's good to close down the available IPs clients can access the server with.

With this change, it is possible to connect using local host still.

As per etcd docs:

> etcd listens on listen-client-urls to accept client traffic. etcd member advertises the URLs specified in advertise-client-urls to other members, proxies, clients. Please make sure the advertise-client-urls are reachable from intended clients. A common mistake is setting advertise-client-urls to localhost or leave it as default when you want the remote clients to reach etcd.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,8 @@ run-etcd:
 	docker run --detach \
 	--net=host \
 	--name calico-etcd quay.io/coreos/etcd:v2.0.11 \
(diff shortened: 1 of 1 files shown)


tomdee:
Will I still be able to use `etcdctl ls` with no additional options after this change? If so, LGTM.

robbrockbank:
Yup etcdctl still works using the default localhost

Issue #512, opened 2015-10-02
Support HTTPS in etcd URL schemes


loshkovskyi:
Right now it's not possible to use Calico with etcd cluster where all URLs use HTTPS. It would be great to have such option for security implications. The same is for Calico Kubernetes plugin, as everything I can do is to change the `ETCD_AUTHORITY` variable which is not enough to access the etcd store.

lxpollitt:
+1

tomdee:
We should prioritise fixing this. The code should be easy but we need to agree on a design. Any input would be appreciated.

One option would be to use another environment variable such as `ETCD_SCHEME` which defaults to `http` but can also take the value `https`

@loshkovskyi What are your thoughts?

loshkovskyi:
Here is an example how I can access etcd store over HTTPS:
```
etcdctl --cert-file /etc/origin/master/master.etcd-client.crt \
--key-file /etc/origin/master/master.etcd-client.key \
--ca-file /etc/origin/master/master.etcd-ca.crt \
--peers https://osv3-master-01.example.com:2379,https://osv3-node-01.example.com:2379,https://osv3-node-02.example.com:2379 \
ls
```
So if I set `ETCD_AUTHORITY` to `https`, I need several more variables to set paths to the certificates.

robbrockbank:
I've been looking into a design for this.  Please touch base with me before tackling this one.

TrimBiggs:
Fixed by #654

PR #654, opened 2015-11-26
Add support for secure etcd

TrimBiggs:


euprogramador:
there is some predicting this feature available this? I am implementing a ETCD production, which uses client certificate for authentication, and need this feature, you know how it could make the connection with the ETCD via calicoctl in ETCD with client authentication?

TrimBiggs:
@euprogramador We are hoping to get this feature in this week.  This will include documentation on how to configure parameters for your calico node using the CA certificate for server authentication and optional client-key and client-certificate for client authentication.

Makefile
```diff
@@ -93,13 +98,38 @@ ut-circle: calico_test/.calico_test.created dist/calicoctl
...
-	@-docker rm -f calico-etcd
+	@-docker rm -f calico-etcd calico-etcd-ssl
 	docker run --detach \
 	--net=host \
 	--name calico-etcd quay.io/coreos/etcd:v2.0.11 \
 	--advertise-client-urls "http://$(LOCAL_IP_ENV):2379,http://127.0.0.1:2379" \
 	--listen-client-urls "http://0.0.0.0:2379"
 
+## Run etcd with SSL in a container using full certificate/key/CA certificate verification. Used primarily by STs.
+run-etcd-ssl-all: generate-certs
```
robbrockbank: I think we can drop the "all"

Makefile
```diff
@@ -119,6 +149,20 @@ st: run-etcd dist/calicoctl docker calico_test/.calico_test.created busybox.tar
...
+	           -e HOST_CHECKOUT_DIR=$(HOST_CHECKOUT_DIR) \
+	           -e ETCD_SECURITY_TYPE=ssl-all \
+	           --rm -ti \
+	           -v /var/run/docker.sock:/var/run/docker.sock \
+	           -v `pwd`:/code \
+	           calico/test \
+	           sh -c 'cp -ra tests/st/* /tests/st && cd / && nosetests $(ST_TO_RUN) -sv --nologcapture --with-timer $(ST_OPTIONS)'
+
 semaphore:
 	# Clean up unwanted files to free disk space.
```
robbrockbank: Need to update the semaphore build to include a subset of the secure STs

Makefile
```diff
@@ -119,6 +148,22 @@ st: run-etcd dist/calicoctl docker calico_test/.calico_test.created busybox.tar
...
 
+## Run the STs in a container using etcd with SSL certificate/key/CA verification.
+st-ssl: run-etcd-ssl dist/calicoctl docker calico_test/.calico_test.created busybox.tar calico-node.tar routereflector.tar
+	docker run --uts=host \
+	           --pid=host \
+	           --net=host \
+	           --privileged \
+	           -e HOST_CHECKOUT_DIR=$(HOST_CHECKOUT_DIR) \
+	           -e ETCD_SECURITY_TYPE=ssl \
+	           -e HOST_CERT_DIR=`pwd`/certs \
```
robbrockbank: Why do you want to use the same directory hierarchy in the container?  (i.e. why the 'pwd' here?)

calico_test/tests/st/test_base.py
```diff
@@ -38,9 +41,17 @@ def setUp(self):
 
         # Delete /calico if it exists. This ensures each test has an empty data
         # store at start of day.
-        subprocess.check_output(
-            "curl -sL http://%s:2379/v2/keys/calico?recursive=true -XDELETE"
-            % self.ip, shell=True)
+        etcd_ssl_type = os.getenv("ETCD_SECURITY_TYPE", None)
+        if etcd_ssl_type == "ssl":
```
robbrockbank: Is it ssl or tls?

Actually do we need this extra parameter, can't we base it off whether we have a host cert directory?  Not too fussed about this though.

calico_test/tests/st/utils/docker_host.py
```diff
@@ -95,11 +96,29 @@ def calicoctl(self, command):
...
         # Export the environment, in case the command has multiple parts, e.g.
         # use of | or ;
         calicoctl = "export %s; %s" % (etcd_auth, calicoctl)
+        if etcd_ssl_type == "ssl":
+            if self.dind:
+                etcd_ca = ETCD_CA_FILE
+                etcd_cert = ETCD_CERT_FILE
+                etcd_key = ETCD_KEY_FILE
+            else:
+                cert_path = os.getenv("HOST_CERT_DIR")
```
robbrockbank: Could you not have passed a HOST_CERT_DIR through to the DinD container that allows you to just have a single branch of code?

calico_test/tests/st/test_base.py
```diff
@@ -38,9 +41,17 @@ def setUp(self):
...
+            subprocess.check_output(
+                "curl --cacert %s --cert %s --key %s "
```
robbrockbank: Do we not have to worry about DinD here?

calico_test/tests/st/utils/docker_host.py
```diff
@@ -95,11 +96,29 @@ def calicoctl(self, command):
         :return: The output from the command with leading and trailing
         whitespace removed.
         """
+        etcd_ssl_type = os.environ.get("ETCD_SECURITY_TYPE", None)
         calicoctl = os.environ.get("CALICOCTL", "/code/dist/calicoctl")
+
         etcd_auth = "ETCD_AUTHORITY=%s:2379" % get_ip()
         # Export the environment, in case the command has multiple parts, e.g.
         # use of | or ;
         calicoctl = "export %s; %s" % (etcd_auth, calicoctl)
```
robbrockbank: This line should realy be in an else branch.

calico_test/tests/st/test_base.py
TrimBiggs: SSL and TLS are used somewhat interchangeable nowadays.  SSL is the more common name, and usually describes either.

We can't base it off of having the certs dir because the certs dir is not cleaned up after an st-ssl or after run-etcd-ssl.

Makefile
TrimBiggs: This is a bit funny.  Both this HOST_CERT_DIR and the volume are required for non-DinD tests.

When using DinD, the Docker container hosts have `calico-docker` mounted as  `/code`. Passing the etcd certs to the `calicoctl node` command works fine because we know we can set the etcd environment variable paths from `/code/certs`.

However, when not using DinD, there are two things that happen:
1) Volume mounts during `docker run` will copy directories from the host since the st container uses the hosts docker socket.
2) The `calicoctl node` command is run from the st container, which checks the existence of the etcd certs on the machine that run the command (which is the st container).  This fails without mounting the hosts's certs directory as the same path in the test-container.

The `HOST_CERT_DIR` environment variable is needed to have a reference to the certs location on the host when mounting the cert files in calicoctl node.

calico_test/tests/st/utils/docker_host.py
TrimBiggs: Why is that? It is used in all cases.

calico_test/tests/st/utils/docker_host.py
TrimBiggs: Yes, it can.

calico_test/tests/st/test_base.py
TrimBiggs: Nope, this is always called from within the st test container.

Makefile
robbrockbank: Hmm, doesn't that mean you don't need the ```-v `pwd`/certs:`pwd`/certs``` directive at all?

calico_test/tests/st/test_base.py
robbrockbank: I meant if we just pass in the CERTs directory environment variable

calico_test/tests/st/utils/docker_host.py
```diff
@@ -95,11 +96,29 @@ def calicoctl(self, command):
...
+        if etcd_ssl_type == "ssl":
+            if self.dind:
```
robbrockbank: Need to comment all this stuff.

Makefile
TrimBiggs: You do need the files in the st test container because of 2) in the comment above.  If the env value filepaths do not exist on the test container then the `calicoctl node` command will fail, saying that it cannot read the files.

calico_test/tests/st/test_base.py
TrimBiggs: Ah, yes.  Now that the HOST_CERT variable is passed in I can remove the ETCD_SECURITY_TYPE variable.

Makefile
robbrockbank: Oh yes yes yes I gotcha.  calicoctl is not only invoking docker, but it also performs some file consistency checks locally, so there is the requirement to have the same file locations inside the container and on the main docker host (i,.e. the actual host).

That's irritating.

calico_test/tests/st/utils/docker_host.py
robbrockbank: Yup scrap that - I didn't see you were using string stuff to concat it

Makefile
TrimBiggs: Done below.

calico_test/tests/st/utils/route_reflector.py
```diff
@@ -47,21 +48,41 @@ def __enter__(self):
...
-                           r'-XPUT -d value="{'
-                             r'\"ip\":\"%s\",'
-                             r'\"cluster_id\":\"%s\"'
-                           r'}"' % (get_ip(), rr.ip, rr.ip, cluster_id))
+                if os.getenv("ETCD_SCHEME", None) == "https":
+                    # Etcd is running with SSL/TLS, pass the key values
+                    etcd_ca = "-e ETCD_CA_CERT_FILE=%s" % ETCD_CA
+                    etcd_cert = "-e ETCD_CERT_FILE=%s" % ETCD_CERT
+                    etcd_key = "-e ETCD_KEY_FILE=%s" % ETCD_KEY
+                    rr.execute("docker run --privileged --net=host -d "
```
robbrockbank: Side note:  You also need to update the README.md doc in calico-bird that describes the RR.

calicoctl/calicoctl.py
```diff
@@ -57,6 +57,10 @@
 import calico_ctl.ipam
 from calico_ctl.utils import print_paragraph
 
+import logging
+
+# Suppress warnings from urllib3 when etcd is used without CA cert.
```
robbrockbank: I guess we don't need this anymore?

calico_test/tests/st/utils/docker_host.py
```diff
@@ -95,11 +96,29 @@ def calicoctl(self, command):
...
+                etcd_ca = ETCD_CA_FILE
```
robbrockbank: Did it not make sense to volume mount the same full directory in DinD as on the host so we could use the same path names in the DinD and non-DinD cases?

robbrockbank:
Tim, some minor comments and questions in the PR.

robbrockbank:
Also, looks like you need to rebase

docs/EtcdSecureCluster.md
```diff
@@ -0,0 +1,75 @@
+<!--- master only -->
+> ![warning](../../images/warning.png) This document applies to the HEAD of the calico-docker source tree.
```
robbrockbank: Link to image is not correct

robbrockbank:
If you're only making minor updates then feel free to checkin.

TrimBiggs:
Fixes #512

euprogramador:
@TrimBiggs I am using senginx to make a proxy bind local address

PR #661, opened 2015-12-03
Release script improvements.

tomdee:
- Add versions to __init__.py
- Add dry-run option
- Allow all git command to be run automatically
- Allow the cwd to be different to the script directory

### DIFF
--- a/release-scripts/do_release.py
+++ b/release-scripts/do_release.py
@@ -2,23 +2,26 @@
 """do_release.py
 
 Usage:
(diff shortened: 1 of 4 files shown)


release-scripts/do_release.py
```diff
@@ -244,9 +247,9 @@ def update_master():
     actions()
```
robbrockbank: We should be updating the __init__ versions for the other libraries in master.

Also, I don't see that we are updating the __init__ versions for the candidate branch either.

PR #675, opened 2015-12-11
Add validation check before changing bird config

TrimBiggs:


TrimBiggs:
This is a partial fix for https://github.com/projectcalico/calico-bird/issues/12, preventing BIRD from updating its config if the updated config will cause bird to fail.

robbrockbank:
LGTM

PR #677, opened 2015-12-14
Tweak print_paragraph to support a file parameter

alexwlchan:
This change allows messages from print_paragraph to be directed to (for example) sys.stderr, which is useful for things like error messages.

### DIFF
--- a/calicoctl/calico_ctl/utils.py
+++ b/calicoctl/calico_ctl/utils.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import print_function
+
 import socket
 import os
 import sys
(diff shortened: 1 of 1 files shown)


calicoctl/calico_ctl/utils.py
```diff
@@ -38,18 +40,19 @@ def enforce_root():
...
     """
     Print a fixed width (80 chars) paragraph of text.
     :param msg: The msg to print.
+    :param file: The text stream to write to (default sys.stdout)
     :return: None.
     """
-    print "\n".join(textwrap.wrap(msg, width=80))
-    print
+    print("\n".join(textwrap.wrap(msg, width=80)), file=file)
+    print("")
```
robbrockbank: strictly - shouldn't this be output to file too?

alexwlchan: D'oh. Fixed.

robbrockbank:
lgtm

PR #680, opened 2015-12-14
Use debian:jessie to build calicoctl for python 2.7.9

TrimBiggs:
Switch from using `debian:wheezy` to `debian:jessie` when building the calicoctl binary. This fixes a potential security issue when running etcd with SSL using urlib3 with Python 2.7.3. See [this page](https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning) from urllib3 for more details.

`wheezy` uses Python 2.7.3, whereas `jessie` uses 2.7.9.


TrimBiggs:
Closing for now.  We may want to have urllib3 use OpenSSL instead.

PR #687, opened 2015-12-15
Update release scripts to be safer and use correct master versions

TrimBiggs:


TrimBiggs:
I've rearranged the script a bit to be more clear about what the user actions should be verses what the script does for the user.

I've also forced the user to make the push to update master so that the user must review the changes before they go in.

release-scripts/do_release.py
```diff
@@ -207,17 +223,20 @@ def cut_release():
     new_version = release_data["versions"]["version"]
     para("The codebase has been updated to reference the GitHub release "
          "artifacts.")
-    actions()
-    bullet("Add, commit and push the updated files to "
+    bullet("Adding, committing and pushing the updated files to "
```
robbrockbank: What about if the --dry-run flag is set?

release-scripts/do_release.py
```diff
@@ -91,6 +91,18 @@
     (re.compile(r'__version__\s*=\s*".*"'),
      '__version__ = "{version-no-v}-dev"'),
 
+    (re.compile(r'__libnetwork_plugin_version__\s*=\s*".*"'),
```
robbrockbank: Not sure it's quite this simple.  At the moment we've broken the old behavior of having master calico-docker point at latest libnetwork-plugin and latest calico/node.  But can't simply set to latest because it will break the behavior of defaulting the versions to the versions in the last release.

release-scripts/do_release.py
TrimBiggs: This will require a change to lines 173-181 above also, which already existed like this.

TrimBiggs: Actually, isn't this what we would want? Isn't the dry-run flag meant to be for changes like this?

release-scripts/do_release.py
TrimBiggs: I don't understand what you mean by your last sentence. Do we want to have the versions from the last (previous) release? I thought we wanted to have the versions of the most recent release.

These versions are only used in calicoctl/__init__.py.  This does not update any other version strings in the project.  The values from this file that are used outside of the file are the calicoctl version (in the same-named command) and the kubernetes/rkt versions to set the default plugin version to pull.

Should we be adding `-dev` to the names of the libnetwork, libcalico, and felix version strings to show that it is not bound to a release?

robbrockbank: You are quite right - I'd mis-remembered how the libnetwork and calico-node Docker images were specified in the doc string - I *thought* that it uses the versions in the __init__ file, but actually for those two it doesn't - they are hard coded in master to be  "latest".

Yeah I was thinking -dev might be sensible too.   Just need to make sure we handle that correctly when reading it in at start of day.

release-scripts/do_release.py
robbrockbank: Yeah ignore me again!

robbrockbank:
LGTM

PR #693, opened 2015-12-17
Added Mesos-Marathon documentation

djosborne:


docs/mesos/README.md
```diff
@@ -46,3 +46,17 @@ a.)  We've bundled Mesos with netmodules into a convenient RPM - follow the [Mes
...
+```
+$ docker run \
+-e MARATHON_MASTER=zk://<ZOOKEEPER-IP>:2181/mesos \
+-e MARATHON_ZK=zk://<ZOOKEEPER-IP>:2181/marathon \
+-p 8080:8080 \
+djosborne/marathon:ip_per_task
+```
+The Marathon UI has does not yet include a field for specifiying NetworkInfo, so we'll use the command line to send Marathon an App's JSON instead:
+```
+$ curl -X PUT -H "Content-Type: application/json" http://localhost:8080/v2/groups/calico-apps  -d '{"id":"/calico-apps", "apps": [{"id": "hello-world-1", "cmd": "ifconfig", "cpus": 0.1, "mem": 64.0, "ipAddress": {"groups": ["netgroup-1"]}}]}'
```
spikecurtis: Can you pretty-print this blob of JSON and explicitly explain the new `ipAddress` field?

PR #696, opened 2015-12-22
Auto assign does not immediately reassign the same IPs

robbrockbank:


TrimBiggs:
LGTM, this fixes some of the STs, the other failures are do to an unrelated issue.

PR #705, opened 2015-12-31
pool should always come from client

tonicbupt:
may fix [this](https://github.com/projectcalico/libcalico/issues/68)

when comparing two IPPools, values should be specified, if create one IPPool, `masquerade`, `ipip`, `ipam` may not be the same as stored in etcd.

### DIFF
--- a/calicoctl/calico_ctl/container.py
+++ b/calicoctl/calico_ctl/container.py
@@ -32,7 +32,7 @@ from subprocess import CalledProcessError
 from netaddr import IPAddress, IPNetwork
 from calico_ctl import endpoint
 from pycalico import netns
-from pycalico.datastore_datatypes import IPPool, Endpoint
+from pycalico.datastore_datatypes import Endpoint
 from pycalico.ipam import AlreadyAssignedError
 
 from connectors import client
--- a/calicoctl/calico_ctl/container.py
+++ b/calicoctl/calico_ctl/container.py
@@ -507,7 +507,7 @@ def get_ip_and_pool(ip):
         pool = get_pool_or_exit(ip)
     elif ip is not None and '/' in ip:
         cidr = IPNetwork(ip)
-        pool = get_pool_or_exit(IPAddress(cidr.first))
+        pool = get_pool_by_cidr_or_exit(cidr)
         if IPNetwork(ip).version == 4:
             result = assign_any(1, 0, pool=(pool, None))
             ip = result[0][0]
@@ -529,6 +529,15 @@ def get_ip_and_pool(ip):
             sys.exit(1)
 
     return (ip, pool)
+
+
+def get_pool_by_cidr_or_exit(cidr):
+    pools = client.get_ip_pools(cidr.version)
+    for pool in pools:
+        if pool.cidr == cidr:
+            return pool
+    print "%s is not found" % cidr
+    sys.exit(1)
 
 
 def get_pool_or_exit(ip):
--- a/calicoctl/calico_ctl/container.py
+++ b/calicoctl/calico_ctl/container.py
@@ -507,7 +507,7 @@ def get_ip_and_pool(ip):
         pool = get_pool_or_exit(ip)
     elif ip is not None and '/' in ip:
         cidr = IPNetwork(ip)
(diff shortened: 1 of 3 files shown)


calicoctl/tests/unit/utils_test.py
```diff
@@ -50,8 +50,6 @@ def test_validate_cidr(self, cidr, expected_result):
         ('aa:bb::ff', 4, False),
         ('aa:bb::ff', 6, True),
         ('1111:2222:3333:4444:5555:6666:7777:8888', 6, True),
-        ('4294967295', 4, True),
-        ('5000000000', 4, False)
```
tonicbupt: this may be bug from `netaddr`, anyway just ignore this.

spikecurtis:
@tonicbupt Thanks for the PR!  In order to review and accept the fix, we would like you to sign a contributor's agreement, basically certifying that this code isn't owned by someone who might not want it in Calico.

http://docs.projectcalico.org/en/latest/contribute.html#contributor-agreements

tonicbupt:
@spikecurtis done

calicoctl/calico_ctl/container.py
```diff
@@ -506,7 +506,8 @@ def get_ip_and_pool(ip):
             ip = result[1][0]
         pool = get_pool_or_exit(ip)
     elif ip is not None and '/' in ip:
-        pool = IPPool(ip)
+        cidr = IPNetwork(ip)
+        pool = client.get_pool(IPAddress(cidr.first))
```
robbrockbank: If you specify a CIDR that does not exist, the pool returned will be None.  I think we actually want to just exit here if the pool does not exit.

I would replace client.get_pool() with the get_pool_or_exit() which will exit with an error if the pool does not exist.

tonicbupt: Yes I just checked that function, will change it here.

tonicbupt:
@robbrockbank ping

robbrockbank:
@tonicbupt Looking into this now.

calicoctl/calico_ctl/container.py
```diff
@@ -506,7 +506,8 @@ def get_ip_and_pool(ip):
             ip = result[1][0]
         pool = get_pool_or_exit(ip)
     elif ip is not None and '/' in ip:
-        pool = IPPool(ip)
+        cidr = IPNetwork(ip)
+        pool = get_pool_or_exit(IPAddress(cidr.first))
```
robbrockbank: Sorry I didn't spot this before.  Checking for just the first IP address is not sufficient here.  The API mentions that the CIDR must _exactly_ match the CIDR of a pool - so I think it is better to write another function to query the pools and compare the CIDR with that of the pool.  We so something similar in our libnetwork plugin (see below).  Could either update get_pool_or_exit() to accept a CIDR and do an exact match, or implement a separate method just for handling getting a pool from a CIDR (and exiting if it doesn't exist).

```
def get_pool(pool_cidr):
    """
    Return the Calico IP Pool with the specified pool CIDR, or None if not
    present.
    :param pool_cidr:  The pool CIDR.
    :return:  IPPool - The Calico pool, or None if not configured.
    """
    pools = client.get_ip_pools(pool_cidr.version, ipam=True)
    for pool in pools:
        if pool.cidr == pool_cidr:
            return pool
    return None
```

See https://github.com/projectcalico/libnetwork-plugin/blob/master/libnetwork/driver_plugin.py

tonicbupt: It's only because `client.get_pool` accepts only `IPAddress` object, so just use the first IP of the CIDR will do the same effect only if there's no intersection for any of each two CIDRs. If they do, won't anything wrong happen when use `client.get_pool`... ? Anyway I will change this as you said...

tonicbupt: And does this mean that `__eq__` method of `IPPool` object can only compare `cidr` of each other and just ignore `ipam`, `masquerade` ... If so, that would be a better solution

robbrockbank:
LGTM.  Thanks for the PR.  Merging now.

PR #706, opened 2016-01-04
docker start error - continue vagrant install

hernad:
vagrant install stops if service docker start unsuccessful


robbrockbank:
Hi @hernad - do you have a particular scenario where docker is failing to start?  Does this change just hide the fact that docker is failing?

docs/calico-with-docker/docker-network-plugin/vagrant-ubuntu/Vagrantfile
```diff
@@ -70,7 +70,7 @@ Vagrant.configure(2) do |config|
       host.vm.provision :shell, inline: %Q|sudo sh -c 'echo "DOCKER_OPTS=\"--cluster-store=etcd://#{primary_ip}:2379\"" > /etc/default/docker'|
 
       # Restart docker.
-      host.vm.provision :shell, inline: "sudo start docker"
+      host.vm.provision :shell, inline: "sudo start docker || true"
```
robbrockbank: What is the fail case you were hitting here?  I'm just wondering why the || true is necessary?

robbrockbank:
Hi @hernad 

Thanks for the PR! In order to accept the fix, we would like you to sign a contributor's agreement, basically certifying that this code isn't owned by someone who might not want it in Calico.

http://docs.projectcalico.org/en/latest/contribute.html#contributor-agreements

Would it be possible to sign that.  Also - I'm still curious as to the failure case that you were working around with the "|| true".

Thanks,

Rob

hernad:
Major refactor vagrant/ubuntu:

1) making one shell script (faster, easier for deployment)

2) provision idempotency

hernad:
Hi @robbrockbank,

> do you have a particular scenario where docker is failing to start?

The problem manifested after additional vagrant provisioning process (e.g. vagrant provision calico-01)
But this is outdated now, after major refactor I submitted.

I thing this version is much better, especially if you need  additional provisioning of virtual machines.

hernad:
I have signed contribution agreement.
Regards,
Ernad

TrimBiggs:
Hi @hernad.

We're quite busy at the moment, but I should have time to take a look at your PR later this week.

In the meantime, could you rebase your branch off of `projectcalico:master`? I believe the tests are failing on this PR because something is out of sync.

hernad:
resync with upstream

TrimBiggs:
Hi @hernad,

Thanks for syncing with master. I had some time to look into your PR changes, and I have a few comments.

I am still not quite sure what this change was fixing originally, and it is a bit harder to see now with the shell changes. Can you provide a bit more detail?  I think it would make sense for the initial change for docker failing to be the focus of this PR, since the shell script changes aren't really related to the the original issue.  I am very interested in understanding this issue, and I want to make sure we get in a fix for you and anyone else who may hit the problem.

Considering the shell script changes, I tested running both your script and the original vagrant script locally.  I actually found that the original script was faster (~26 min) compared to the script using the single shell command (~35 min).  This is a fairly significant difference.

Also, having the vagrant file as a shell script makes debugging a bit more difficult.  With native vagrant calls, it is easy to see the logical progression of the virtual machine creation.  You can debug errors by checking line number when a failure occurs, rather than needing to search through the file for the specific shell line.  It is useful to see the vagrant commands being run and displayed on the command line as well, especially for warnings and error messages, which you don't see with the shell script.

With these thoughts in mind, I'd suggest reverting this PR to your previous change and providing a bit more detail about what problem this PR fixes.  We definitely want to fix up the initial problem you were hitting if it is causing an issue.

Please let me know your thoughts.  We really appreciate your interest in using and improving Calico :)

TrimBiggs:
Hi @hernad,

Just wanted to check in to see if you've read my comment.  Please let me know your thoughts!

hernad:
Hi @TrimBiggs, I have read your comment, but I am working on other projects so this topic is staled  in my job list, sorry :( 


I didn't answer immediately because I wanted to make some speed tests related to your observations, but I couldn't provide that time

Anyway, these are some of my thoughts:

1) The speed issue

In my usage case, download of dependencies takes time (ADSL connection, github 300-500 KB/s) .
If I want to restart provision process, it takes significant time. In my version, there are checks (if docker already installed, if etcd already installed ) which makes this process much more efficient.

2) Debugging

You are right. Anyway  I like shell scripts, so my debug procedure is :) :

a) vagrant ssh
b) analyze generated shell script  into the session (/tmp/vagrant).

The shell script can be more verbose to provide current status of operations to the developer.

TrimBiggs:
Hi @hernad,

Thank you for the details and info about your process and usage of the shell vagrant script.

At this time, I'm going to close this PR because of the reasons I mentioned in my comment above, and also since the rest of our vagrant scripts use native ruby calls to vagrant as well.  It would be a pretty big change to modify each vagrant script, and I don't think it makes sense to have a single shell vagrant script.

I would still like to fix the initial issue you hit with `service docker start` failing, if you continue to see this.  If you have a reliable repro that you can list steps for, that would be great.

Thanks again for your contribution, we really appreciate the time you took to make Calico better.

PR #707, opened 2016-01-04
Add ICMPv6 support to calicoctl profile rules

### LINKED ISSUE #691: calicoctl can't match on ICMPv6
As described in https://github.com/projectcalico/calico/issues/938, inserting an ICMP rule for v6 addresses doesn't work.  

calicoctl/libcalico uses "icmp" for the protocol but felix is expecting "icmpv6".  This passes validation in Felix and the kernel accepts it even for an IPv6 rule, but it doesn't match the right packets.

We could fix in Felix by mapping "icmp" to different kernel protocols depending on whether we're generating an IPv4 or IPv6 rule.  However, since ICMP and ICMPv6 are quite different, the other parameters on the rule might be wrong too so maybe we should fix in the UI and force people to be explicit?

By other parameters, I mean the type and code, which don't line up.  e.g. ping is 7 in ICMP but it's 128 in ICMPv6.

TrimBiggs:
Fixes #691

### DIFF
--- a/calicoctl/calico_ctl/profile.py
+++ b/calicoctl/calico_ctl/profile.py
@@ -25,6 +25,9 @@ Usage:
       icmp [(type <ICMPTYPE> [(code <ICMPCODE>)])]
            [(from [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
            [(to   [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])] |
+      icmpv6 [(type <ICMPTYPE> [(code <ICMPCODE>)])]
+             [(from [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
+             [(to   [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])] |
       [(from [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
       [(to   [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])]
     )]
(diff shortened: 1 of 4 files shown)


calicoctl/calico_ctl/profile.py
```diff
@@ -203,6 +218,8 @@ def profile(arguments):
                 protocol=protocol,
                 icmp_type=arguments.get("<ICMPTYPE>"),
                 icmp_code=arguments.get("<ICMPCODE>"),
+                icmpv6_type=arguments.get("<ICMPV6TYPE>"),
+                icmpv6_code=arguments.get("<ICMPV6CODE>"),
```
robbrockbank: As mentioned in the libcalico change.  I think this should use icmp_type and icmp_code for both versions

calicoctl/calico_ctl/profile.py
```diff
@@ -35,6 +38,9 @@
       icmp [(type <ICMPTYPE> [(code <ICMPCODE>)])]
            [(from [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
            [(to   [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])] |
+      icmpv6 [(type <ICMPTYPE> [(code <ICMPCODE>)])]
+             [(from [(tag <SRCTAG>)] [(cidr <SRCCIDR>)])]
+             [(to   [(tag <DSTTAG>)] [(cidr <DSTCIDR>)])] |
```
robbrockbank: Shouldn't we be validating the CIDRs also?  ICMPv6 using IPv6 CIDRs and sim. for ICMP (v4).

Maybe worth checking that we don't mix IPv4 and IPv6 source and destination CIDRs too.

TrimBiggs: Added in validate function

calicoctl/calico_ctl/profile.py
```diff
@@ -103,8 +110,10 @@ def validate_arguments(arguments):
 
     # Validate IPs
     cidr_ok = True
+    cidr_list = []
     for arg in ["<SRCCIDR>", "<DSTCIDR>"]:
         if arguments.get(arg) is not None:
+            cidr_list.append(arguments[arg])
```
robbrockbank: This doesn't check that the two CIDR versions are self consistent

robbrockbank:
LGTM

PR #709, opened 2016-01-05
calicoctl status exits with 1 on error

### LINKED ISSUE #697: calicoctl status should return code 1 if errors detected
Currently, the return code is 0 even when it detects an error:
```
root@6044a56f5186:/isolator#  calicoctl status
calico-node container not running
root@6044a56f5186:/isolator# echo $?
0
```

There should be a sensible way to tell if calico is healthy via calicoctl. 

Use Case: the Netmodules test framework will fail if it is executed before the `calico-node` container has started on the slave. Currently, we use an arbitrary 20 second delay to give calico-node time to launch. But it would be much more sensible to have it wait until `calicoctl status` gives a return code of 0 to ensure that it is up, running, and healthy, before starting the tests.

TrimBiggs:
Fixes #697

### DIFF
--- a/tests/st/test_status.py
+++ b/tests/st/test_status.py
@@ -14,6 +14,7 @@
 
 from test_base import TestBase
 from tests.st.utils.docker_host import DockerHost
+from tests.st.utils.exceptions import CommandExecError
 
 """
 Test calicoctl status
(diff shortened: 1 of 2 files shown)


robbrockbank:
looks like there are valid st failures

tests/st/test_status.py
```diff
@@ -28,5 +28,5 @@ def test_status(self):
         """
         Test that the status command can be executed.
         """
-        with DockerHost('host', dind=False, start_calico=False) as host:
+        with DockerHost('host', dind=False, start_calico=True) as host:
```
robbrockbank: What about testing when calico is not running

TrimBiggs: Added (below)

tests/st/test_status.py
```diff
@@ -28,5 +29,17 @@ def test_status(self):
...
+        """
+        Test that the status command fails when calico node is not running
+        """
+        with DockerHost('host', dind=False, start_calico=False) as host:
+            try:
+                host.calicoctl("status")
+            except CommandExecError as e:
+                self.assertEquals(e.returncode, 1)
+                self.assertEquals(e.output,
+                                  "calico-node container not running\n")
```
robbrockbank: you should also be asserting if you don't get a command exec error.

TrimBiggs: Done

robbrockbank:
LGTM

PR #717, opened 2016-01-07
Filter out additional bridge interfaces

### LINKED ISSUE #713: calico guessed wrong host IP address
Following the k8s install instructions for Ubuntu, I started calico via systemd.  It used the wrong IP address for BGP, choosing the IP address on (I think) the `lxcbr0` device instead of the one on `eth0`.  

It's probably best if I make the instructions explicit about the IP anyway but I thought we had heuristics in place that should have done the right thing.

robbrockbank:
Fixes #713


TrimBiggs:
LGTM

PR #719, opened 2016-01-07
Add AS number and IP address to BGP details

### LINKED ISSUE #714: Add local BGP addresses to calicoctl status
Adding that would make it easier to diagnose if calico has chosen the wrong local IP address to connect to the peer.

We could also add the AS numbers.

robbrockbank:
Fixes #714

### DIFF
--- a/calicoctl/calico_ctl/status.py
+++ b/calicoctl/calico_ctl/status.py
@@ -25,8 +25,10 @@ import sys
 from prettytable import PrettyTable
 from requests import ConnectionError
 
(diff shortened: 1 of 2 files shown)


robbrockbank:
@TrimBiggs This will be broken until the libcalico changes go in.

TrimBiggs:
You'll have to update the `calicoctl status` docs as well with the new output.

robbrockbank:
Good spot!  Will do.

PR #731, opened 2016-01-12
Remove the --kubernetes and --rkt commands

tomdee:


calicoctl/calico_ctl/node.py
```diff
@@ -297,26 +274,6 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
...
-            install_plugin(KUBERNETES_PLUGIN_DIR_BACKUP, url)
-
-    # Install rkt plugin
-    if rkt:
-        try:
-            # Attempt to install to the default rkt directory
-            install_plugin(RKT_PLUGIN_DIR, RKT_BINARY_URL)
-        except OSError:
-            # Use the backup directory
-            install_plugin(RKT_PLUGIN_DIR_BACKUP, RKT_BINARY_URL)
```
caseydavenport: Should we be removing the entire `install_plugin` function?  I don't see it used anywhere else.

caseydavenport:
One comment, then LGTM.

PR #734, opened 2016-01-13
Rename calico-docker to calico-containers

robbrockbank:
Tom - rename of calico-docker to calico-containers.

Search and replace with some fix up.


tomdee:
Do we need to wait till tomorrow to get the opinions @eepyaich on the analytics chage?
And is it safe to change the URLs in the badges? e.g. are the circleci and semaphore projects still using the old name?

robbrockbank:
@tomdee Good spot!  CircleCI is fine, but Semaphore and Coveralls is not.  I'll change those back and update the issue that it still needs doig.

@eepyaich was ok with us taking the hit.

eepyaich:
Yep - I'm happy to take the hit on the analytics.  Better to be right going forward and fix up the history...

PR #740, opened 2016-01-15
Annotation based policy doc

caseydavenport:


tomdee:
This has WIP in title but is assigned to me. Is it ready to review?

caseydavenport:
Yep sorry, it's ready.

docs/kubernetes/Policy.md
```diff
@@ -0,0 +1,133 @@
...
+        "type": "k8s-annotations",
+        "k8s_api_root": "<KUBERNETES_API_ROOT>",
+        "k8s_auth_token": "<AUTH_TOKEN>"
+    }
+}
+```
+
+The following configuration optons are supported in the `policy` section:
+- `type`: The type of policy to use.  Currently, only `k8s-annotations` is supported.
+- `k8s_api_root`: (Optional) Location of the Kubernetes API.  Default: `https://10.100.0.1:443/api/v1/`
```
tomdee: Should this default to http:////10.100.0.1:443/api/v1/ (i.e. no s)? It would mean simpler config on non TLS enabled cluster?

docs/kubernetes/Policy.md
```diff
@@ -0,0 +1,133 @@
...
+- `k8s_auth_token`: (Optional) ServiceAccount token for accessing a secure API.  Default: `None`
+
+Once you have modified the network configuration file, you will need to restart the kubelet to pick up the changes.
```
tomdee: Just for my education is this really true? The kubelet reads the config on startup only?

tomdee:
Couple of questions, otherwise looks good. Go ahead and merge when you've squashed.

docs/kubernetes/Policy.md
caseydavenport: I don't think it would - that is the service VIP for the Kubernetes API, which I believe is authenticated even when TLS is disabled (could be wrong though, so I'll double check).

docs/kubernetes/Policy.md
caseydavenport: Yep :(

I think it does it to prevent having to read from disk all the time, which makes sense, but it's much more prone to mistakes when mucking around with config like this.

docs/kubernetes/Policy.md
caseydavenport: API service is still authenticated, can't access using service unless using TLS

PR #763, opened 2016-01-25
Remove the calico/build image from this repo

tomdee:
Moving it to libcalico instead since that repo contains the main versioned part.


caseydavenport:
LGTM

PR #764, opened 2016-01-25
Simplify build of calicoctl now that we have a newer PyInstaller

tomdee:


tomdee:
@caseydavenport This is passing tests. Could you take a look when you get a mo.

caseydavenport:
LGTM.

Have you checked that a built binary actually works?  I've noticed issues with the calico-cni builds since the new build container.

PR #770, opened 2016-01-29
[DO NOT MERGE] Add instructions to install Docker to Mesos RPM

### LINKED ISSUE #753: Calico-Mesos RPM should install latest docker
https://github.com/projectcalico/calico-containers/blob/master/docs/mesos/RpmInstallCalicoMesos.md,
I follow the above url to install mesos+calico, but we have to use docker 1.9.0. but the calico-mesos.rpm use docker 1.8, so how i do this?

djosborne:
Do not merge until new calico-mesos release is cut, as 0.1.3 will install docker automatically.
This doc update resolves #753


djosborne:
Deprecated, documentation is moved to https://github.com/projectcalico/calico-mesos-deployments

PR #773, opened 2016-02-01
Use hostname in etcd authority when running etcd with SSL

TrimBiggs:


docs/EtcdSecureCluster.md
```diff
@@ -15,21 +15,23 @@ Calico supports insecure and TLS/certificate-enabled etcd clusters.
...
+* **`ETCD_KEY_FILE`**: The full path to the client key file for accessing the 
  etcd cluster.
 
 For example:
 
 ```
-export ETCD_AUTHORITY=127.0.0.1:2379
+export ETCD_AUTHORITY=hostname:2379
 export ETCD_SCHEME=https
 export ETCD_CA_CERT_FILE=/path/to/ca.crt
```
tomdee: Can you make sure you provide an example that uses the paths created my the make target? I want to be able to run `make create-certs` (or whatever) then just copy and paste the exact environment I need for running calicoctl talking to secure etcd.

Makefile
```diff
@@ -162,6 +168,12 @@ st-ssl: run-etcd-ssl dist/calicoctl docker calico_test/.calico_test.created busy
 	           calico/test \
 	           sh -c 'cp -ra tests/st/* /tests/st && cd / && nosetests $(ST_TO_RUN) -sv --nologcapture --with-timer $(ST_OPTIONS)'
 
+add-ssl-hostname:
```
tomdee: Why is this required? Can't we just use the actual hostname?

robbrockbank: Might it be a little bad blatting over /etc/hosts with your own host name configuration.

TrimBiggs: The certificate requires a specific hostname to be used.  The tool for creating certificates uses json files that have the hostname hardcoded.  We could potentially have a template that creates a json file using your hostname, but I think this is fine.

robbrockbank:
Make sure libcalico will be ok - and the LGTM

PR #777, opened 2016-02-02
Don't depend on docker for calicoctl status

tomdee:


calicoctl.spec
```diff
@@ -0,0 +1,28 @@
+# -*- mode: python -*-
```
caseydavenport: Isn't this a generated file? Should this be checked in?

calicoctl/calico_ctl/status.py
```diff
@@ -109,20 +110,22 @@ def pprint_bird_protocols(version):
     """
     # Based on the IP version, run the appropriate BIRD command, and select
     # the appropriate separator char for an IP address.
+    if getattr(sys, 'frozen', False ):
```
caseydavenport: extra space at the end - `False )`

calicoctl.spec
caseydavenport: Nevermind, I figured it out.

caseydavenport:
Yeah, this looks swell.

PR #781, opened 2016-02-02
Use systemd-run and update docs

tomdee:


caseydavenport:
LGTM

caseydavenport:
Do we need "  - path: /home/core/get_rkt.sh" anymore?

PR #783, opened 2016-02-03
Fix status command

tomdee:
And allow users to choose their own image

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -436,6 +436,10 @@ def _start_node_container_rkt(ip, ip6, node_image, etcd_envs,
     files to mount on the container
     :return: None.
     """
+    if node_image == "calico/node:latest":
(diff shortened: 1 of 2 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -436,6 +436,10 @@ def _start_node_container_rkt(ip, ip6, node_image, etcd_envs,
     files to mount on the container
     :return: None.
     """
+    if node_image == "calico/node:latest":
```
caseydavenport: What about `calico/node:vX.X.X`? Don't we still want to pull those from docker?

Will the release script update this to be the correct version on release?

caseydavenport:
One comment.

calicoctl/calico_ctl/node.py
tomdee: I _think_ it will. I'll be sure to check when I do the release.

caseydavenport:
LGTM

PR #785, opened 2016-02-03
Handle multiple calico/node containers in rkt status

caseydavenport:


calicoctl/calico_ctl/status.py
```diff
@@ -145,8 +145,18 @@ def check_container_status_rkt():
...
+        statuses = [c[2] for c in containers]
+        running = "running" in statuses
+
+        # If one is running, status is "running".  Else, use the status of
+        # the first container.
+        status = "running" if running else statuses[0]
+
+        # Print status.  If it at least one is running, this will display
+        # "running" status.
+        print "calico-node container is running. Status: %s" % status
```
tomdee: Won't `calico-node container is running. Status: exited` look a bit odd? What about just changing the text to `calico-node container status: %s`

tomdee:
One comment. Otherwise LGTM

PR #790, opened 2016-02-04
Support calico/ctl being run inside a container

### LINKED ISSUE #309: Package calicoctl into a calico/ctl docker image
This may or may not support the commands that interact with docker (e.g. starting a node and adding networking to container). Maybe they would only work if the docker sock is mapped into the container...

tomdee:
Add dockerfile and better support in calicoctl node for ignoring things it can't do.

Fixes #309

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -33,7 +33,7 @@ from subprocess32 import call
 from checksystem import check_system
 from connectors import client
 from connectors import docker_client
-from utils import REQUIRED_MODULES
+from utils import REQUIRED_MODULES, running_in_container
 from utils import get_container_ipv_from_arguments
 from utils import hostname
 from utils import print_paragraph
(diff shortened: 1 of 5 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -227,22 +227,23 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
...
 
-    # Print warnings for any known system issues before continuing
-    using_docker = True if runtime == 'docker' else False
-    (_, _, etcd_ok) = \
-        check_system(quit_if_error=False, libnetwork=libnetwork_image,
-                     check_docker=using_docker)
+        # Print warnings for any known system issues before continuing
+        using_docker = True if runtime == 'docker' else False
+        (_, _, etcd_ok) = \
+            check_system(quit_if_error=False, libnetwork=libnetwork_image,
```
caseydavenport: don't we still want to check if etcd is running and the correct version?

caseydavenport:
Tests failing, and one comment, but otherwise LGTM.

PR #794, opened 2016-02-05
Add more diags and fix broken journalctl call

### LINKED ISSUE #738: diags should save results of `ip addr`
As the title states, it would be useful if the `calicoctl diags` command would, in addition to saving `ip route`, also saved `ip addr` output in the diags bundle.

robbrockbank:
Fixes #738

### DIFF
--- a/calicoctl/calico_ctl/diags.py
+++ b/calicoctl/calico_ctl/diags.py
@@ -160,6 +160,11 @@ def save_diags(log_dir, runtime="docker"):
             print "  - Error running ipset. Maybe you need to run as root."
             f.writelines("Error running ipset: %s\n" % e)
 
+    # Ask Felix to dump stats to its log file - ignore errors as the
+    # calico/node container might not be running.  Gathering of the logs is
+    # dependent on environment.
+    subprocess.call(["pkill", "-SIGUSR1", "felix"])
+
     # If running under rkt, get the journal for the calico/node container.
     print "Copying journal for calico-node.service"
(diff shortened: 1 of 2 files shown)


robbrockbank:
Casey could you review this one?

Primary purpose was to add "ip addr" output to diags.  However, I was also hitting exceptions running the diags command due to the journalctl changes.  I've updated to use the DiagsFileWriter which catches all exceptions and writes a generic log to the file rather than bailing out of the diags script.

robbrockbank:
Fixes #738

calicoctl/calico_ctl/diags.py
```diff
@@ -144,23 +160,22 @@ def save_diags(log_dir, runtime="docker"):
...
+            journalctl = sh.Command._create("journalctl")
+            f.write(journalctl("-u", "calico-node.service", "--no-pager"))
+        except sh.CommandNotFound as e:
+            print "  - Missing command: %s" % e.message
+            f.writelines("Missing command: %s\n" % e.message)
+        except sh.ErrorReturnCode_1 as e:
+            print "  - Error running journalctl."
+            f.writelines("Error running journalctl: %s\n" % e)
+
+    # Ask Felix to dump stats to its log file - ignore errors as the
```
caseydavenport: I think this bit about Felix needs to go before the journalctl dump.  When running under rkt, calico/node logs all go to journalctl.

robbrockbank: Ah gotcha - I'll beef up the comments then,

caseydavenport:
One comment, once that is done it should be good to go.

PR #799, opened 2016-02-05
Fix release script and update docs to point to correct versions

robbrockbank:


release-scripts/do_release.py
```diff
@@ -95,7 +95,7 @@
     (re.compile(r'__felix_version__\s*=\s*".*"'),
      '__felix_version__ = "{calico-version}-dev"'),
 
-    (re.compile(r'https://github\.com/projectcalico/calico\-docker/blob/.*/README\.md'),
+    (re.compile(r'https://github\.com/projectcalico/calico\-containers/blob/v*/README\.md'),
```
caseydavenport: We don't want to update `master` links?

robbrockbank: This just updates the top of each MD page - so that the link for the latest release is up to date.

caseydavenport:
LGTM

PR #800, opened 2016-02-05
Add check for bad pool size when adding pools with calicoctl

### LINKED ISSUE #505: pools with ranges don't work well with IPAM
`calicoctl pool range add ...` creates a bunch of small pools that add up to the range identified.  This is a problem if Calico IPAM is used, since we use a fixed-size block, so it will ignore any pools smaller than the block size (64-addresses in latest version).

In the worst case, if no pools are the block size or larger, we will completely fail to allocate addresses.

c.f. error logs in Kubernetes: http://pastebin.com/PGMe67Ws

TrimBiggs:
Fixes #505

### DIFF
--- a/calicoctl/calico_ctl/pool.py
+++ b/calicoctl/calico_ctl/pool.py
@@ -33,6 +33,7 @@ import netaddr
 from netaddr import IPNetwork, IPRange, IPAddress
 from prettytable import PrettyTable
 from pycalico.datastore_datatypes import IPPool
+from pycalico.block import CidrTooSmallError, BLOCK_PREFIXLEN
 
 from connectors import client
 from utils import (validate_cidr, validate_ip,
@@ -109,7 +110,8 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
 
     :param cidrs: The pools to set in CIDR format, e.g. 192.168.0.0/16
     :param version: 4 or 6
-    :param ipip: Use IP in IP for this pool.
+    :param ipip: Use IP in IP for the pool(s).
+    :param masquerade: Enable masquerade (outgoing NAT) for the pool(s).
     :return: None
     """
     if version == 6 and ipip:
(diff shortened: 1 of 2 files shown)


calicoctl/calico_ctl/pool.py
```diff
@@ -151,8 +162,18 @@ def ip_pool_range_add(start_ip, end_ip, version, ipip, masquerade):
             sys.exit(1)
 
     cidrs = netaddr.iprange_to_cidrs(start_ip, end_ip)
+    new_pools = []
     for ip_net in cidrs:
-        new_pool = IPPool(ip_net.cidr, ipip=ipip, masquerade=masquerade)
+        try:
+            new_pools.append(IPPool(ip_net.cidr, ipip=ipip, masquerade=masquerade))
+        except CidrTooSmallError:
+            print "IPv%s range must start and end on a %s prefix network." \
```
robbrockbank: I think this error message needs a little work.  I'm wondering if we should make it more transparent what this command does - i.e. inform the user that we are breaking the range into CIDRs and storing as CIDRs.  If we don't then the comment about the offending pool would seem a little odd.

Not sure the commend about starting and ending on a /26 prefix network is true.  I think it is more that the start and end ip address must have at least the lower n bits unset.  I've no idea what the terminology would be for that - although you could talk about an individual pool being too small.

TrimBiggs: How about:
```
"The IP range is split into multiple CIDRs using prefix length of '/26' or lower.  When formed into CIDRs, the range specified includes %s with prefix length %s, which is too large.  No pools were added."
```

TrimBiggs: Or possibly displaying all pools that would be added.

robbrockbank: I'm happy with the pools being displayed

robbrockbank:
lgtm

PR #806, opened 2016-02-08
Add kernel version check to calicoctl checksystem cmd

TrimBiggs:


calicoctl/calico_ctl/checksystem.py
```diff
@@ -246,3 +253,37 @@ def _check_etcd_version():
...
+        print "ERROR: The kernel version does not match expected semantic " \
+              "versioning."
+        return False
+
+    version_list = version_string.split(".")
+
+    version_list = [int(num) for num in version_list]
+
+    i = 0
+    while i < len(version_list):
```
robbrockbank: you can just do version_list > compare_version_list

TrimBiggs: oo that's convenient.

calicoctl/calico_ctl/checksystem.py
```diff
@@ -246,3 +253,37 @@ def _check_etcd_version():
...
+
+
+def _check_kernel_version():
+    """
+    Check that the kernel version meets the minimum required version for Calico.
+    """
+    full_version = platform.uname()[2]
+
+    # We just need the Major, and Minor, and Bug fix version numbers
+    regex = re.compile("([\d]\.[\d]+\.[\d]+)[-+].*")
```
robbrockbank: I wonder if there is always a - or +?   or can it ever be just a simple version string 1.2.3?

TrimBiggs: Yeah I think that makes sense.

calicoctl/calico_ctl/checksystem.py
```diff
@@ -246,3 +253,37 @@ def _check_etcd_version():
...
+    Ex.    "2.24.32-100-generic"
+           "3.19.0a-dev"
+           "2.8.1"
+           "3.0"
+           "4.5-rc3"
+    """
+    full_version = platform.uname()[2]
+
+    # Make sure version number makes sense and captures version numbers
+    regex = re.compile("^([\d]\.[\d]{1,2}(\.[\d]{1,2})?)(?:\D.*)?$")
```
robbrockbank: Wouldn't the first bit be easier as something like `([\d](\.[\d]?){1,3}`

i.e. capture the first number and between 1-3  .x 's

robbrockbank: It doesn't seem right to assume the number of digits for each version segment.  I think you probably want something like this:

`^(\d+(?:\.\d+){1,3})\D.*$`

Where you are saying the is always a first number followed by 1-3 .x following by something non-numerical and some other stuff.

robbrockbank:
lgtm once the regex is sorted out

calicoctl/calico_ctl/checksystem.py
```diff
@@ -246,3 +253,37 @@ def _check_etcd_version():
             system_ok = False
 
     return system_ok
+
+
+def _check_kernel_version():
+    """
+    Check that the kernel version meets the minimum required version for Calico.
+    Valid kernel versions have 1 Kernel digit, 1-2 Major digits, and optionally
```
robbrockbank: Is it right to talk about the number of digits?  Surely the key thing is that each segment is a number, not a set number of digits?

calicoctl/calico_ctl/checksystem.py
TrimBiggs: That could allow you to have version numbers of 2.85789.1 as a valid number.  I think we should limit it to 3 digits max since no number in the linux kernel version has ever been larger than 3 digits (3.4.110 is the largest 3-digit kernel version used).

I'd be happy changing this to:
`^(\d{1,3}(?:\.\d{1,3}){1,3})(?:\D.*)?$`

the `(?:\D.*)?` is important to allow standalone versions like `3.0` or `2.19.26`.

I originally didn't care about capturing the 4th number in an older version string because we don't care about it in this case, but I'll include this for reuse.

What do you think?

calicoctl/calico_ctl/checksystem.py
TrimBiggs: Yup, will change.

calicoctl/calico_ctl/checksystem.py
robbrockbank: If there is a compelling reason to limit the number of allowed digits then fine.  But strictly speaking 2.85789.1 is a valid format despite there never having been such a Kernel version, no?

Yup point taken about the last bit needing the 0 or 1 repetitions.

PR #811, opened 2016-02-11
Use docker 1.10 in STs

TrimBiggs:


Makefile
```diff
@@ -75,11 +75,19 @@ routereflector.tar:
 	docker pull calico/routereflector:latest
 	docker save --output routereflector.tar calico/routereflector:latest
 
-## Download the latest docker binary
 docker:
+	# Download the docker 1.9.1 binary
 	curl https://get.docker.com/builds/Linux/x86_64/docker-1.9.1 -o docker
 	chmod +x docker
 
+docker-1-10:
```
robbrockbank: This all seems rather specific to the semaphore environment, I would just add it there.  Does anything rely on the docker make target - seems odd that we have it now that there are standard ways to install 1.9.1

Makefile
```diff
@@ -75,11 +75,19 @@ routereflector.tar:
...
+	# Download the docker 1.10.0 binary and start it
+	curl https://get.docker.com/builds/Linux/x86_64/docker-1.10.0 -o docker
+	chmod +x docker
+	service docker stop
+	cp ./docker $(shell which docker)
+	service docker start
```
robbrockbank: Maybe run docker version so you can check the semaphore logs

tomdee:
You should use the latest version which is 1.10.1

calico_test/tests/st/utils/workload.py
```diff
@@ -45,12 +45,14 @@ def __init__(self, host, name, image="busybox", network="bridge"):
         has ping.
         :param network: The DockerNetwork to connect to.  Set to None to use
         default Docker networking.
+        :param ip: The ip address to assign to the container (libnetwork only).
```
robbrockbank: Don;t think we need the libnetwork only note - it is a valid docker option.

Makefile
```diff
@@ -76,9 +76,9 @@ routereflector.tar:
 	docker pull calico/routereflector:latest
 	docker save --output routereflector.tar calico/routereflector:latest
 
-## Download the latest docker binary
 docker:
-	curl https://get.docker.com/builds/Linux/x86_64/docker-1.9.1 -o docker
+	# Download the docker 1.10.0 binary
+	curl https://get.docker.com/builds/Linux/x86_64/docker-1.10.0 -o docker
```
robbrockbank: 1.10.1 now - we should probably be using that.

robbrockbank:
Couple of comments then LGTM

PR #813, opened 2016-02-12
Update rkt guide with nwe version of CNI plugin

tomdee:


docs/cni/rkt/README.md
```diff
@@ -122,31 +122,30 @@ cat >/etc/rkt/net.d/10-calico-net2.conf <<EOF
...
 ```
 
-
 ## 4. Create a webserver on one of the networks
 
 With the networks created, let's start a container on `calico-01`
 
 ### On calico-01
-Create a webserver using an `nginx` container. We'll put this on `net1` network.
+Create a webserver using an `busybox` container running an `httpd` daemon. We'll put this on `net1` network.
```
caseydavenport: Why the switch to busybox?

tomdee: It's smaller

PR #815, opened 2016-02-16
Fix typo

depay:
Signed-off-by: depay <depay19@163.com>


robbrockbank:
Hi @depay.  Thanks for the PR.  Apologies for not merging - as it happens I made a similar fix in this file alongside an addition to prevent us from configuring IPv6 when it is not enabled on the OS.

PR #817, opened 2016-02-18
Removed mesos documentation

djosborne:
Replaced all calico-mesos documentation with links to their new location


docs/mesos/DockerizedDeployment.md
```diff
@@ -1,220 +1 @@
...
-```
-> NOTE: Some tests require multiple hosts to ensure cross-host communication, and may fail unless you are running 2+ agents.
-
-Additionally, you can launch your own tasks using Marathon. See our [Marathon Task Launch Instructions](README.md#3-launching-tasks) for more information.
-
-[calico]: http://projectcalico.org
-[mesos]: https://mesos.apache.org/
-[net-modules]: https://github.com/mesosphere/net-modules
-[![Analytics](https://ga-beacon.appspot.com/UA-52125893-3/calico-containers/docs/mesos/DockerizedDeployment.md?pixel)](https://github.com/igrigorik/ga-beacon)
+The contents you are looking for have moved to the new calico-mesos-deployments repo. [Follow this link to view them](https://github.com/projectcalico/calico-mesos-deployments/blob/master/docs/DockerizedDeployment.md)
```
robbrockbank: why no just make "calico-mesas-deployments" linkable?

robbrockbank:
Don't you need to update the main README to start pointing to the new repo?

docs/mesos/DockerizedDeployment.md
djosborne: My thought was that I only want them clicking on the link that takes them directly to the doc they were trying to look at originally

djosborne:
The main README does point to the new repo

robbrockbank:
Yup this is OK.  I'm still not convinced about keeping the old files around though - my concern is that it is a maintenance nightmare.  What happens if we restructure the deployments repo?

I'm ok with these going in, but I think these files should be considered deprecated and we will ultimately just remove them.

PR #824, opened 2016-02-24
Set enforce_root() when running calicoctl status getting bird info

### LINKED ISSUE #822: calicoctl status needs "enforce_root"
Since we've changed the status command to access the `/var/run/calico` directory on the host, it needs an enforce_root.

e.g
```
[cloud-user@ken1 ~]$ calicoctl status
calico-node container is running. Status: Up 12 seconds
Running felix version 1.3.0rc6

IPv4 BGP status
IP: 192.168.98.6 AS Number: 64511 (inherited)
Unable to connect to server control socket (/var/run/calico/bird.ctl): Permission denied
Couldn't connect to bird. Try running as root.
IPv6 BGP status
```

TrimBiggs:
Fixes #822

### DIFF
--- a/calicoctl/calico_ctl/status.py
+++ b/calicoctl/calico_ctl/status.py
@@ -161,6 +161,9 @@ def pprint_bird_protocols(version):
     :param version:  The IP version (4 or 6).
     :return: None.
     """
+    # This needs to be run as root to access the bird data in /var/run/calico
+    enforce_root()
+
     # Based on the IP version, run the appropriate BIRD command, and select
     # the appropriate separator char for an IP address.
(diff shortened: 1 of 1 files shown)


calicoctl/calico_ctl/status.py
```diff
@@ -163,6 +164,8 @@ def pprint_bird_protocols(version):
     """
     # Based on the IP version, run the appropriate BIRD command, and select
```
caseydavenport: I'd keep the above comment below this line, and then add a comment that says "We need to run as root so we can access the BIRD control socket"

caseydavenport:
One comment then LGTM.

PR #834, opened 2016-02-26
Require calicoctl node/node stop to be run as root

### LINKED ISSUE #829: Node stop produces weird output
```
core@calico-04 ~ $ calicoctl node stop
Failed to stop calico-node.service: Interactive authentication required.
Node stopped
```

Needs to be run as sudo, but the above output doesn't indicate that.

### LINKED ISSUE #833: calicoctl node errors do not stop it from being run
Two scenarios I spotted,

Running "calicoctl node" without sudo seems to be allowed - I thought we used to check that it was being run as sudo - has this changed?

Trying to start node with libnetwork with Docker 1.8.1 produces an error message about incorrect Docker version, but then still starts up the libnetwork plugin container.  Is this as expected?

TrimBiggs:
Fixes #829

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -24,20 +24,16 @@ from pycalico.datastore import (ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT,
                                 ETCD_KEY_FILE_ENV, ETCD_CERT_FILE_ENV,
                                 ETCD_CA_CERT_FILE_ENV, ETCD_SCHEME_ENV,
                                 ETCD_SCHEME_DEFAULT)
-from pycalico.datastore_datatypes import BGPPeer
-from pycalico.datastore_datatypes import IPPool
+from pycalico.datastore_datatypes import BGPPeer, IPPool
 from pycalico.netns import remove_veth
 from pycalico.util import get_host_ips
 from subprocess32 import call
 
 from checksystem import check_system
-from connectors import client
-from connectors import docker_client
-from utils import REQUIRED_MODULES, running_in_container
-from utils import get_container_ipv_from_arguments
-from utils import hostname
-from utils import print_paragraph
-from utils import validate_ip, validate_asn, convert_asn_to_asplain
+from connectors import client, docker_client
+from utils import (REQUIRED_MODULES, running_in_container, enforce_root,
+                   get_container_ipv_from_arguments, hostname, print_paragraph,
+                   validate_ip, validate_asn, convert_asn_to_asplain)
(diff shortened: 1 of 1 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -33,11 +33,9 @@
 from checksystem import check_system
 from connectors import client
 from connectors import docker_client
-from utils import REQUIRED_MODULES, running_in_container
-from utils import get_container_ipv_from_arguments
-from utils import hostname
-from utils import print_paragraph
-from utils import validate_ip, validate_asn, convert_asn_to_asplain
+from utils import (REQUIRED_MODULES, running_in_container, enforce_root,
```
robbrockbank: should probably consolidate the ones above for conformity.

robbrockbank:
Tim - I thought calicoctl node used to enforce root.  Do you know if it did, or it's always been missing.

If it did, do you know why it got removed?

calicoctl/calico_ctl/node.py
```diff
@@ -574,6 +575,9 @@ def node_stop(force):
     Stop the Calico node.  This stops the containers (calico/node and
     calico/node-libnetwork) that are started by calicoctl node.
     """
+    # The command has to be run as root to stop the calico-node service
```
robbrockbank: DO we need it on the stop?  The reason we need it on the start is because we need to load kernel modules.

robbrockbank: Ah - we prolly need it foir systemctl call

robbrockbank: Any idea why the systemctl command is in here.  Seems kind of odd to me that we are putting service stuff in this command.

robbrockbank:
LGTM.   I've a few questions but probably shouldn't stop you checking it in.

TrimBiggs:
I specifically remember needing to run the node command as root.  I looked at the history of the command back to November of last year when we changed to using this repo, but there was no `enforce_root()` call anywhere in the history.

calicoctl/calico_ctl/node.py
TrimBiggs: I find it a bit strange too. I'll check around.

caseydavenport: I can answer that: we've baked in `systemd` for when we're running in rkt.  Take a look at how `calicoctl node --runtime=rkt` works.

caseydavenport: Long story short, yes we do want the `systemctl` commands here because that is how we manage the `calico/node` container when running in rkt.

PR #843, opened 2016-03-04
Update AWS docs to use Disable Source/Dest. Check instead of IP-in-IP

matthewdupre:
I think this covers all the relevant docs, but it's possible I've missed some (particularly if there's anything outside this repository).

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -25,33 +25,33 @@ Container Ports to the Internet](ExposePortsToInternet.md).
 
 ## Can Calico containers use any IP address within a pool, even subnet network/broadcast addresses?
 
-Yes!  Calico is fully routed, so all IP address within a Calico pool are usable as 
-private IP addresses to assign to a workload.  This means addresses commonly 
-reserved in a L2 subnet, such as IPv4 addresses ending in .0 or .255, are perfectly 
+Yes!  Calico is fully routed, so all IP address within a Calico pool are usable as
+private IP addresses to assign to a workload.  This means addresses commonly
+reserved in a L2 subnet, such as IPv4 addresses ending in .0 or .255, are perfectly
 okay to use.
 
 ## How do I get network traffic into and out of my Calico cluster?
-The recommended way to get traffic to/from your Calico network is by peering to 
-your existing data center L3 routers using BGP and by assigning globally 
-routable IPs (public IPs) to containers that need to be accessed from the internet. 
-This allows incoming traffic to be routed directly to your containers without the 
+The recommended way to get traffic to/from your Calico network is by peering to
+your existing data center L3 routers using BGP and by assigning globally
+routable IPs (public IPs) to containers that need to be accessed from the internet.
+This allows incoming traffic to be routed directly to your containers without the
 need for NAT.  This flat L3 approach delivers exceptional network scalability
 and performance.
 
-A common scenario is for your container hosts to be on their own 
-isolated layer 2 network, like a rack in your server room or an entire data 
-center.  Access to that network is via a router, which also is the default 
+A common scenario is for your container hosts to be on their own
+isolated layer 2 network, like a rack in your server room or an entire data
+center.  Access to that network is via a router, which also is the default
 router for all the container hosts.
 
-If this describes your infrastructure, the 
+If this describes your infrastructure, the
 [External Connectivity tutorial](ExternalConnectivity.md) explains in more detail
 what to do. Otherwise, detailed datacenter networking recommendations are given
 in the main [Project Calico documentation](http://docs.projectcalico.org/en/latest/index.html).
-We'd also encourage you to [get in touch](http://www.projectcalico.org/contact/) 
+We'd also encourage you to [get in touch](http://www.projectcalico.org/contact/)
 to discuss your environment.
 
 ### How can I enable NAT for outgoing traffic from containers with private IP addresses?
-If you want to allow containers with private IP addresses to be able to access the 
+If you want to allow containers with private IP addresses to be able to access the
 internet then you can use your data center's existing outbound NAT capabilities
 (typically provided by the data center's border routers).
 
@@ -63,19 +63,19 @@ node on which each container is hosted.
 ```
 Where `<CIDR>` is the CIDR of your IP pool, for example `192.168.0.0/16`.
 
-Remember: the security profile for the container will need to allow traffic to the 
-internet as well. You can read about how to configure security profiles in the 
+Remember: the security profile for the container will need to allow traffic to the
+internet as well. You can read about how to configure security profiles in the
 [Advanced Network Policy tutorial](AdvancedNetworkPolicy.md).
 
 ### How can I enable NAT for incoming traffic to containers with private IP addresses?
-As discussed, the recommended way to get traffic to containers that 
+As discussed, the recommended way to get traffic to containers that
 need to be accessed from the internet is to give them public IP addresses and
 to configure Calico to peer with the data center's existing L3 routers.
 
-In cases where this is not possible then you can configure incoming NAT 
+In cases where this is not possible then you can configure incoming NAT
 (also known as DNAT) on your data centers existing border routers. Alternatively
 you can configure incoming NAT with port mapping on the host on which the container
-is running on. 
+is running on.
 ```
 iptables -A PREROUTING -t nat -i eth0 -p tcp --dport <EXPOSED_PORT> -j DNAT  --to <CALICO_IP>:<SERVICE_PORT>
 ```
@@ -87,10 +87,17 @@ The command will need to be run each time the host is restarted.
 
 Remember: the security profile for the container will need to allow traffic to the exposed port as well.  You can read about how to configure security profiles in the [Advanced Network Policy](AdvancedNetworkPolicy.md) guide.
 
-### Can I run Calico in a public cloud environment? 
+### Can I run Calico in a public cloud environment?
 Yes.  If you are running in a public cloud that doesn't allow either L3 peering or L2 connectivity between Calico hosts then you can specify the `--ipip` flag your Calico IP pool:
 ```
 ./calicoctl pool add <CIDR> --ipip --nat-outgoing
 ```
 Calico will then route traffic between Calico hosts using IP in IP.
+
+In AWS, you can instead disable `Source/Dest. Check` as long as all your instances are in the same subnet of your VPC.  This will provide the best performance.  You can disable this with the CLI, or right click the instance in the EC2 console, and `Change Source/Dest. Check` from the `Networking` submenu.
+```
+aws ec2 modify-instance-attribute --instance-id <INSTANCE_ID> --source-dest-check "{\"Value\": false}"
+...
+./calicoctl pool add <CIDR> --nat-outgoing
+```
 [![Analytics](https://calico-ga-beacon.appspot.com/UA-52125893-3/calico-containers/docs/FAQ.md?pixel)](https://github.com/igrigorik/ga-beacon)
--- a/docs/calico-with-docker/AWS.md
+++ b/docs/calico-with-docker/AWS.md
@@ -8,16 +8,16 @@
 
 # Running the Calico tutorials on AWS
 Calico is designed to provide high performance massively scalable virtual
-networking for private data centers. But you can also run Calico within a 
-public cloud such as Amazon Web Services (AWS).  The following instructions 
-show how to network containers using Calico routing and the Calico security 
+networking for private data centers. But you can also run Calico within a
+public cloud such as Amazon Web Services (AWS).  The following instructions
+show how to network containers using Calico routing and the Calico security
 model on AWS.
 
 ## 1. Getting started with AWS
-These instructions describe how to set up two CoreOS hosts on AWS.  For more general background, see 
+These instructions describe how to set up two CoreOS hosts on AWS.  For more general background, see
 [the CoreOS on AWS EC2 documentation](https://coreos.com/docs/running-coreos/cloud-providers/ec2/).
 
-Download and install AWS Command Line Interface: 
+Download and install AWS Command Line Interface:
 ```
 curl "https://s3.amazonaws.com/aws-cli/awscli-bundle.zip" -o "awscli-bundle.zip"
 unzip awscli-bundle.zip
@@ -35,17 +35,17 @@ aws configure
 ```
 Note: Your `<Region Name>` can be found on the front page of the EC2 dashboard under the "Service Health" text.
 
-Your AWS user needs to have the policy AmazonEC2FullAccess or be in a group with this policy in order to run the ec2 
+Your AWS user needs to have the policy AmazonEC2FullAccess or be in a group with this policy in order to run the ec2
 commands.  This can be set in the Services>IAM>Users User configuration page of the web console.
-For more information on configuration and keys, see Amazon's 
+For more information on configuration and keys, see Amazon's
 [Configuring the AWS Command Line Interface][configure-aws-cli].
 
 ## 2. Setting up AWS networking
-Before you can use Calico to network your containers, you first need to configure AWS to allow your hosts to talk to 
+Before you can use Calico to network your containers, you first need to configure AWS to allow your hosts to talk to
 each other.
 
-A Virtual Private Cloud (VPC) is required on AWS in order to configure Calico networking on EC2. Your AWS account should have a default VPC that instances 
-automatically attach to when they are created. 
+A Virtual Private Cloud (VPC) is required on AWS in order to configure Calico networking on EC2. Your AWS account should have a default VPC that instances
+automatically attach to when they are created.
 
 To check if you have a default VPC, run the following command, then save VPC ID as an environment variable to use later.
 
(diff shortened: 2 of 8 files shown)


docs/calico-with-docker/AWS.md
```diff
@@ -190,33 +190,38 @@ aws ec2 run-instances \
...
   --security-group-ids $SECURITY_GROUP_ID \
   --user-data file://<PATH_TO_CLOUD_CONFIG>/user-data-others
-#  --subnet $SUBNET_ID 
+#  --subnet $SUBNET_ID
 #  Include the subnet param above if using a non-default VPC
 ```
 
+Finally, disable `Source/Dest. Check` to allow containers to talk between hosts.  You can disable this with the CLI, or in the `Networking` part of the instances' right click menus.  Run the following for each server:
+```
+aws ec2 modify-instance-attribute --instance-id <INSTANCE_ID> --source-dest-check "{\"Value\": false}"
```
caseydavenport: It doesn't look like we explain where to get <INSTANCE_ID> from - we should probably mention where that value comes from.

Also, we should use `$INSTANCE_ID` so that this command can be copied / pasted.

docs/FAQ.md
```diff
@@ -96,10 +96,17 @@ The command will need to be run each time the host is restarted.
...
 
-### Can I run Calico in a public cloud environment? 
+### Can I run Calico in a public cloud environment?
 Yes.  If you are running in a public cloud that doesn't allow either L3 peering or L2 connectivity between Calico hosts then you can specify the `--ipip` flag your Calico IP pool:
 ```
 ./calicoctl pool add <CIDR> --ipip --nat-outgoing
 ```
 Calico will then route traffic between Calico hosts using IP in IP.
+
+In AWS, you can instead disable `Source/Dest. Check` as long as all your instances are in the same subnet of your VPC.  This will provide the best performance.  You can disable this with the CLI, or in the `Networking` part of the instances' right click menus.
```
TrimBiggs: Can you say how this will affect the deployment?

Also (nitpick) the last sentence here is a bit wordy, maybe change the second part to:
"...or you can right click an instance in the EC2 console and choose `Change Source/Dest. Check` under `Networking`."

docs/calico-with-docker/AWS.md
TrimBiggs: Agree with Casey.  Make sure to have `$INSTANCE_ID_1` and `$INSTANCE_ID_2` after creating each instance.  Probably easiest to just have both modify-instance commands written here for copying and pasting.

docs/calico-with-docker/AWS.md
```diff
@@ -190,33 +190,38 @@ aws ec2 run-instances \
...
 ```
 
 There are two worked examples you can follow: Calico as a Docker network
-plugin, or Calico without Docker networking.  Select the instructions based on 
+plugin, or Calico without Docker networking.  Select the instructions based on
 the networking option that you chose for the cloud config in step (3).
 
 > In the worked example, be sure to follow the additional instructions for
-configuring `ipip` and `nat-outgoing`. 
+configuring `ipip` and `nat-outgoing`.
```
TrimBiggs: This isn't necessarily true anymore if we don't require `ipip`.

docs/calico-with-docker/AWS.md
```diff
@@ -283,38 +288,39 @@ curl http://<host public ip>:80
...
 ```
 Repeat for all relevant Calico interfaces on the host.
 
-Note that the value used is 8981.  This is because Calico traffic will flow 
-through the IP-in-IP tunnel.  When packets enter the IP-in-IP tunnel, an IP 
+If you're using IP-in-IP, instead use a value of 8981 throughout.
+This is because Calico traffic will flow
+through the IP-in-IP tunnel.  When packets enter the IP-in-IP tunnel, an IP
 header of length 20 is added to the packet, summing to a total size of 9001.
+You'll also need to set the MTU of the `tunl0` interface to 8981.
```
TrimBiggs: This is probably going away in the next couple weeks, but for simplicity just add this to the line above:

"Repeat for all relevant Calico interfaces and the `tunl0` interface on the host."

docs/calico-with-docker/AWS.md
```diff
@@ -323,20 +329,20 @@ docker run -v /usr/local/bin:/target jpetazzo/nsenter
...
 ```
 
-Finally, modify the MTU of the Calico interface on the container. If you 
-specified an interface when adding the container to Calico, change `eth1` to be 
+Finally, modify the MTU of the Calico interface on the container. If you
+specified an interface when adding the container to Calico, change `eth1` to be
 the name of the interface you passed in.
 ```
-sudo ifconfig eth1 mtu 8981
+sudo ifconfig eth1 mtu 9001
```
TrimBiggs: Why is this changed?

docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -63,13 +63,25 @@ You should see output like this on each node
...
-If you are running in the cloud, you will need to first configure an IP Pool
+If you are running in a cloud other than AWS, you will need to first configure an IP Pool
 with the `ipip` and `nat-outgoing` options.
 
 On either node:
 
     calicoctl pool add 192.168.0.0/16 --ipip --nat-outgoing
-    
+
+Instead, if you are running in AWS and all your endpoints are in the same subnet of your VPC,
```
TrimBiggs: I don't think "Instead" is necessary here.

docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -63,13 +63,25 @@ You should see output like this on each node
...
+disable Source/Dest. checks on each instance.  You can disable this with the CLI,
+or in the `Networking` part of the instances' right click menus.
```
TrimBiggs: Same comment as previous reference to this.

TrimBiggs: Might just be cleaner to include this line and the one below by saying:

"You can disable this by right-clicking on the instance from the EC2 console and choosing `Change Source/Dest. Check`, or by running the following AWS CLI command for each instance:"

docs/cni/kubernetes/GCE.md
```diff
@@ -73,7 +73,7 @@ gcloud compute instances create \
   --image-project coreos-cloud \
   --image coreos-stable-835-9-0-v20151208 \
   --machine-type n1-standard-1 \
-  --metadata-from-file user-data=cloud-config/master-config.yaml
+  --metadata-from-file user-data=cloud-config/master-config-ipip.yaml
```
TrimBiggs: Did you remove or modify the master-config.yaml?

TrimBiggs: Ah, got it, nvm.  Removing comment.

TrimBiggs: Is there somewhere that we want to recommend using the `master-config.yaml` instead of `master-config-ipip.yaml` where `ipip` is not in use?

docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -63,13 +63,25 @@ You should see output like this on each node
 
 If you are not running in the cloud you may skip this step and jump to step 4.
 
-If you are running in the cloud, you will need to first configure an IP Pool
+If you are running in a cloud other than AWS, you will need to first configure an IP Pool
```
TrimBiggs: I think it would look nicer if you had two subsections now that the cloud services would use different commands.

Maybe change the main section to just be **Running in the Cloud**, then have subsections of **GCE / Digital Ocean**  and **AWS**

docs/cni/kubernetes/GCE.md
matthewdupre: Yes: all the other references to this file; they just don't show up in the diff because they didn't need changing.

TrimBiggs: Gotcha.

docs/calico-with-docker/AWS.md
matthewdupre: No need to leave room for IPIP by default (and a potential 0.2% performance gain!!!).  Looking at it again though, the wording above is poor and doesn't allow for people turning IPIP on (which TBH I'd rather rip out entirely, but anyway).

I'll just take the 0.2% hit I think...

TrimBiggs: Ah, I see, I think I misread the above. I think this is fine as 9001, but good to note that here should be 8981 also if using ip-ip.

docs/FAQ.md
matthewdupre: Nitpick: agreed, active is better.

As for how it will affect, I'm going to assert that I have: it "will provide the best performance".  If we're going to explain anything in more detail, it should be IPIP.  I don't want to create FUD where there's really no grounds for it by needlessly explaining things.

matthewdupre:
@TrimBiggs back to you

docs/calico-with-docker/AWS.md
```diff
@@ -283,38 +293,39 @@ curl http://<host public ip>:80
...
 ```
 Repeat for all relevant Calico interfaces on the host.
 
-Note that the value used is 8981.  This is because Calico traffic will flow 
-through the IP-in-IP tunnel.  When packets enter the IP-in-IP tunnel, an IP 
+If you're using IP-in-IP, instead use a value of 8981 throughout.
+This is because Calico traffic will flow
+through the IP-in-IP tunnel.  When packets enter the IP-in-IP tunnel, an IP
 header of length 20 is added to the packet, summing to a total size of 9001.
+Repeat for all relevant Calico interfaces and the tunl0 interface on the host.
```
TrimBiggs: Ah, when I made my comment, I meant to modify the sentence 6 lines up, just below the command.  I think you can just use this sentence there and get rid of it here.

docs/calico-with-docker/AWS.md
```diff
@@ -173,13 +173,15 @@ aws ec2 run-instances \
   --key-name mykey \
   --security-group-ids $SECURITY_GROUP_ID \
   --user-data file://<PATH_TO_CLOUD_CONFIG>/user-data-first
-#  --subnet $SUBNET_ID 
+#  --subnet $SUBNET_ID
 #  Include the subnet param above if using a non-default VPC
+
+INSTANCE_ID_1=<InstanceId>
```
TrimBiggs: There should be a note either be a comment added above here to tell the used to copy the instance ID into this command, or a line above the command to also save the returned instance-id from the output.

This goes for `INSTANCE_ID_2` below also.

TrimBiggs:
Two more comments, then should be good to go

docs/calico-with-docker/AWS.md
matthewdupre: TBH I'm not a fan of this syntax, I think having these additional steps just makes the doc longer (when the variables are generally used once), and people still have to copy the values around.  Every extra sentence in a doc reduces the chance that people will actually follow all the other ones by some tiny percentage, unmeasurable individually but very clear in aggregate.

If we want our docs to be shell scripts, then they should be shell scripts...

caseydavenport: We've generally tried to keep the large commands as copy-and-pasteable as possible - this makes it less frustrating for people who are trying to follow the guide.

e.g:
- copy the instance ID and export it, copy this command and execute it.

rather than:
- Copy this command somewhere.  Copy the instance ID somewhere. Delete the bit that was a template, fill it in with the instance ID.  Copy and paste that into the terminal (or similar steps).

Especially when you need to capture a bit of configuration in one part of the guide, and then use it again later on, environment variables help with that.  We've received feedback from various docs testers, including @lwr20 that the environment variable method is a better UX.

matthewdupre:
@TrimBiggs OK, tweaked, rebased and squashed.

TrimBiggs:
LGTM

matthewdupre:
Found a missed doc.  It's not the best wording, but it's compact, and these docs are starting to get far too long (considering how many of them cover the same material!)

docs/calico-with-docker/docker-network-plugin/README.md
```diff
@@ -128,6 +128,12 @@ specific environment, you may need to choose different CIDRs.
 So, once you have decided which type of network to create, following the
 appropriate instructions for one of *a)*, *b)*, *c)* or *d)*.
 
+For AWS, chose a) or c), and Disable Source/Dest. checks on your instances with
+the following EC2 CLI command or by right clicking the Instance, and selecting
```
TrimBiggs: Nitpick: Instance -> instance

docs/calico-with-docker/docker-network-plugin/README.md
```diff
@@ -128,6 +128,12 @@ specific environment, you may need to choose different CIDRs.
...
+it from the Networking submenu.
+
+    aws ec2 modify-instance-attribute --instance-id $INSTANCE_ID --source-dest-check "{\"Value\": false}"
```
TrimBiggs: `$INSTANCE_ID` -> `<instance_id>`

docs/calico-with-docker/docker-network-plugin/README.md
```diff
@@ -128,6 +128,12 @@ specific environment, you may need to choose different CIDRs.
 So, once you have decided which type of network to create, following the
 appropriate instructions for one of *a)*, *b)*, *c)* or *d)*.
 
+For AWS, chose a) or c), and Disable Source/Dest. checks on your instances with
+the following EC2 CLI command or by right clicking the Instance, and selecting
+it from the Networking submenu.
```
TrimBiggs: Maybe use the specific name of the option to click instead of "it"? It's not called Disable Source/Dest. here, I think it's `Change Source/Dest. Check`.

docs/calico-with-docker/docker-network-plugin/README.md
TrimBiggs: "... right clicking the instance **in the EC2 console**, ..." ?

docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -63,13 +63,26 @@ You should see output like this on each node
 
 If you are not running in the cloud you may skip this step and jump to step 4.
 
-If you are running in the cloud, you will need to first configure an IP Pool
-with the `ipip` and `nat-outgoing` options.
+### 3.1 AWS
+
+If all your endpoints are in the same subnet of your VPC, disable Source/Dest. checks on each instance.  You can disable this with the CLI, or right click the instance in the EC2 console and `Change Source/Dest. Check` from the `Networking` submenu.
+
+On a node with the AWS CLI installed, for each instance:
```
TrimBiggs: Sorry didn't mention this before, but this could be on any machine with AWS CLI installed, doesn't have to be one of the nodes.  Could just say "Using the AWS CLI..."

docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -63,13 +63,26 @@ You should see output like this on each node
...
+
+    aws ec2 modify-instance-attribute --instance-id <INSTANCE_ID> --source-dest-check "{\"Value\": false}"
+
+And then configure the `nat-outgoing` option on either Calico node:
```
TrimBiggs: Maybe this could be clearer about running regardless of setting the Source/Dest. Check in CLI or in the EC2 console, I'd be worried people might think it is just when running the CLI.

Something like: "After disabling Source/Dest. Checks, configure the `nat-outgoing` option on your Calico IP pool from any Calico node:"

docs/cni/kubernetes/AWS.md
```diff
@@ -212,10 +212,15 @@ aws ec2 run-instances \
...
   --security-group-ids $SECURITY_GROUP_ID \
   --user-data file://cloud-config/node-config.yaml
-#  --subnet $SUBNET_ID 
+#  --subnet $SUBNET_ID
 #  Include the subnet param above if using a non-default VPC
 ```
 
+Finally, disable `Source/Dest. Check` on each instance (including the master) to allow routing between pods without needing IP in IP.  All instances must be in the same subnet.  You can do this with the CLI, or in the `Networking` part of the instances' right click menus.
+```
+aws ec2 modify-instance-attribute --instance-id $INSTANCE_ID --source-dest-check "{\"Value\": false}"
```
TrimBiggs: The `$INSTANCE_ID` environment variables are not defined in this doc.  You should add these like you did in the  `docs/AWS.md` file and probably modify this command in a similar way to have each necessary `aws cli` call.

TrimBiggs:
Made another pass and added some more comments, should be good once those are in.

matthewdupre:
If you ask me, the value gained by endlessly tweaking the formatting of environment variables is less than the value gained by publishing a day earlier.  Particularly considering most of the rest of the doc doesn't conform anyway.  I've made the updates, please feel free to make any further adjustments you'd like yourself.

TrimBiggs:
I think a few of my comments were nitpicky (which I noted) but the majority were important to get in for clarity and uniformity in the docs.  I'm not quite sure what point you're trying to make with your comment, but the better our docs are, the better the user experience is for our customers, and my opinion is that we should make our docs as easy to understand as possible with little hassle.

That being said, this LGTM apart from merge conflicts.

Issue #844, opened 2016-03-04
Cannot start node when IPv6 is not enabled on OS.


robbrockbank:
When starting calicoctl node, exists complaining that IPv4 forwarding could not be enabled.

Looks like this is actually an IPv6 forwarding issue because IPv6 is not enabled on the customers OS.  We need to only try and enable forwarding if IPv6 is configured, and we need to fix the error message.

TrimBiggs:
@robbrockbank Can this be closed now?

robbrockbank:
Fixed by #846

PR #846, opened 2016-03-04
Only enable IPv6 forwarding if IPv6 is enabled

robbrockbank:


robbrockbank:
We could go further with this to, for example, prevent IPv6 configuration, don't start the IPv6 BIRD etc - however, this is a minimum to allow calico node to actually be started.  This also requires a resolution to the Calico issue https://github.com/projectcalico/calico/issues/989 - but that can be handled separately.

calicoctl/calico_ctl/node.py
```diff
@@ -562,10 +563,11 @@ def _setup_ip_forwarding():
...
 
     try:
-        with open('/proc/sys/net/ipv6/conf/all/forwarding', 'w') as f:
-            f.write("1")
+        if ipv6_enabled():
+            with open('/proc/sys/net/ipv6/conf/all/forwarding', 'w') as f:
+                f.write("1")
     except:
-        print "ERROR: Could not enable ipv4 forwarding."
+        print "ERROR: Could not enable ipv6 forwarding."
```
TrimBiggs: We'll have to close the older PR that fixes this.

robbrockbank: yup that was on my radar

TrimBiggs:
Fixes #844

TrimBiggs:
LGTM

PR #848, opened 2016-03-05
Move Calico As a system Service to docs, not libnetwork

TrimBiggs:


robbrockbank:
I think it's worth one sentence at the start of the document indicating that this guide is currently focussed on running the services as Docker containers (just so it's clear up front).  I'm wondering if this guide should just be in the Docker sub section or not and then maybe have a different guides for rkt etc.

TrimBiggs:
I think this should still be in the main `docs` directory since the `calicoctl node` service can be useful for  other deployments.

I added the sentence and am merging shortly

PR #851, opened 2016-03-08
Build a simple binary for use on WIndows/Mac management stations

tomdee:


calicoctl/calico_ctl/diags.py
```diff
@@ -25,7 +25,12 @@
                        [default: docker]
 """
 import sys
-import sh
+from sys import platform as _platform
+
+if _platform == "linux" or _platform == "linux2":
```
caseydavenport: Do we have an issue raised for not requiring `sh`?

calicoctl/calicoctl.py
```diff
@@ -37,6 +35,22 @@
 See 'calicoctl <command> --help' to read about a specific subcommand.
 """
 
+otheros_usage = """Override the host:port of the ETCD server by setting the environment variable
```
caseydavenport: Probably worth a comment explaining why we have two docstrings.

caseydavenport:
Minor comments, then LGTM.

PR #854, opened 2016-03-08
Remove use of "sh" module from diags.py

tomdee:


calicoctl/calico_ctl/diags.py
```diff
@@ -50,19 +45,24 @@ def diags(arguments):
...
 
-    # Check runtime.  
-    runtime = arguments.get("--runtime")
-    if not runtime in ["docker", "rkt"]:
-        print "Invalid runtime specified: '%s'" % runtime
-        sys.exit(1)
 
-    save_diags(arguments["--log-dir"], runtime)
-    sys.exit(0)
+def write_diags(f, command):
```
robbrockbank: I think it would make more sense to make this part of the DiagsFileWriter.

tomdee: But DiagsFileWriter doesn't know what command was run

robbrockbank: I was thinking that DiagsFileWriter could have additional methods for writing and writing and running command.  That way the file object is completely wrapped up in the file writer.

Not too fussed either way tbh.

calicoctl/calico_ctl/diags.py
```diff
@@ -222,6 +121,27 @@ def save_diags(log_dir, runtime="docker"):
...
+        output = subprocess.check_output(command,
+                                         shell=True,
+                                         stderr=subprocess.STDOUT)
+        # Strip out non letters and numbers from the command to form the filename
+        filename = re.sub(r'[^a-zA-Z0-9 -]', "", command)
+
+        # And substitute underscore for spaces
+        filename = re.sub(r'\s', "_", filename)
+
+        with DiagsErrorWriter(temp_diags_dir, filename) as f:
```
robbrockbank: Using the DiagsFileWriter like this doesn't really buy you anything.  The point of that class was to protect against errors in the code writing the diags from trickling up the stack and causing the diags to bail out.  Better that the DiagsErrorWrite context manager encase as much of the processing as possible - especially the sub process call.

tomdee: So shall I just get rid of DiagsErrorWriter?

robbrockbank:
LGTM.  I'm fine with the UTs being removed for now - I agree that they weren't testing anything useful.

PR #861, opened 2016-03-09
Set nf_conntrack_max during calicoctl node

### LINKED ISSUE #855: calicoctl node should set this sysctl "net.netfilter.nf_conntrack_max"
See http://docs.projectcalico.org/en/1.3.0/configuration.html#system-configuration

tomdee:
Fixes #855

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -233,8 +233,8 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
         except OSError:
             pass
 
-        # We will always want to setup IP forwarding
         _setup_ip_forwarding()
+        _set_nf_conntrack_max()
 
     # Print warnings for any known system issues before continuing
         if runtime == 'docker' and not running_in_container():
@@ -558,7 +558,7 @@ def _setup_ip_forwarding():
     try:
         with open('/proc/sys/net/ipv4/ip_forward', 'w') as f:
             f.write("1")
-    except:
+    except Exception:
         print "ERROR: Could not enable ipv4 forwarding."
         sys.exit(1)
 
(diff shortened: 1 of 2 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -570,6 +570,22 @@ def _setup_ip_forwarding():
...
+
+    To avoid this becoming a problem, we recommend increasing the conntrack
+    table size. To do so, run the following commands:
+    """
+    try:
+        with open('/proc/sys/net/netfilter/nf_conntrack_max', 'w') as f:
+            f.write("1000000")
+    except:
+        print "ERROR: Could not set nf_contrack_max"
+        sys.exit(1)
```
caseydavenport: Do we want to fail entirely if we can't do this? Or should we just emit a warning and let it continue?

It's not like we won't work at all in this case.

calicoctl/calico_ctl/node.py
```diff
@@ -570,6 +570,22 @@ def _setup_ip_forwarding():
...
+    run a lot of workloads on a given host, or if your workloads create a lot
+    of TCP connections or bidirectional UDP streams.
+
+    To avoid this becoming a problem, we recommend increasing the conntrack
+    table size. To do so, run the following commands:
+    """
+    try:
+        with open('/proc/sys/net/netfilter/nf_conntrack_max', 'w') as f:
+            f.write("1000000")
+    except:
```
caseydavenport: Do you mean `except Exception` or something even more specific?

caseydavenport:
Couple of comments.

calicoctl/calico_ctl/node.py
tomdee: OK, changed to warning.

caseydavenport:
Failing tests, but otherwise LGTM.

PR #862, opened 2016-03-10
Quit calicoctl node command if docker fails

### LINKED ISSUE #833: calicoctl node errors do not stop it from being run

TrimBiggs:
Fixes #833

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -236,18 +236,18 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
         # We will always want to setup IP forwarding
         _setup_ip_forwarding()
 
-    # Print warnings for any known system issues before continuing
+        # Print warnings for any known system issues before continuing
         if runtime == 'docker' and not running_in_container():
             using_docker = True
         else:
             using_docker = False
 
-        (_, _, etcd_ok) = \
+        (_, docker_ok, etcd_ok) = \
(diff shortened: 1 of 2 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -236,18 +236,21 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
...
 
-    # Print warnings for any known system issues before continuing
+        # Print warnings for any known system issues before continuing
         if runtime == 'docker' and not running_in_container():
             using_docker = True
         else:
             using_docker = False
 
-        (_, _, etcd_ok) = \
+        (_, docker_ok, etcd_ok) = \
```
caseydavenport: This is wrong, I think.

We can't assume that Docker is always running on a host.  `calicoctl node` supports being run on systems where only `rkt` is installed.

I think this needs to be something like:

```
if not (etcd_ok and (docker_ok or runtime == 'rkt')):
    sys.exit(1)
```

caseydavenport: Actually, never mind.  Looks like check_system handles this.

robbrockbank:
One comment then LGTM

PR #865, opened 2016-03-11
Add node show command and allow removing any node from cluster

### LINKED ISSUE #830: Ability to remove a calico BGP peer when using full-mesh
When using full-mesh BGP, the only way to remove a BGP peer from your cluster is like so:
```
calicoctl node stop
calicoctl node remove
```

By default, these commands should be run on the host which you are trying to remove.  However, you can remove another nodes configuration by setting the HOSTNAME variable when removing:

```
HOSTNAME=some-other-host calicoctl node remove
```

This works OK, but it requires that you know the hostname of the node that you're trying to remove (which isn't always the case, especially if the host no longer exists).

We want something like:
```
calicoctl node remove <ip address>
```

and/or a way to detect and clean up BGP peers that no longer exist.

TrimBiggs:
Fixes #830 
- [x] Merge libcalico
- [x] Add unit tests


tomdee:
LGTM

PR #867, opened 2016-03-14
Use NO_DEFAULT_POOLS env var to skip default pool creation

### LINKED ISSUE #866: There is no way to stop Calicoctl from creating a default pool
It would be nice to be able to suppress the default pool creation e.g. with a --no-default-pool flag or an env variable.

tomdee:
Fixes #866

### DIFF
--- a/calico_node/filesystem/startup.py
+++ b/calico_node/filesystem/startup.py
@@ -256,18 +256,19 @@ def main():
     # Verify that the chosen IP exists on the current host
     warn_if_unknown_ip(ip, ip6)
 
-    # Set up etcd
-    ipv4_pools = client.get_ip_pools(4)
-    ipv6_pools = client.get_ip_pools(6)
+    if os.getenv("NO_DEFAULT_POOLS", "").lower() != "true":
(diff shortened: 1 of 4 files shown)


calicoctl/startup.py
```diff
@@ -255,15 +255,16 @@ def main():
...
-    # Set up etcd
-    ipv4_pools = client.get_ip_pools(4)
-    ipv6_pools = client.get_ip_pools(6)
-
-    # Create default pools if required
-    if not ipv4_pools:
-        client.add_ip_pool(4, DEFAULT_IPV4_POOL)
-    if not ipv6_pools and ipv6_enabled():
-        client.add_ip_pool(6, DEFAULT_IPV6_POOL)
+    if not os.getenv("NO_DEFAULT_POOLS"):
```
fasaxc: I'd probably (over?)engineer this to require particular values to be set.  Sometimes it's easier for a wrapper to always set an environment variable but to set it to, say, "true" or "false" dynamically.

robbrockbank:
LGTM

PR #868, opened 2016-03-14
Remove SSL todo as we've upgraded Python

### LINKED ISSUE #682: Upgrade calicoctl Python version for improved security when using Etcd with SSL
When running calicoctl with etcd, urllib3 prints a warning to the screen:
```
InsecurePlatformWarning: A true SSLContext object is not available. This prevents 
urllib3 from configuring SSL appropriately and may cause certain SSL connections 
to fail. For more information, see 
https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning
```

We can fix this by:
1) Upgrading the version of python being used to create the calicoctl binary (currently 2.7.3, upgrade to 2.7.9+).  This involves something like changing the Dockerfile to use `debian:jessie` instead of `debian:wheezy`.  The Kubernetes plugin (and potentially others) also uses `debian:wheezy` and requires some thought

2) Modifying urllib3 to use OpenSSL as its backend ssl module, as described in the URL in the warning above

tomdee:
Fixes #682

### DIFF
--- a/calicoctl/calicoctl.py
+++ b/calicoctl/calicoctl.py
@@ -53,32 +53,27 @@ Usage: calicoctl <command> [<args>...]
 See 'calicoctl <command> --help' to read about a specific subcommand.
 """
 
-import sys
 import signal
+import sys
 import traceback
+from sys import platform as _platform
 
 from docopt import docopt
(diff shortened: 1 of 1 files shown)


TrimBiggs:
LGTM

PR #871, opened 2016-03-14
Reject pools that overlap with existing pools

### LINKED ISSUE #802: Calicoctl Pool allows overlapping pools to be programmed
This currently allows you to program pools such as:
```
+----------------+---------+
|   IPv4 CIDR    | Options |
+----------------+---------+
| 10.10.10.0/24  |         |
| 10.10.10.0/25  |         |
| 10.10.10.0/26  |         |
+----------------+---------+
```

We currently reject all overlapping ranges in the `ip_pool_range_add` function called by `calicoctl pool range add <cidr_start> <cidr_end>`, but not in `ip_pool_add`.

Pools now must start/end at a `/26` prefix, meaning that any cider net smaller than `/26` is not allowed.  So any overlapping pool that is programmed either fully encapsulates the preprogrammed cidr or is fully encapsulated by the preprogrammed cidr.

If a new pool fully encapsulates another pool (such as adding `10.10.0.0/16` when `10.10.10.0/24` exists), we should either
- reject it
- remove old pool and accept new pool (this will require some changes to backend)
- add the not yet programmed remaining cidr(s) as separate pool(s)

If a new pool falls within an existing pool (such as adding `10.10.10.0/24` when `10.10.0.0/16` exists), just reject it.

TrimBiggs:
Fixes #802

### DIFF
--- a/calicoctl/calico_ctl/pool.py
+++ b/calicoctl/calico_ctl/pool.py
@@ -122,19 +122,65 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
         print "IP in IP not supported for IPv6 pools"
         sys.exit(1)
 
-    # TODO Reject any cidrs that overlap with existing cidrs in the pool
-    pools = []
+    current_pools = client.get_ip_pools(version)
+    new_pools = []
+
+    # Ensure new pools are valid and do not overlap with each other or existing
+    # pools.
     for cidr in cidrs:
+
         try:
-            pools.append(IPPool(cidr, ipip=ipip, masquerade=masquerade))
+            pool = IPPool(cidr, ipip=ipip, masquerade=masquerade)
+
         except InvalidBlockSizeError:
             print "An IPv%s pool must have a prefix length of %s or lower." \
                   "\nGiven: %s.\nNo pools added." % \
                   (version, BLOCK_PREFIXLEN[version], cidr)
             sys.exit(1)
-    for pool in pools:
-        client.add_ip_pool(version, pool)
 
+        # Check if new pool overlaps with any existing pool
+        overlapping_pool = _get_overlapping_pool(pool, current_pools)
+        if overlapping_pool:
+            print "Cannot add IP pool %s - pool overlaps with an " \
+                  "existing pool %s" % (cidr, overlapping_pool.cidr)
+            sys.exit(1)
+
+        # Check if this new pool overlaps with any other new pool
+        overlapping_pool = _get_overlapping_pool(pool, new_pools)
+        if overlapping_pool:
+            print "Cannot add IP pool %s - pool overlaps with another " \
+                  "new pool %s" % (cidr, overlapping_pool.cidr)
+            sys.exit(1)
+
+        # Append pool to pending list of new pools to add to Calico
+        new_pools.append(pool)
+
+
+    # Make client call to add each pool to Calico
+    for new_pool in new_pools:
+        client.add_ip_pool(version, new_pool)
+
+
+def _get_overlapping_pool(pool, other_pools):
+    """
+    Check if the given pool overlaps with any pool in the list of other_pools.
+
+    Ignore when a pool's CIDR is an exact match of another pool's CIDR in case
+    a pool is being updated.
+
+    :param pool: IPPool to check overlap for.
+    :param other_pools: List of IPPools to check for overlap.
+    :return: The first IPPool in other_pools that overlaps with pool, or None if
+    no overlap.
+    """
+    for other_pool in other_pools:
+        # Allow the cidr to be exactly the same in case pool is being updated,
+        # for example to add/remove IP-IP support.
+        if ((pool.cidr in other_pool.cidr or other_pool.cidr in pool.cidr) and
+            (pool.cidr != other_pool.cidr)):
+            return other_pool
+
+    return None
 
 def ip_pool_range_add(start_ip, end_ip, version, ipip, masquerade):
     """
--- a/calicoctl/tests/unit/pool_test.py
+++ b/calicoctl/tests/unit/pool_test.py
@@ -54,18 +54,15 @@ class TestPool(unittest.TestCase):
             # Call method under test for each test case
             self.assertEqual(m_sys_exit.called, sys_exit_called)
 
+    @patch("calico_ctl.pool.client", autospec=True)
     @patch("calico_ctl.pool.IPPool", autospec=True)
-    def test_add_bad_pool_size(self, m_IPPool):
+    def test_add_bad_pool_size(self, m_IPPool, m_client):
         """
         Test ip_pool_add exits when pool with bad prefix is passed in.
         """
         m_IPPool.side_effect = InvalidBlockSizeError
-        with patch('sys.exit', autospec=True) as m_sys_exit:
-            # Call method under test
-            pool.ip_pool_add(["10.10.10.10/32"], 4, False, False)
-
-            # Call method under test for each test case
-            self.assertTrue(m_sys_exit.called)
+        self.assertRaises(SystemExit, pool.ip_pool_add, cidrs=["10.10.10.10/32"],
+                          version=4, ipip=False, masquerade=False)
 
     @patch("calico_ctl.pool.client", autospec=True)
     @patch("calico_ctl.pool.IPPool", autospec=True)
@@ -73,14 +70,12 @@ class TestPool(unittest.TestCase):
         """
         Test ip_pool_range_add exits when range with bad prefix is passed in.
         """
-        m_client.get_ip_pools = Mock()
         m_client.get_ip_pools.return_value = []
         m_IPPool.side_effect = InvalidBlockSizeError
         with patch('sys.exit', autospec=True) as m_sys_exit:
             # Call method under test
             pool.ip_pool_range_add("10.10.10.10", "10.10.11.10", 4, False, False)
 
-            # Call method under test for each test case
             self.assertTrue(m_sys_exit.called)
 
     @patch("time.sleep")
@@ -102,3 +97,45 @@ class TestPool(unittest.TestCase):
         self.assertEqual(m_pool.disabled, True)
         m_client.release_pool_affinities.assert_called_once_with(m_pool)
         m_client.remove_ip_pool.assert_called_once_with(4, net1.ip)
+
+    @patch("calico_ctl.pool.client", autospec=True)
+    def test_add_overlapping_existing_pool(self, m_client):
+        """
+        Test ip_pool_add exits when a pool is added that falls within an
+        existing pool.
+        """
+        m_client.get_ip_pools.return_value = [IPPool("10.10.10.0/24")]
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
+            pool.ip_pool_add(cidrs=["10.10.10.0/25"], version=4,
+                             ipip=False, masquerade=False)
+
+            self.assertTrue(m_sys_exit.called)
+
+    @patch("calico_ctl.pool.client", autospec=True)
+    def test_add_overlapping_existing_pool_2(self, m_client):
+        """
+        Test ip_pool_add exits when a pool is added that fully encompasses an
+        existing pool.
+        """
+        m_client.get_ip_pools.return_value = [IPPool("10.10.10.0/26")]
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
+            pool.ip_pool_add(cidrs=["10.10.10.0/24"], version=4,
+                             ipip=False, masquerade=False)
+
+            self.assertTrue(m_sys_exit.called)
+
+    @patch("calico_ctl.pool.client", autospec=True)
+    def test_add_overlapping_new_pools(self, m_client):
+        """
+        Test ip_pool_add exits when two new pools overlap with
+        each other.
+        """
+        m_client.get_ip_pools.return_value = []
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
+            pool.ip_pool_add(cidrs=["10.10.10.0/25", "10.10.10.0/26"],
+                             version=4, ipip=False, masquerade=False)
+
+            self.assertTrue(m_sys_exit.called)


calicoctl/calico_ctl/pool.py
```diff
@@ -118,18 +118,32 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    current_pool_nets = [IPNetwork(pool.cidr) for pool in current_pools]
+
+    new_pools = []
+
+    for new_cidr in cidrs:
+        # Check if new pool overlaps with any existing pool
+        new_cidr_net = IPNetwork(new_cidr)
+        for cur_pool_net in current_pool_nets:
+
+            if cur_pool_net in new_cidr_net or new_cidr_net in cur_pool_net:
```
robbrockbank: probably need to check the supplied pools don't overlap with each other too.

and typo below:  exisitng

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+
+        # Check if new pool overlaps with any other new pool
+        overlapping_new_cidr = _check_if_cidr_overlaps_cidrs(new_cidr,
+                                   [pool.cidr for pool in new_pools])
+        if overlapping_new_cidr:
+            print "Cannot add IP pool %s - pool conflicts with another " \
+                  "new pool %s" % (new_cidr, overlapping_new_cidr)
+            sys.exit(1)
+
+        # Add cidr as a new IP pool
```
djosborne: Can you modify this comment to indicate that they are not actually being added, but instead: `# Add cidr to list of pools to be added`

My original pass through I thought they were being added here, which would be bad as the script would exit out halfway through. It's good that that isn't the case, but bad that the comment makes it seem so

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
         except InvalidBlockSizeError:
             print "An IPv%s pool must have a prefix length of %s or lower." \
                   "\nGiven: %s.\nNo pools added." % \
-                  (version, BLOCK_PREFIXLEN[version], cidr)
+                  (version, BLOCK_PREFIXLEN[version], new_cidr)
             sys.exit(1)
-    for pool in pools:
-        client.add_ip_pool(version, pool)
 
+    for new_pool in new_pools:
```
djosborne: Would also add a comment here indicating `# commit pool addition`

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
 
-    # TODO Reject any cidrs that overlap with existing cidrs in the pool
-    pools = []
-    for cidr in cidrs:
+    current_pools = client.get_ip_pools(version)
+    current_pool_nets = [IPNetwork(pool.cidr) for pool in current_pools]
+
+    new_pools = []
+
+    for new_cidr in cidrs:
```
djosborne: And a comment here, that this whole section is for validation only

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    :param cidrs: List of Cidrs to check for overlap.
+    :return: The first offending overlapping cidr, or None.
+    """
+    cidr_net = IPNetwork(cidr)
+    for other_cidr in cidrs:
+        other_cidr_net = IPNetwork(other_cidr)
+        if cidr_net in other_cidr_net or other_cidr_net in cidr_net:
+            return other_cidr_net.cidr
+
+    return None
```
djosborne: Functions return None by default, so can probably remove this

calico_test/requirements.txt
```diff
@@ -1,4 +1,6 @@
 coverage
+
+
```
djosborne: Can you yank out these new lines?

calicoctl/tests/unit/pool_test.py
```diff
@@ -102,3 +101,40 @@ def test_ip_pool_remove(self, m_client, m_sleep):
...
         m_client.remove_ip_pool.assert_called_once_with(4, net1.ip)
+
+    @patch("calico_ctl.pool.client", autospec=True)
+    def test_add_overlapping_existing_pool(self, m_client):
+        """
+        Test ip_pool_add exits when a pool is added that overlaps with an
+        existing pool.
+        """
+        # Test adding pool that falls within an existing pool
+        m_client.get_ip_pools = Mock()
```
djosborne: If you'd like, you can one line these two lines:
```
m_client.get_ip_pools = Mock(return_value=[IPPool("10.10.10.0/24")])
```

calicoctl/tests/unit/pool_test.py
```diff
@@ -102,3 +101,40 @@ def test_ip_pool_remove(self, m_client, m_sleep):
...
+        m_client.get_ip_pools.return_value = [IPPool("10.10.10.0/24")]
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
+            pool.ip_pool_add(["10.10.10.0/25"], 4, False, False)
```
djosborne: Would be more clear for the next guy if you key value'd these params when calling

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    new_pools = []
+
+    for new_cidr in cidrs:
+        new_cidr_net = IPNetwork(new_cidr)
+
+        # Check if new pool overlaps with any existing pool
+        overlapping_old_cidr = _check_if_cidr_overlaps_cidrs(new_cidr,
+                                   [net.cidr for net in current_pool_nets])
+        if overlapping_old_cidr:
+            print "Cannot add IP pool %s - pool conflicts with exisitng " \
```
djosborne: typo: `exisitng`

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+                  (version, BLOCK_PREFIXLEN[version], new_cidr)
             sys.exit(1)
-    for pool in pools:
-        client.add_ip_pool(version, pool)
 
+    for new_pool in new_pools:
+        client.add_ip_pool(version, new_pool)
+
+
+def _check_if_cidr_overlaps_cidrs(cidr, cidrs):
```
djosborne: This method's purpose might be more clear if it were named `_get_overlapping_cidr`

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
         print "IP in IP not supported for IPv6 pools"
         sys.exit(1)
 
-    # TODO Reject any cidrs that overlap with existing cidrs in the pool
-    pools = []
-    for cidr in cidrs:
+    current_pools = client.get_ip_pools(version)
```
djosborne: Lots of variables with similar names and purposes floating around here. Since this guy is only used to create `current_pool_nets`, why not just overwrite current_pools?
```
current_pools = client.get_ip_pools(version)
current_pools = [IPNetwork(pool.cidr) for pool in current_pools]
```

Better yet, could one line it:
```
current_pools = [IPNetwork(pool.cidr) for pool in client.get_ip_pools(version)]
```

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    current_pool_nets = [IPNetwork(pool.cidr) for pool in current_pools]
+
+    new_pools = []
+
+    for new_cidr in cidrs:
+        new_cidr_net = IPNetwork(new_cidr)
```
djosborne: Same thing here, could just overwrite it - `cidr = IPNetwork(cidr)`

djosborne:
@TrimBiggs Any thoughts on merging the functionality that's now implemented separately by both  `ip_pool_range_add` and `ip_pool_add`?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+
+        # Check if new pool overlaps with any existing pool
+        overlapping_old_cidr = _check_if_cidr_overlaps_cidrs(new_cidr,
+                                   [net.cidr for net in current_pool_nets])
```
robbrockbank: Isn't `[net.cidr for net in current_pool_nets]` just the same as `current_pool_nets`?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+        if overlapping_old_cidr:
+            print "Cannot add IP pool %s - pool conflicts with exisitng " \
+                  "pool %s" % (new_cidr, overlapping_old_cidr)
+            sys.exit(1)
+
+        # Check if new pool overlaps with any other new pool
+        overlapping_new_cidr = _check_if_cidr_overlaps_cidrs(new_cidr,
+                                   [pool.cidr for pool in new_pools])
+        if overlapping_new_cidr:
+            print "Cannot add IP pool %s - pool conflicts with another " \
```
robbrockbank: I think the word conflicts is a little vague.

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
         print "IP in IP not supported for IPv6 pools"
         sys.exit(1)
 
-    # TODO Reject any cidrs that overlap with existing cidrs in the pool
-    pools = []
-    for cidr in cidrs:
+    current_pools = client.get_ip_pools(version)
+    current_pool_nets = [IPNetwork(pool.cidr) for pool in current_pools]
```
robbrockbank: Isn't pool.cidr already an IPNetwork?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    for new_pool in new_pools:
+        client.add_ip_pool(version, new_pool)
+
+
+def _check_if_cidr_overlaps_cidrs(cidr, cidrs):
+    """
+    Check if a cidr overlaps at all with any cidrs in the list.
+
+    :param cidr: Cidr to check overlap for.
+    :param cidrs: List of Cidrs to check for overlap.
```
robbrockbank: CIDR

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,58 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
 
+    for new_pool in new_pools:
+        client.add_ip_pool(version, new_pool)
+
+
+def _check_if_cidr_overlaps_cidrs(cidr, cidrs):
+    """
+    Check if a cidr overlaps at all with any cidrs in the list.
+
+    :param cidr: Cidr to check overlap for.
```
robbrockbank: It seems like these parameters are a mix of strings and IPNetworks - better to have everything the same.  I'd stick to passing in IPNetworks - then you don't have to convert in this function.

calicoctl/calico_ctl/pool.py
robbrockbank: I actually couldn't see new_cidr_net being referenced anywhere.

calicoctl/calico_ctl/pool.py
TrimBiggs: I think it's better to be explicit with a return value.

robbrockbank:
A couple of other things to think about:

-  When displaying an error message - do you want to display the normalized or the entered CIDR.  My expectation would be the latter - although perhaps not the end of the world if we display the normalized one.

-  I think you need to allow the same pool to be re-added.  It's not ideal, but at the moment we are able to do this and it basically just updates the existing pool.

calicoctl/calico_ctl/pool.py
TrimBiggs: pool.cidr is a CIDR string from IPNetwork.cidr:
`self.cidr = IPNetwork(cidr).cidr`

calicoctl/calico_ctl/pool.py
TrimBiggs: This was passing in CIDR string values for this param and the list of CIDRs.  I think this makes sense since we have cidr strings passed in and the pool.cidr value is a string cidr.

calicoctl/tests/unit/pool_test.py
```diff
@@ -102,3 +101,40 @@ def test_ip_pool_remove(self, m_client, m_sleep):
...
+        m_client.get_ip_pools = Mock(return_value=[IPPool("10.10.10.0/24")])
+        with patch('sys.exit', autospec=True) as m_sys_exit:
+            # Call method under test
+            pool.ip_pool_add(cidrs=["10.10.10.0/25"], version=4,
+                             ipip=False, masquerade=False)
+
+            self.assertTrue(m_sys_exit.called)
+
+        # Test adding pool that holds existing pool.
+        m_client.get_ip_pools.return_value = [IPPool("10.10.10.0/26")]
```
djosborne: Should these be separate tests?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,56 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    """
+    Check if the given cidr overlaps with any CIDR in the cidrs list.
+
+    :param cidr: CIDR to check overlap for.
+    :param cidrs: List of CIDRs to check for overlap.
+    :return: The first CIDR in cidrs that overlaps with cidr, or None.
+    """
+    cidr_net = IPNetwork(cidr)
+    for other_cidr in cidrs:
+        other_cidr_net = IPNetwork(other_cidr)
```
djosborne: I'm finding this method a bit hard to follow. One thing that's raising a flag for me is that `other_cidr` is from `cidrs` which is already a collection of `IPNetwork`s. So you're constructing an IPNetwork instance from an IPNetwork instance, which doesn't seem efficient...

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,56 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+
+def _get_overlapping_cidr(cidr, cidrs):
+    """
+    Check if the given cidr overlaps with any CIDR in the cidrs list.
+
+    :param cidr: CIDR to check overlap for.
+    :param cidrs: List of CIDRs to check for overlap.
+    :return: The first CIDR in cidrs that overlaps with cidr, or None.
+    """
+    cidr_net = IPNetwork(cidr)
```
djosborne: can we overwrite this as well?
`cidr = IPNetwork(cidr)`?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,56 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
             sys.exit(1)
-    for pool in pools:
-        client.add_ip_pool(version, pool)
 
+    # Make client call to add each pool to Calico
+    for new_pool in new_pools:
+        client.add_ip_pool(version, new_pool)
+
+
+def _get_overlapping_cidr(cidr, cidrs):
```
djosborne: Maybe change `cidrs` => `other_cidrs`, so its easier to follow where `other_cidr` came from?

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,56 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+
+    :param cidr: CIDR to check overlap for.
+    :param cidrs: List of CIDRs to check for overlap.
+    :return: The first CIDR in cidrs that overlaps with cidr, or None.
+    """
+    cidr_net = IPNetwork(cidr)
+    for other_cidr in cidrs:
+        other_cidr_net = IPNetwork(other_cidr)
+        if ((cidr_net in other_cidr_net or other_cidr_net in cidr_net) and
+            not (cidr_net == other_cidr_net)):
```
djosborne: Why are we ignoring the equal case? I think I may have saw a comment from rob that its because we allow it through as an "update", right? If so, definitely need a comment explaining that, but more importantly, the name and description of this function suddenly has a loophole. It may be best to move the equal check out to the function that calls this. Might need to think about this a bit...

calicoctl/calico_ctl/pool.py
TrimBiggs: `cidrs` is not a list of IPNetworks, it is a list of CIDRs.  I've changed the name of this to be `other_cidrs`.

TrimBiggs: Removed old comment. Fixing up naming now.

calicoctl/calico_ctl/pool.py
TrimBiggs: I've added a comment about this.  I think having it in this function with a comment is alright since equality is a slightly different check than overlap here.

Alternatively we can return all CIDRs that overlap instead of short circuiting with the first CIDR, but I'm not fussed to change it. What do you think?

calicoctl/calico_ctl/pool.py
robbrockbank: Are you sure?

calicoctl/tests/unit/pool_test.py
```diff
@@ -102,3 +101,40 @@ def test_ip_pool_remove(self, m_client, m_sleep):
...
         m_client.remove_ip_pool.assert_called_once_with(4, net1.ip)
+
+    @patch("calico_ctl.pool.client", autospec=True)
+    def test_add_overlapping_existing_pool(self, m_client):
+        """
+        Test ip_pool_add exits when a pool is added that overlaps with an
+        existing pool.
+        """
+        # Test adding pool that falls within an existing pool
+        m_client.get_ip_pools = Mock(return_value=[IPPool("10.10.10.0/24")])
```
robbrockbank: m_client is already a Mock no?  In which case I think you can simplify this as:

`m_client.get_ip_pools.return_value = blah`

calicoctl/calico_ctl/pool.py
robbrockbank: Yeah beefed up comment should be good - that'd keep me happy :-)

TrimBiggs:
Ready for another round.

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,61 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
         sys.exit(1)
 
-    # TODO Reject any cidrs that overlap with existing cidrs in the pool
-    pools = []
+    current_pool_nets = [pool.cidr for pool in client.get_ip_pools(version)]
+    new_pools = []
+
+    # Validate pool CIDRs
     for cidr in cidrs:
+        cidr_net = IPNetwork(cidr)
```
djosborne: `cidr` and `cidr_net` are still being thrown around interchangeably, which is making things harder to follow. I tested and confirmed this to be true:
```
str(IPNetwork(cidr)) == cidr
```

This means that your log messages can `%s` in IPNetwork instances and get the same effect as if they were fed string representation instead.

Therefore, I think you should update this line to:
`cidr = IPNetwork(cidr)`

From scanning over the code, you will not need the string verion of the cidr at all. Additionally, when you construct an IPPool using the cidr on line 148, the init method for IPPool can handle both strings and IPNetwork objects, so it to will handle the IPNetwork object just fine

TrimBiggs: I was under the impression (as per Rob's comment above) that the user may specify the `cidr` as `10.10.10.10/24` whereas `str(IPNetwork("10.10.10.10/24"))` would return `10.10.10.0/24`, but I just checked and this is not true.

djosborne: `str(IPNetwork("10.10.10.10/24").cidr)` *would* return `10.10.10.0/24`. Therefore, right before this line, should you also enforce that the user is specifiying a valid by throwing an error if `IPNetwork(cidr) != IPNetwork(cidr).cidr`?

TrimBiggs: It's not invalid to pass in `10.10.10.10/24`, so I don't think we should fail it.

Also: `IPNetwork("10.10.10.10/24") == IPNetwork("10.10.10.0/24")`

TrimBiggs:
Just chatted with @robbrockbank.  To avoid confusion, I'm going to convert the input to a pool and compare the pools instead of cidrs/networks since the naming schemes make it difficult to follow the type of object.  Update coming shortly...

TrimBiggs:
Updated.

calicoctl/calico_ctl/pool.py
```diff
@@ -122,19 +122,65 @@ def ip_pool_add(cidrs, version, ipip, masquerade):
...
+    :param pool: IPPool to check overlap for.
+    :param other_pools: List of IPPools to check for overlap.
+    :return: The first IPPool in other_pools that overlaps with pool, or None if
+    no overlap.
+    """
+    for other_pool in other_pools:
+        # Allow the cidr to be exactly the same in case pool is being updated,
+        # for example to add/remove IP-IP support.
+        if ((pool.cidr in other_pool.cidr or other_pool.cidr in pool.cidr) and
+            not (pool.cidr == other_pool.cidr)):
```
robbrockbank: what's wrong with `pool.cidr != other_pool.cidr`  ?

TrimBiggs: heh..

robbrockbank:
I'm good with this now - once tests pass.

PR #874, opened 2016-03-15
Ignore flannel bridge when getting node IP

### LINKED ISSUE #814: Flannel0 bridge IP gets autodetected by calicoctl node as the public IP
It should be added to the ignore list, or even better, why not ignore all bridges?

TrimBiggs:
Fixes #814


tomdee:
LGTM

PR #875, opened 2016-03-15
Make startup.py a standalone script

tomdee:
- It depends on libcalico but nothing else (...well netaddr too)


calico_node/filesystem/startup.py
```diff
@@ -228,10 +244,6 @@ def main():
         print "AS environment (%s) is not a AS number." % as_num
         sys.exit(1)
 
-    # Check etcd connectivity first.
-    if not check_etcd_version():
```
robbrockbank: Why are we removing this?

calico_node/filesystem/startup.py
```diff
@@ -262,7 +274,7 @@ def main():
     # Create default pools if required
     if not ipv4_pools:
         client.add_ip_pool(4, DEFAULT_IPV4_POOL)
-    if not ipv6_pools and ipv6_enabled():
+    if not ipv6_pools and os.path.exists('/proc/sys/net/ipv6'):
```
robbrockbank: I prefer either having the ipv6_enabled() method or at least a comment to demystify the path exists statement.

robbrockbank:
Tom, I'm fine for this to go in to stop things being so broken, but I haven't really had the time to convince myself the overall approach is all lovely.  I can't immediately think of a better approach though.

tomdee:
OK, merging and will fix up more tomorrow

PR #878, opened 2016-03-18
Rkt and multihost fixes for startup script

tomdee:


tests/st/no_orchestrator/test_mainline_multi_host.py
```diff
@@ -23,7 +24,15 @@ def test_multi_host(self):
         Test mainline functionality without using an orchestrator plugin on
         multiple hosts.
         """
-        with DockerHost('host1') as host1, DockerHost('host2') as host2:
+        with DockerHost('host1') as host1, DockerHost('host2', start_calico=False) as host2:
+            # Start calico manually on host2
+            host2.execute("docker run -d --net=host --privileged "
+                          "--name=calico-node -e IP=%s "
+                          "-e ETCD_AUTHORITY=%s:2379 -e ETCD_SCHEME=http "
```
robbrockbank: This won't work for SSL tests which is a bit naughty.  I realize that SSL tests are bust at the moment in general, but surely we can add the placeholder code here?

tomdee: Good point - I'll work on a fix.

robbrockbank:
LGTM once you fix the SSL comment

PR #881, opened 2016-03-22
Reinstate SSL STs in semaphore with saved cfssl binaries

TrimBiggs:


TrimBiggs:
Looks like this is failing due to a ST change when the SSL STs were disabled . Looking into it.

TrimBiggs:
This is passing now. @tomdee, mind giving it another look?

tomdee:
LGTM

PR #882, opened 2016-03-23
Add support for ETCD_ENDPOINTS

matthewdupre:


calicoctl/calico_ctl/checksystem.py
```diff
@@ -224,8 +227,11 @@ def check_etcd_version():
         detected_version_string_raw = \
             client.etcd_client.api_execute("/version", 'GET').data
     except EtcdConnectionFailed:
+        etcd_env = os.getenv(ETCD_ENDPOINTS_ENV, "")
```
tomdee: Pretty sure you don't need the `, ""` here

calicoctl/calico_ctl/connectors.py
```diff
@@ -25,15 +25,6 @@
...
-# If an ETCD_AUTHORITY is specified in the environment variables, validate
-# it.
-etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
-if etcd_authority and not validate_hostname_port(etcd_authority):
-    print_paragraph("Invalid %s. It must take the form <address>:<port>. "
-                    "Value provided is '%s'" % (ETCD_AUTHORITY_ENV,
-                                                etcd_authority))
-    sys.exit(1)
-
 try:
```
tomdee: You tested this right?

calicoctl/calico_ctl/node.py
```diff
@@ -262,8 +262,10 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
     if not os.path.exists(log_dir):
         os.makedirs(log_dir)
 
-    # The format of the authority string has already been validated.
+    # The format of the authority and endpoints strings have already been
+    # validated.
     etcd_authority = os.getenv(ETCD_AUTHORITY_ENV, ETCD_AUTHORITY_DEFAULT)
+    etcd_endpoints = os.getenv(ETCD_ENDPOINTS_ENV, None)
```
tomdee: You don't need the `,None` - it's the default

calico_node/filesystem/etc/rc.local
```diff
@@ -27,12 +27,22 @@ else
 
     # Run confd twice.  Our confd TOML files are also generated from confd, so
     # running twice ensures our starting configuration is correct.
-    confd -confdir=/etc/calico/confd -onetime -node=${ETCD_SCHEME:=http}://${ETCD_AUTHORITY} \
+    case "$ETCD_ENDPOINTS" in
```
tomdee: I don't get this craziness. Why are you pulling out the ETCD_SCHEME? And why the two different cases?

calico_node/filesystem/etc/service/confd/run
```diff
@@ -2,7 +2,18 @@
 exec 2>&1
 sed "s/HOSTNAME/$HOSTNAME/" /etc/calico/confd/templates/bird_aggr.toml.template > /etc/calico/confd/conf.d/bird_aggr.toml
 sed "s/HOSTNAME/$HOSTNAME/" /etc/calico/confd/templates/bird6_aggr.toml.template > /etc/calico/confd/conf.d/bird6_aggr.toml
+# ETCD_ENDPOINTS overrides ETCD_SCHEME / ETCD_AUTHORITY, if present
+case "$ETCD_ENDPOINTS" in
```
tomdee: Again, I'm confused by this.

tomdee:
OK, I've finished looking at this - I'll take another look after markups

calicoctl/calico_ctl/checksystem.py
matthewdupre: I don't care, so I've deleted it; but in pretty much all the other places we call this function we pass "" in.

calicoctl/calico_ctl/node.py
TrimBiggs: I think this needs to be `""`.  When added to the `etcd_envs` below, it will add it as `"None"`, rather than `""`, then Felix may interpret `"None"` as a path and fail.

matthewdupre: We pass , None in many other places.  I actually prefer leaving it; I think it's faster to read with the Nones than forcing people to pull the default from memory (/ the internet).

calico_node/filesystem/etc/rc.local
matthewdupre: There are two cases because that's the simplest way to set ETCD_SCHEME when endpoints is in use.  I used a case, not something shorter because it's not specific to any shell, and is quite easy to understand even if you aren't a *sh expert.

That leaves the question of why do we need ETCD_SCHEME.  Well, currently we pass it to confd separately (-scheme).

I've looked at the confd code, and -scheme is used to get the nodes from DNS.  We don't do this, so I think we can just retire -scheme.

TrimBiggs: IIRC, the `-scheme` was necessary for passing in to confd.  If you make this change, be sure to test it.  I'd expect it to fail.

matthewdupre: I haven't got to that yet, but looking at the confd source code it only seems to be used when looking the nodes up (i.e. -node is empty).

TrimBiggs:
We should have docs explaining this change, or an issue should be raised to add docs.

tomdee:
LGTM - pls hit merge when the STs are passing.

PR #893, opened 2016-03-24
Move specific release version replacements above general ones and improve escaping

matthewdupre:
simple-binary target doesn't work on release candidates, because it looks for libcalico.git@v0.13.0@master.  This is changed wrongly by the release scripts from libcalico.git@master: they try to do the right thing, but the specific libcalico@master regex falls further down the list than the generic libcalico one, so does nothing.


tomdee:
What would the issue? Just escaping? What about just putting `r` at the start?

matthewdupre:
Main issue is the ordering.  They run top to bottom, but there was a general one that did the bad thing above the specific one that would do the right thing (escaping permitting).

tomdee:
LGTM - can you stick in a comment about the ordering and fix up the use of r'' if you want.

PR #917, opened 2016-04-12
Allow felix to load its own kernel modules

tomdee:


matthewdupre:
This part all looks fine.

I took a glance at Felix to see what modules it loads, and it looks like it will load conntrack, but I didn't see an obvious ip6tables load.

PR #918, opened 2016-04-12
Fix stars URLs.

caseydavenport:


matthewdupre:
@caseydavenport is this good to go in?

caseydavenport:
No, not yet - still needs to be tested.  As far as I know the current URLs work with the current Vagrant cluster setup, but not with other clusters.  This shouldn't block a release.

PR #922, opened 2016-04-18
Add guide for installing Calico with DCOS

TrimBiggs:


docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+modifying `mem-limit-framework`, `mem-limit-install`,
+`mem-limit-etcd-proxy`, `mem-limit-node`, and
+`mem-limit-libnetwork`.
+
+![alt tag](../images/mesos/dcos-calico-config-changes.png)
+
+You could also speed up the install by increasing
+`max-concurrent-restarts`, allowing Calico to restart
+the Mesos agent and Docker services on more than one
+agent at a time.
```
robbrockbank: You should probably addd a big fat warning about this though.  Not generally recommended - but fine if you're turning up a new system.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+Calico will now begin installation on your cluster.
+
+### Check Calico Status
+It will take a few minutes for Calico to finish
+installing on your cluster.
+
+Calico runs the following tasks on each agent:
+ - Install Docker multi-host networking
+ - Install Mesos net-modules
+ - Configure an Etcd proxy
```
robbrockbank: It actually configures the proxy first, so probably worth moving to the top.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+   (note that it may take a few minutes for Calico
+    to appear).
+ - Once the Calico service is `Healthy`, click the
+   square pop-out icon next to the service name to
+   open the Calico status page in a new tab.
+
+![alt tag](../images/mesos/dcos-calico-open-status.png)
+
+If the install is still in its early stages, the
+webpage may fail to load since the Calico service
```
robbrockbank: The issue here is more that the install restarts Docker and the Agent - which will cause a restart of the framework when it happens on the same agent.  Should be nothing to worry about and the web page will come back online once the framework is restarted.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+ - Go to the **Universe** tab
+ - Find the Calico package
+ - Click the **Install** button then choose **Advanced
+   Installation** in the pop-up window.
+
+![alt tag](../images/mesos/dcos-calico-package-install.png)
+
+##### At this point you have a few options:
+
+In non-production environments used for testing,
```
robbrockbank: I'd probably say that in non-production environments used for testing, you might have extremely limited resources available on each host (this is certainly the case for the default DC/OS vagrant installation) blah blah blah

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+
+##Prerequisites
+
+### Mesos Master and Agents
+
+You will need to be running a cluster with at least one Mesos master
+and and at least one Mesos agent (though we recommend having two or
+more agents).
+
+We currently support Mesos version `v0.28.0` and on for use with DCOS.
```
robbrockbank: and on ?

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+Additionally, we support Centos7 as an operating system for use with
+the Unified Containerizer and Docker Containerizer, though additional
+OS's *may* work (though have not been fully tested!) with the Docker
```
robbrockbank: They won't work since we don't install net modules in those cases.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+##Prerequisites
+
+### Mesos Master and Agents
+
+You will need to be running a cluster with at least one Mesos master
+and and at least one Mesos agent (though we recommend having two or
+more agents).
+
+We currently support Mesos version `v0.28.0` and on for use with DCOS.
+Additionally, we support Centos7 as an operating system for use with
```
robbrockbank: The word additionally sounds very odd in this instance.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+the Unified Containerizer and Docker Containerizer, though additional
+OS's *may* work (though have not been fully tested!) with the Docker
+Containerizer.
+
+### Etcd
+Calico uses etcd as its datastore.
```
robbrockbank: That's a short paragraph.  I think it's worth covering the following:
-  Calico requires an etcd datastore.
-  A service discovery URL is required to access the datastore - which can be configured in the Calico service advanced options.
-  You can quickly spin up an etcd cluster using the Universe package as follows.  By default, the Calico service is configured to use the default service discovery URL used by this etcd service (in other words - if you install etcd as below, you don't need to edit the service discovery url in the Calico config).

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+
+You can check the status of etcd by visiting the **Services**
+tab. You can install Calico once etcd appears with a status
+of **`Healthy`** (note that this could take a few minutes).
+
+## Install Calico
+You can easily install Calico from within the DCOS UI.
+
+### Get Calico Package
+You will need to link the Calico package from our git
```
robbrockbank: Not any more.  It is in the main Universe packages.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+
+ - Go to the **Services** tab
+ - Find Calico in the list of running services
+   (note that it may take a few minutes for Calico
+    to appear).
+ - Once the Calico service is `Healthy`, click the
+   square pop-out icon next to the service name to
+   open the Calico status page in a new tab.
+
+![alt tag](../images/mesos/dcos-calico-open-status.png)
```
robbrockbank: would be nice to display the status page too.

docs/mesos/DCOS.md
```diff
@@ -0,0 +1,134 @@
...
+ - Find the Calico package
+ - Click the **Install** button then choose **Advanced
+   Installation** in the pop-up window.
+
+![alt tag](../images/mesos/dcos-calico-package-install.png)
+
+##### At this point you have a few options:
+
+In non-production environments used for testing, 
+you can set each memory limit option to have a
```
robbrockbank: We should also somewhere suggest pointing them at the DCOS vagrant installer to try a test cluster.

robbrockbank:
Tim, despite the large number of comments, this doc is fine and could go in once you address things that need to be changed (such as the repo), and then add more details/tweaks at a later stage.

docs/mesos/DCOS.md
robbrockbank: although add a note that Docker is old (see https://github.com/dcos/dcos-vagrant/issues/39)  therefore the calico service will not enable Calico on Docker unless you first upgrade Docker _before_ launching Calico service.

docs/mesos/DCOS.md
TrimBiggs: Yup, couldn't get a screenshot by the time I wrote this.

docs/mesos/DCOS.md
TrimBiggs: Yes, I was planning to add that after the release

docs/mesos/DCOS.md
TrimBiggs: changed to "or later"

PR #935, opened 2016-04-22
Remove reference to DOCKER_HOST from troubleshooting guide

TrimBiggs:


robbrockbank:
LGTM - so happy for this to be merged.

Suggestion though - rather than removing completely, we could rework to talk about environment variables when running calicoctl (since calicoctl node requires sudo, so same logic applies).

TrimBiggs:
Partially fixes #935

PR #938, opened 2016-04-25
Fixed calico start failure bug

djosborne:
Fixed bug where calicoctl node would fail to start up without being passed --ip, due to a failure to convert ipaddress to a string

### DIFF
--- a/calico_node/filesystem/startup.py
+++ b/calico_node/filesystem/startup.py
@@ -247,7 +247,7 @@ def main():
                                     "virbr.*", "lxcbr.*", "veth.*",
                                     "cali.*", "tunl.*", "flannel.*"])
         try:
-            ip = ips.pop()
+            ip = str(ips.pop())
         except IndexError:
             print "Couldn't autodetect a management IP address. Please " \
                   "provide an IP address by rerunning the container with the" \
--- a/calico_test/tests/st/utils/docker_host.py
+++ b/calico_test/tests/st/utils/docker_host.py
@@ -28,14 +28,29 @@ class DockerHost(object):
     """
     A host container which will hold workload containers to be networked by
     Calico.
+
+    :param calico_node_autodetect_ip: When set to True, the test framework
+    will not perform IP detection, and will run `calicoctl node` without explicitly
+    passing in a value for --ip. This means calico-node will be forced to do its IP detection.
     """
     def __init__(self, name, start_calico=True, dind=True,
                  additional_docker_options="",
                  post_docker_commands=["docker load -i /code/calico-node.tar",
-                                       "docker load -i /code/busybox.tar"]):
+                                       "docker load -i /code/busybox.tar"],
+                 calico_node_autodetect_ip=False):
         self.name = name
         self.dind = dind
         self.workloads = set()
+        self.ip = None
+        """
+        An IP address value to pass to calicoctl as `--ip`. If left as None, no value will be passed,
+        forcing calicoctl to do auto-detection.
+        """
+
+        self.ip6 = None
+        """
+        An IPv6 address value to pass to calicoctl as `--ipv6`. If left as None, no value will be passed.
+        """
 
         # This variable is used to assert on destruction that this object was
         # cleaned up.  If not used as a context manager, users of this object
@@ -67,7 +82,8 @@ class DockerHost(object):
                                 retries=10)
             for command in post_docker_commands:
                 self.execute(command)
-        else:
+        elif not calico_node_autodetect_ip:
+            # Find the IP so it can be specified as `--ip` when launching node later.
             self.ip = get_ip(v6=False)
             self.ip6 = get_ip(v6=True)
 
@@ -132,14 +148,11 @@ class DockerHost(object):
         :param as_num: The AS Number for this node.  A value of None uses the
         inherited default value.
         """
-        args = ['node', '--ip=%s' % self.ip]
-        try:
-            if self.ip6:
-                args.append('--ip6=%s' % self.ip6)
-        except AttributeError:
-            # No ip6 configured
-            pass
-
+        args = ['node']
+        if self.ip:
+            args.append('--ip=%s' % self.ip)
+        if self.ip6:
+            args.append('--ip6=%s' % self.ip6)
         args.append(options)
 
         cmd = ' '.join(args)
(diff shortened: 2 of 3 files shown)


calico_node/filesystem/startup.py
```diff
@@ -261,7 +261,7 @@ def main():
     # This is required because the confd templates expect to be able to fill in
     #  the IP by fetching it from the environment.
     with open('startup.env', 'w') as f:
-        f.write("IP=" + ip)
+        f.write("IP=" + str(ip))
```
TrimBiggs: I think this should be done at line 250.  All of the calls below should be expecting the string version of the `ip`. The calls below expect string IPs as well, but again, we don't have strong testing for this in STs.

tests/st/no_orchestrator/test_mainline_single_host.py
```diff
@@ -17,6 +17,39 @@
...
+
+            # Test the teardown commands
+            host.calicoctl("profile remove TEST_GROUP")
+            host.calicoctl("container remove %s" % workload1)
+            host.calicoctl("container remove %s" % workload2)
+            host.calicoctl("pool remove 192.168.0.0/16")
+            host.calicoctl("node stop")
+            host.calicoctl("node remove")
+
+
```
TrimBiggs: extra line

calico_node/filesystem/startup.py
djosborne: I think that doing it at line 250 will  only bandaid the fact that we now receive `IPAddress` objects from `get_host_ips`, but all our calico-containers functions expect strings to be passed around. We should either be getting a string from `get_host_ips`, or all the below functions should be updated to take `IPAddress` objects

I could use some input on that, @robbrockbank @TrimBiggs ?

djosborne: I would convert everything to handle IPAddresses but I have speed and memory concerns if we're constantly encapsulating strings in the larger objects. Maybe we can take the hit considering our python binary isn't focused on either of those currently...

TrimBiggs: There was a recent change to specifically have `get_host_ips` return IPAddress objects, but this call was missed during the update.

I'm not opposed to having the other functions use IPAddress objects (it's not a large change). I don't think it will have much of an effect on speed/memory and I don't think we really care that much for this call if it did, this is a one-time run to start the node.

calico_test/tests/st/utils/docker_host.py
```diff
@@ -132,7 +136,9 @@ def start_calico_node(self, options=""):
         :param as_num: The AS Number for this node.  A value of None uses the
         inherited default value.
         """
-        args = ['node', '--ip=%s' % self.ip]
+        args = ['node']
+        if self.ip:
+            args.append('--ip=%s' % self.ip)
         try:
             if self.ip6:
                 args.append('--ip6=%s' % self.ip6)
```
robbrockbank: We don't need to catch AttributeError anymore - looks like that was a horrid hack because we didn't initialize ip6 which we do now.

tests/st/no_orchestrator/test_mainline_single_host.py
```diff
@@ -17,6 +17,39 @@
...
+        with DockerHost('host', dind=False, autodetect_ip=False) as host:
+            host.calicoctl("status")
+            host.calicoctl("profile add TEST_GROUP")
+
+            # Create a workload on each host.
+            workload1 = host.create_workload("workload1")
+            workload2 = host.create_workload("workload2")
+
+            # Add the nodes to Calico networking.
+            host.calicoctl("container add %s 192.168.1.1" % workload1)
```
robbrockbank: This isn't testing auto.

tests/st/no_orchestrator/test_mainline_single_host.py
```diff
@@ -17,6 +17,39 @@
 from tests.st.utils.docker_host import DockerHost
 
 class TestNoOrchestratorSingleHost(TestBase):
+    def test_single_host_auto_ipv4(self):
```
robbrockbank: Maybe change to auto detect -- i originally read this as auto-assign.

calico_test/tests/st/utils/docker_host.py
robbrockbank: Ugh - I stand corrected - it looks like we still don't initialize the ip6 in the dind branch - can we just make sure we always assign ip6, so we don't have to catch an attribute error.

calico_node/filesystem/startup.py
djosborne: True, last time I remember it being a concern was in libcalico when encapsulating etcd endpoints, which is obviously a much higher scale. Since this is oneshot, I agree

tests/st/no_orchestrator/test_mainline_single_host.py
robbrockbank: Hmm, actually this needs some docs or something to clarify.  There are two bits of auto detection going on here: (1) auto-detecting in the test framework so that we pass the IPv4 address through into calicoctl, and the other is calicoctl auto detecting the IPv4 address when it is not supplied.   It's not immediately obvious which this is without sifting through the code.

calico_node/filesystem/startup.py
robbrockbank: It is nicer to use IPAddress consistently when possible - but might be sensible to be pragmatic here since the other methods are not part of a public API.

caseydavenport:
Am I correct in thinking this is only an issue in master, not in v0.19.0?

TrimBiggs:
@caseydavenport Correct, this is only on master.

calico_test/tests/st/utils/docker_host.py
djosborne: What dind branch?

calico_test/tests/st/utils/docker_host.py
```diff
@@ -28,14 +28,30 @@ class DockerHost(object):
...
         self.workloads = set()
+        self.ip = None
+        """
+        An IP address value to pass to calicoctl as `--ip`. If left as None, no value will be passed,
+        forcing calicoctl to do auto-detection.
+        """
+
+        self.ip6 = None
+        """
+        An IPv6 address value to pass to calicoctl as `--ipv6`. If left as None, no value will be passed,
```
robbrockbank: Actually we only auto-detect IPv4 - so if this is left as none we don't configure the node to use IPv6.

robbrockbank:
One minor comment then LGTM.

PR #945, opened 2016-04-26
Use correct IP type in node startup with get_host_ips change

TrimBiggs:
`get_host_ips` now returns an IPAddress object, but this function is expecting a string IP.


robbrockbank:
lgtm, but i'm wondering why our tests didn't hit this problem - any idea?

djosborne:
There's no ST currently which does auto-detect for IP. I added one and fixed this bug in #938

TrimBiggs:
Closing in favor of #938

PR #948, opened 2016-04-27
Fix Mesos Vagrant Demo

djosborne:
Moved docker provisioner line in mesos vagrant so mesos doesn't get upgraded

### DIFF
--- a/docs/mesos/vagrant-centos/Vagrantfile
+++ b/docs/mesos/vagrant-centos/Vagrantfile
@@ -110,6 +110,9 @@ Vagrant.configure("2") do |config|
       # Selinux => permissive
       host.vm.provision :shell, inline: "setenforce permissive"
 
+      # Install docker here, as otherwise it will run 'yum update' and unexpectedly ugrade mesos
+      host.vm.provision :docker
+
       # Add official Mesos Repos and install Mesos.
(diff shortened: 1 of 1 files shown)


docs/mesos/vagrant-centos/Vagrantfile
```diff
@@ -110,6 +110,9 @@ Vagrant.configure("2") do |config|
       # Selinux => permissive
       host.vm.provision :shell, inline: "setenforce permissive"
 
+      # Install docker here, as otherwise it will run 'yum update' and unexpectedly ugrade mesos
+      host.vm.provision :docker
```
TrimBiggs: Why not provision the images here as well?

TrimBiggs:
One comment, then LGTM

docs/mesos/vagrant-centos/Vagrantfile
djosborne: Downloading Docker Images is split in an `if` statment so we can download correct images for master or agent. Instead of messing with the flow down there and up here by introducing if statemnts, Its simpler to just handle the docker installation here, and allow the image downloading to happen below as normal. The docker provisioner is smart enough to not reattempt any of the docker installation, and it translates into just a `docker pull` down there

TrimBiggs: Calico-node and libnetwork are installed on all machines (master and agents). marathon-lb is the only one that is not on all machines.

PR #962, opened 2016-05-04
Exit with status 1 when node container quits with bad status

### LINKED ISSUE #920: Failed calico-node start exits with success status
I'm launching calico-node using calicoctl in mesos. My tasks fail because they're binding to an IP which is already in use by Calico (which is a mesos problem that I'll need to resolve seperately myself).

However, calicoctl is exiting with status code 0 when this happens, instead of an error. This is bug.

The implications of this is that my mesos tasks think they've completed successfully and reporting that they've `FINISHED` instead of `FAILED`:
```
sh -c './calicoctl node --detach=false'
Forked command at 4309
Running Docker container with the following command:

docker run --restart=always --net=host --privileged --name=calico-node -e HOSTNAME=djo-dcos-agent-3.gulfstream.datcon.co.uk -e IP= -e IP6= -e CALICO_NETWORKING=true -e AS= -e NO_DEFAULT_POOLS= -e ETCD_AUTHORITY=127.0.0.1:2379 -e ETCD_SCHEME=http -v /var/log/calico:/var/log/calico -v /var/run/calico:/var/run/calico calico/node:v0.18.0

Calico node is running with id: b0c21b9831319d2ea7f0a6c26cfd262760f8241a847c81053eb10f9654d4f013
Waiting for successful startup
No IP provided. Using detected IP: 198.51.100.3
ERROR: IP address 198.51.100.3 is already in use by host djo-dcos-agent-2.gulfstream.datcon.co.uk. Calico requires each compute host to have a unique IP. If this is your first time running the Calico node on this host, ensure that another host is not already using the same IP address.
Calico node failed to start
Command exited with status 0 (pid: 4309)
```

TrimBiggs:
Fixes #920 

When the node fails due to some unknown error to `calicoctl`, `calicoctl` still exits with status 0.

Easily reproducable scenario is starting the calico node, changing hostname, then restarting the calico node.  `calicoctl` will show that the container did not start, but status is still 0.

This change ensures we are setting a clean status if we are expecting to shut down the container, otherwise exit with status 1.

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -836,6 +836,7 @@ def _attach_and_stream(container, startup_only):
         sys.exit(0)
     signal.signal(signal.SIGTERM, handle_sigterm)
     stop_container_on_exit = True
+    exit_code = 1
 
     output = docker_client.attach(container, stream=True)
     try:
@@ -847,8 +848,12 @@ def _attach_and_stream(container, startup_only):
     except KeyboardInterrupt:
         # Mainline. Someone pressed Ctrl-C.
         print "Stopping Calico node..."
+        stop_container_on_exit = True
+        exit_code = 130
     finally:
         # Could either be this process is being killed, or output generator
         # raises an exception.
         if stop_container_on_exit:
             docker_client.stop(container)
+            # If the container is stopped, some sort of error occurred.
+            sys.exit(exit_code)
--- a/calicoctl/tests/unit/node_test.py
+++ b/calicoctl/tests/unit/node_test.py
@@ -94,6 +94,7 @@ class TestAttachAndStream(unittest.TestCase):
                                          call("from the container.")])
         self.assertEqual(m_stdout.write.call_count, 2)
         m_docker_client.stop.assert_called_once_with(m_container)
+        m_sys.exit.assertcalled_once_with(130)
 
     @patch("calico_ctl.node.docker_client", spec=DockerClient)
     @patch("calico_ctl.node.sys", spec=sys)
@@ -121,7 +122,6 @@ class TestAttachAndStream(unittest.TestCase):
         m_docker_client.attach.assert_called_once_with(m_container,
                                                        stream=True)
         self.assertFalse(m_container.called)
-        m_sys.exit.assert_called_once_with(0)
         m_stdout.write.assert_has_calls([call("Some output\n"),
                                          call("from the container."),
                                          call("\nThis output is printed, but "
@@ -134,6 +134,9 @@ class TestAttachAndStream(unittest.TestCase):
         m_docker_client.stop.assert_has_calls([call(m_container),
                                                call(m_container)])
         self.assertEqual(m_docker_client.stop.call_count, 2)
+        # sys.exit gets called twice: once when handling the SIGTERM and once
+        # when stopping the container for an unknown reason
+        m_sys.exit.assert_has_calls([call(0), call(1)])
 
 
 class TestNode(unittest.TestCase):


calicoctl/calico_ctl/node.py
```diff
@@ -847,8 +848,13 @@ def handle_sigterm(sig, frame):
...
+        clean_exit = True
     finally:
         # Could either be this process is being killed, or output generator
         # raises an exception.
         if stop_container_on_exit:
             docker_client.stop(container)
+            # If the container is stopped due to unknown circumstances, exit
+            # with failed status code.
+            if not clean_exit:
+                sys.exit(1)
```
djosborne: The docker-py library has a method called `wait` that is identical to `attach`, but instead of returning stdout/stderr on exit, it returns the return code of the container. I did a quick test and running `wait` on a stopped container will immediately return the ret code.

I think a nicer solution would be to just `sys.exit(docker_client.wait(container))`

djosborne: Actually I suppose `sys.exit(0)` here would break our flow, since we don't want to quit on success? But at the very least, I think we should bubble up the return code from calico-node instead of converting it to a 1 or 0

robbrockbank: Actually you could do what dan suggests:

```
try:
    rc = docker_client.wait(container)
    sys.exit(rc)
except KeyboardException:
    sys.exit(0)
finally:
    do finally stuff
```

This should work IIRC because sys.exit() just raises a SystemExit exception, so you're finally branch should still be hit.

Couple of things though:  (1) we need to make it super clear that this method will always do a sys.exit() - although you could arguably remove the sys.exit() from the Keyboardinterrupt branch and let it percolate up.

and (2) do we lose any useful logging if we use wait rather than attach?

TrimBiggs: I don't think that will work since `attach()` returns the output of the `calicoctl node` command whereas `wait()` returns the status code.

Turk and I chatted about this a bit earlier.  Gist was:

- We want the sys.exit to return the status of the calicoctl command, not the container's status (which `docker_client.wait()` returns).
- `docker_client.wait(container)` will always return a non-zero status on a container that has been stopped (such as with `docker_client.stop(container)`)
- KeyboardInterrupt should probably return an error status (SIGINT status code is 130) when the `calicoctl` command is stopped with ctrl-C (other unix commands have code 130 when SIGINT'd)

I think we should be running as normal if we reach the success state, but we should exit with `sys.exit(130)` if we had a keyboard interrupt and `sys.exit(1)` otherwise if we have not hit the success state.

TrimBiggs: Actually, I'm thinking `attach` vs `wait` a bit more and doing some minor tests to check if it would work

robbrockbank: I'm not sure what you mean by attach() returns the output of the calicoctl node command.  You mean it returns the output from container? 

My personal take is that I don't think we care what error code is returned from the container - so using wait doesn't feel particularly useful to me.  I think that is in agreement with what you mention above.

TrimBiggs: Yes, I meant the `wait()` returns the status of the container, not the command.

And yes, we're in agreement about not really caring about the container status output, but it could be simpler if we have wait instead of attach to detect any error, rather than scrubbing for the string output, but I guess the string detection is already working so why fix it..

TrimBiggs:
Updated

calicoctl/calico_ctl/node.py
```diff
@@ -847,8 +848,12 @@ def handle_sigterm(sig, frame):
...
         print "Stopping Calico node..."
+        stop_container_on_exit = True
+        exit_code = 130
     finally:
         # Could either be this process is being killed, or output generator
         # raises an exception.
         if stop_container_on_exit:
             docker_client.stop(container)
+            # If the container is stopped, some sort of error occurred.
+            sys.exit(exit_code)
```
robbrockbank: So why do we exit with 0 if the container did not successfully start?

TrimBiggs: We don't?

`exit_code` is set to either 1 or 130 in this function, it will never be 0 if the container is stopped here.

TrimBiggs: Where are you seeing exit with 0?

robbrockbank: Oh I'm being dumb - should have expanded the whole thing.  Yup yup looks ok.

robbrockbank: I can't help thinking the structure could be simplified though... let me jot down how:

robbrockbank:
I was being dumb.... LGTM.

PR #964, opened 2016-05-05
Query about 'virtualized environment'

neiljerram:
Proposing this as a starting point for discussion - obviously my text is not actually a suitable change.

### DIFF
--- a/docs/calico-with-docker/without-docker-networking/README.md
+++ b/docs/calico-with-docker/without-docker-networking/README.md
@@ -23,6 +23,9 @@ with a number of installation dependencies.
 
 Follow the instructions in one of the tutorials below to set up a virtualized
 environment using Vagrant or a cloud service - be sure to follow the
+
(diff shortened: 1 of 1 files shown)


docs/calico-with-docker/without-docker-networking/README.md
```diff
@@ -23,6 +23,9 @@ with a number of installation dependencies.
 
 Follow the instructions in one of the tutorials below to set up a virtualized
 environment using Vagrant or a cloud service - be sure to follow the
+
+[NJ] Why would I want a 'virtualized environment', and what does that mean?  My situation is that I already have a Linux host, and that I want to use Docker to create some containers on it, and use Calico to network those.  Is there some 'environment' that I need that I do not already have?
+
```
matthewdupre: No, you should use the manual setup instructions just below.

I think this is the correct ordering as well; most people who come to these instructions are going to use them on VMs.

neiljerram: I'm not sure you've got my point, which is that I genuinely don't understand what 'virtualized environment' is supposed to mean here.

tomdee: It means a "prepackaged" environment

PR #965, opened 2016-05-05
Queries on Manual Setup doc

neiljerram:
Proposing as a starting point for discussion - not an actual change as it currently stands.

### DIFF
--- a/docs/calico-with-docker/without-docker-networking/ManualSetup.md
+++ b/docs/calico-with-docker/without-docker-networking/ManualSetup.md
@@ -28,7 +28,13 @@ They must have the following software installed:
 
 - Docker v1.6 or greater: [Docker][docker]
 - etcd installed and available on each node: [etcd documentation][etcd]
+
+[NJ] What does 'available' mean?  Does this mean that I need a configured and active etcd system, such that I can do 'etcdctl ls /' on each node?  Any recommendations or restrictions on how many etcd masters, or where I can use proxies?
+
 - `ipset`, `iptables`, and `ip6tables` kernel modules.
+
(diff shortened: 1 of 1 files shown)


docs/calico-with-docker/without-docker-networking/ManualSetup.md
```diff
@@ -28,7 +28,13 @@ They must have the following software installed:
 
 - Docker v1.6 or greater: [Docker][docker]
 - etcd installed and available on each node: [etcd documentation][etcd]
+
+[NJ] What does 'available' mean?  Does this mean that I need a configured and active etcd system, such that I can do 'etcdctl ls /' on each node?  Any recommendations or restrictions on how many etcd masters, or where I can use proxies?
+
```
fasaxc: At this stage, we can support most (all?) etcd topologies, with or without proxies.  If you're running a local proxy then we work without config.  If not, you need to configure the correct URL(s) for us to connect to.

matthewdupre:
I think the first of these is fine as it is (see Shaun's comment), and the second we've changed elsewhere.  Closing.

neiljerram:
@matthewdupre @fasaxc I think I didn't express my first point well.  Imagine a new reader who is going through the process step by step.  They see an instruction saying that etcd must be installed - and that's fine, because everyone knows what 'installed' means.  But 'and available'?  Huh, why does it also say that; what does it mean?

I think the answer is nothing, i.e. this is just verbosity - because required etcd setup is in fact covered more properly, just a few lines further below.  I'll propose another PR to clean this up.

neiljerram:
New PR is at https://github.com/projectcalico/calico-containers/pull/1011

PR #972, opened 2016-05-06
Use new validate_port_spec to validate port spec

neiljerram:
(Instead of validate_ports.  validate_port_spec validates the port
specifications that can be specified on the calicoctl command line. This
is different from validate_ports, which validates the port values that
can occur in the etcd data model.)

### DIFF
--- a/calicoctl/calico_ctl/profile.py
+++ b/calicoctl/calico_ctl/profile.py
@@ -77,7 +77,7 @@ from pycalico.datastore import Rules
 from connectors import client, DOCKER_URL
 from utils import print_paragraph, DOCKER_LIBNETWORK_VERSION
 from pycalico.datastore_datatypes import Profile
-from pycalico.util import (validate_characters, validate_ports,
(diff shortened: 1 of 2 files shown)


matthewdupre:
I think I'd prefer validate_ports_str() to validate_port_spec() - it's not very clear what the latter actually means.

neiljerram:
Thanks for your review, @matthewdupre, I've updated the PR.

I believe the CI failures above are because we need the libcalico PR to merge first.  What is the procedure here?

matthewdupre:
OK, all fine.

I think just merge them both, and then deal with the builds if they aren't good after they've all gone in.

neiljerram:
@robbrockbank Please could you review this change?  I believe you wrote the validate_test_arguments tests for DSTPORTS and SRCPORTS, which this change corrects - so I'd appreciate if you could check that I'm not misunderstanding something.

robbrockbank:
LGTM - I'm surprised we hadn't spotted that sooner :-/

PR #976, opened 2016-05-10
Change dind from volume mount to official dind image

djosborne:
Update calico/containers to adhere to changes in [#1](https://github.com/projectcalico/dind/pull/1) whereby we no longer volume mount docker in from the host, but the new calico/dind which is built from the official dind image

### DIFF
--- a/calico_test/Dockerfile
+++ b/calico_test/Dockerfile
@@ -11,9 +11,23 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM jeanblanchard/alpine-glibc
+FROM docker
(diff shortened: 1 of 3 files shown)


calico_test/Dockerfile
```diff
@@ -11,9 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM jeanblanchard/alpine-glibc
+FROM docker
 MAINTAINER Tom Denham <tom@projectcalico.org>
-
+RUN apk add --update iptables ip6tables iproute2 curl && \
```
caseydavenport: Stick a comment in as to why we do this, rather than just use the alpine-libc image.

caseydavenport:
One comment, then looks ok.

PR #978, opened 2016-05-11
Remove line to get docker 1.10 in ubuntu vagrants, 1.11 now default

### LINKED ISSUE #892: Non-libnetwork docker vagrant-ubuntu: docker daemon not started by default
Used the docker non-libnetwork Ubuntu vagrant file for a bit of prototyping.  Found that, apart from on calico-01, docker wasn't started by default.  `sudo start docker` did the trick.

TrimBiggs:
Docker was failing to start when running the ubunutu vagrant demos because a newer version of docker is now the default. We were pulling docker 1.10, but the config for docker 1.11 uses a new flag that 1.10 doesn't know about. Just use the latest 1.11.

Fixes #892

### DIFF
--- a/docs/calico-with-docker/docker-network-plugin/vagrant-ubuntu/Vagrantfile
+++ b/docs/calico-with-docker/docker-network-plugin/vagrant-ubuntu/Vagrantfile
@@ -51,10 +51,6 @@ Vagrant.configure(2) do |config|
           "calico/node:#{calico_node_ver}"
       ]
 
-      # Replace docker with the docker 1.10.1 build
-      host.vm.provision :shell, inline: "stop docker", :privileged => true
-      host.vm.provision :shell, inline: "wget -qO /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-1.10.1", :privileged => true
-
       # Calico uses etcd for calico and docker clustering. Install it on the first host only.
       if i == 1
         # Download etcd and start.
@@ -66,11 +62,9 @@ Vagrant.configure(2) do |config|
         SHELL
       end
 
-      # Set Docker to use etcd for multihost.
+      # Set Docker to use etcd for multihost, then send SIGHUP to Docker process to restart.
       host.vm.provision :shell, inline: %Q|sudo sh -c 'echo "DOCKER_OPTS=\"--cluster-store=etcd://#{primary_ip}:2379\"" > /etc/default/docker'|
-
-      # Restart docker.
-      host.vm.provision :shell, inline: "sudo start docker"
+      host.vm.provision :shell, inline: "restart docker"
 
       # download calicoctl.
       host.vm.provision :shell, inline: "wget -qO /usr/local/bin/calicoctl #{calicoctl_url}", :privileged => true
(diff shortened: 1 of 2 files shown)


docs/calico-with-docker/docker-network-plugin/vagrant-ubuntu/Vagrantfile
```diff
@@ -69,9 +65,6 @@ Vagrant.configure(2) do |config|
       # Set Docker to use etcd for multihost.
       host.vm.provision :shell, inline: %Q|sudo sh -c 'echo "DOCKER_OPTS=\"--cluster-store=etcd://#{primary_ip}:2379\"" > /etc/default/docker'|
```
robbrockbank: We are currently setting the cluster store using the docker daemon command line options.  If we do that, don't we need to restart the docker daemon to pick this up?  I would have not expected this to work - does it?

Perhaps we should also change to use the daemon.json file and then just send in a SIGHUP

docs/calico-with-docker/without-docker-networking/vagrant-ubuntu/Vagrantfile
```diff
@@ -53,10 +53,6 @@ Vagrant.configure(2) do |config|
       host.vm.provision :shell, inline: "wget -qO /usr/local/bin/calicoctl #{calicoctl_url}", :privileged => true
       host.vm.provision :shell, inline: "chmod +x /usr/local/bin/calicoctl"
 
-      # Replace docker with the docker 1.10.1 build
-      host.vm.provision :shell, inline: "stop docker", :privileged => true
-      host.vm.provision :shell, inline: "wget -qO /usr/bin/docker https://get.docker.com/builds/Linux/x86_64/docker-1.10.1", :privileged => true
-
```
robbrockbank: I'm confused how this worked to be honest.  As far as I can tell we stop docker, but then never restart it again.

TrimBiggs: This was failing before (see #892 from the description above) because we were not starting docker after shutting it down.  This was the original reason for making this change.

docs/calico-with-docker/docker-network-plugin/vagrant-ubuntu/Vagrantfile
```diff
@@ -66,11 +62,9 @@ Vagrant.configure(2) do |config|
         SHELL
       end
 
-      # Set Docker to use etcd for multihost.
+      # Set Docker to use etcd for multihost, then send SIGHUP to Docker process to restart.
       host.vm.provision :shell, inline: %Q|sudo sh -c 'echo "DOCKER_OPTS=\"--cluster-store=etcd://#{primary_ip}:2379\"" > /etc/default/docker'|
-
-      # Restart docker.
-      host.vm.provision :shell, inline: "sudo start docker"
+      host.vm.provision :shell, inline: "restart docker", :privileged => true
```
djosborne: `:privileged` is  `True` by default

docs/calico-with-docker/without-docker-networking/vagrant-ubuntu/Vagrantfile
djosborne: ( @robbrockbank  The docker provision install script used to start docker back up automatically, I believe)

djosborne:
Ran through the docker libnetwork demo and this change worked fine. LGTM (besides the default string comment)

TrimBiggs:
Both vagrant demos are working fine. Merging.

PR #980, opened 2016-05-11
Prevent installation of docker 1.10 in vagrantfile

djosborne:
The vagrant docker provisioner installs different versions of docker depending on which version of vagrant you're using. If you're using the latest vagrant, this means docker 1.11. The 1.11 base installation includes flags unsupported in 1.10. So the lines I've removed in this PR would otherwise downgrade docker and break since the daemon is configured with 1.11 flags


TrimBiggs:
Duplicate of #978

PR #981, opened 2016-05-11
Move calico-test to libcalico

djosborne:
- calico/test uses a particular version of libcalico, which means it is tied to libcalico releases, not calico-containers releases
- calico/test is also used by libnetwork testing, so since it contains shared functionality, it should reside in libcalico

[Here's the PR to add this to libcalico](https://github.com/projectcalico/libcalico/pull/112)

A common scenario where the current layout is problematic: I've made a few changes to libcalico (adding/removing functions). I've also made changes to libnetwork to use this new functionality. If I run the unit tests for libnetwork they will fail. They use calico/test to run the unit tests. Calico/test is using libcalico v0.13.0. The only way I can test libnetwork with my libcalico changes is if I go into calico-containers, modify the requirements file of `calico/test` to point to a remote branch of libcalico, and rebuild calico/test.

If we make the change I propose, to test the changes, I would simply have to go into libcalico and run `make calico_test` as it would build a calico/test using the local libcalico changes, instead of installing libcalico from somewhere remote. After, running `make ut` on libnetwork would pass.


Makefile
```diff
@@ -23,7 +20,6 @@ default: help
 all: test                ## Run all the tests
 binary: dist/calicoctl   ## Create the calicoctl binary
 node_image: calico_node/.calico_node.created ## Create the calico/node image
-test_image: calico_test/.calico_test.created ## Create the calico/test image
```
matthewdupre: This is also referenced by the .PHONY line at the top of the file.

Makefile
```diff
@@ -90,13 +82,13 @@ birdcl:
...
 ## Run the UTs in a container.
-ut: calico_test/.calico_test.created
+ut:
 	docker run --rm -v `pwd`/calicoctl:/code calico/test \
 		nosetests $(UT_TO_RUN) -c nose.cfg
 	docker run --rm -v `pwd`/calico_node:/code calico/test \
 		nosetests filesystem/tests --with-coverage --cover-package=filesystem
 
-ut-circle: calico_test/.calico_test.created dist/calicoctl
+ut-circle: dist/calicoctl
```
matthewdupre: We should test the circle build before this goes in.

matthewdupre:
@djosborne couple of little tweaks, and a testing question for you.

Makefile
djosborne: Circle is configured to just run `make ut`. So the tests that this PR triggered show that it should not be a problem.

However, this merge will cause failed dockerhub builds of calico/test, as its configured to point to this repository. My next task is to reinitialize that repo, pointing it at libcalico instead

PR #994, opened 2016-05-20
Fix version link

chengweiv5:


matthewdupre:
Usually version updates like this should be happening automatically as part of our release process (see release-scripts/do_release.py).  As it happens, release-scripts/utils.py explicitly excludes this file from updating, probably because it discusses building from a clone anyway.

I think the right thing to do is probably rip the warning message out entirely, since it's a developer doc anyway and the 'release version' of it is no different.

matthewdupre:
@chengweiv5 if you could make the

PR #1014, opened 2016-06-07
Fix doc about removing node BGP peer

depay:
Signed-off-by: depay <depay19@163.com>


matthewdupre:
Looks good - thanks for contributing :).

PR #1019, opened 2016-06-14
Update URL of slack badge

eepyaich:


eepyaich:
Hold off on this while we try to get HTTPS working.

liljenstolpe:
Adding SSL right now is $20/month on heroku.  There is a beta on a free SSL.  I'd prefer to wait a bit.  Let's go forward with the http version, and I'll add ssl as soon as possible.

PR #1026, opened 2016-06-24
Watch etcd for new ipip pools

djosborne:
Currently, Felix watches etcd and creates a tunl0 interface if a pool with `--ipip` was added. It also watches the host-specific setting `IpInIpTunnelAddr`  and assigns that address to the tunnel. 

However, `IpInIpTunnelAddr` is only set during invocation of `calicoctl node`. This means that if you add an `--ipip` pool without rerunning node on every Host, your host won't have allocated itself an IP, and your tunl interface won't be assigned one.

This PR uses confd to watch for changes in the ip pools, and allocates an `IpInIpTunnelAddr`  and assigns the address if a pool with `--ipip` is added.

### DIFF
--- a/calico_node/filesystem/allocate-ipip-addr.py
+++ b/calico_node/filesystem/allocate-ipip-addr.py
@@ -1,0 +1,29 @@
+# If IPIP is enabled, the host requires an IP address for its tunnel
+# device, which is in an IPIP pool.  Without this, a host can't originate
+# traffic to a pool address because the response traffic would not be
+# routed via the tunnel (likely being dropped by RPF checks in the fabric).
+#
+# This is a oneshot python script that queries etcd for existing pools.
+# If any pool has --ipip, it will ensure this host's tunl0 interface
+# has been assigned an IP from the ipip pool.
+
+import os
+from startup import _ensure_host_tunnel_addr, _remove_host_tunnel_addr
+from pycalico.ipam import IPAMClient
+
+def main():
+	ipv4_pools = client.get_ip_pools(4)
+	ipip_pools = [p for p in ipv4_pools if p.ipip]
+
+	if ipip_pools:
+	    # IPIP is enabled, make sure the host has an address for its tunnel.
+	    _ensure_host_tunnel_addr(ipv4_pools, ipip_pools)
+	else:
+	    # No IPIP pools, clean up any old address.
+	    _remove_host_tunnel_addr()
+
+hostname = os.getenv("HOSTNAME")
+client = IPAMClient()
+
+if __name__ == "__main__":
+    main()
--- a/calico_node/filesystem/etc/calico/confd/templates/tunl-ip.template
+++ b/calico_node/filesystem/etc/calico/confd/templates/tunl-ip.template
@@ -1,0 +1,7 @@
+We must dump all pool data to this file to trigger a resync.
+Otherwise, confd notices the file hasn't changed and won't
+run our python update script.
+ 
+{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}
+  {{if $data.ipip}}{{$data.cidr}}{{end}}
+{{end}}
--- a/tests/st/test_ipip.py
+++ b/tests/st/test_ipip.py
@@ -12,8 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from unittest import skip
+import re
 
+from netaddr import IPAddress, IPNetwork
 from test_base import TestBase
+from tests.st.utils.docker_host import DockerHost
+from time import sleep
+from tests.st.utils.utils import retry_until_success
 
 """
 Test calico IPIP behaviour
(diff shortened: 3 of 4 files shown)


calico_node/filesystem/etc/calico/confd/templates/tunl-ip.template
```diff
@@ -0,0 +1,7 @@
+We must dump all pool data to this file to trigger a resync.
+Otherwise, confd notices the file hasn't changed and won't
+run our python update script.
+ 
+{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}
```
robbrockbank: Worth only writing out the IPIP pools - that way we only do an update when an IPIP pool is added or deleted.

calico_node/filesystem/allocate-ipip-addr.py
```diff
@@ -0,0 +1,29 @@
...
+from startup import _ensure_host_tunnel_addr, _remove_host_tunnel_addr
+from pycalico.ipam import IPAMClient
+
+def main():
+	ipv4_pools = client.get_ip_pools(4)
+	ipip_pools = [p for p in ipv4_pools if p.ipip]
+
+	if ipip_pools:
+	    # IPIP is enabled, make sure the host has an address for its tunnel.
+	    _ensure_host_tunnel_addr(ipv4_pools, ipip_pools)
```
robbrockbank: Do these need to be called by the startup script anymore?

robbrockbank:
General question - having this processing in confd - is there anything we need to worry about for manual installation (e.g. if we install the node components directly on a host rather than in the node container)?

calico_node/filesystem/allocate-ipip-addr.py
djosborne: Yes, the watch I've implemented only activates on a pool change. But the following scenario requires the allocation made by node:

The user adds a pool with `--ipip`, then launches `calicoctl node` on a Host which has not run calico node up till this point. In this scenario, the etcd watch I've added will not trigger, since the pools were not changed.

djosborne:
WRT your general question: Our experimental guide for running uncontainerized calico has the user running confd, and wildcard copying the template files. So if you were to follow the current guide, everything should work, right up to the execution of the newly introduced python script.

However, these changes assume that that new python script exists at `/allocate-ipip-addr.py` on the host, and that the python environment has the right pip libraries installed. We already make the user install the `restart-calico-confd` script. But this PR means they'd also have to install the aformentioned python script, and ensure it has the right libraries.

These are not very easy-to-satisfy requirements, especially on a CoreOs readonly system, for example.

I'll move the python script to `/usr/local/bin`, add a `#!/usr/env/python` at the top, and get rid of the `.py` extension to make it a bit nicer. 

But as for the issue of it being a python script, we could introduce a build system to pyinstall build it as a binary. But that would mean much more overhead and keeping track of it as a seperate entity. 

I think a better option would be to disclose to uncontainerized users that without that script in place, they will not get the automatic --ipip pool update, and will need to rerun calicoctl node on each host to get a tunl ip. The good news is, without the script in place, confd will gracefully fail only the reload command, without affecting the other functionality.

djosborne:
Actually, I just realized that we never documented the move to `startup.py` in our containerless calico guide. So at this point, we basically do not support running dockerless calico anyways. 

I'm leaning more towards merging this PR and looking at an overhaul down the rode, post go implementation...

caseydavenport:
@djosborne - `/usr` is read-only on CoreOS, so moving to that directory won't help you.  I believe `/` is fine?

djosborne:
Didn't realize that, but good to know. Additionally, this script imports functions from `startup.py`, so it needs to sit alongside it in `/`

calico_node/filesystem/allocate-ipip-addr.py
robbrockbank: agreed.  fine.

robbrockbank:
Yup it sounds like our container less calico needs an overhaul.

LGTM for this - but can you raise a ticket to sort out container less calico and maybe put some notes about what was missing (maybe just copy the text from above?)

djosborne:
Raised #1041

tests/st/test_ipip.py
```diff
@@ -29,3 +32,56 @@ class TestIPIP(TestBase):
...
+            self.assert_tunl_ip(host, r'inet 10\.0\.1\.[\d]{1,3}/32', expect=True)
+
+            # Test that by adding another pool, then deleting the first,
+            # we remove the original IP, and allocate a new one from the new pool
+            new_ipv4_pool = "192.168.0.0/16"
+            host.calicoctl("pool add %s --ipip" % new_ipv4_pool)
+            host.calicoctl("pool remove %s" % ipv4_pool)
+            self.assert_tunl_ip(host, r'inet 192\.168\.[\d]{1,3}.[\d]{1,3}/32')
+
+    def assert_tunl_ip(self, host, re_expr, expect=True):
```
robbrockbank: This is a rather odd interface IMHO.  You are asserting the tunl IP but you are passing in the net regex.  This function should really be doing the match and pulling out the IP address - and then I'd imagine you'd want to check that the IP is from one a number of available pools or CIDRs.

tests/st/test_ipip.py
```diff
@@ -29,3 +32,56 @@ class TestIPIP(TestBase):
     @skip("Not written yet")
     def test_ipip(self):
         pass
+
+    def test_ipip_addr_assigned(self):
+        with DockerHost('host', dind=False, start_calico=False) as host:
```
robbrockbank: Probably an overkill - but should we test it all is good when starting the node without any pool being added (and setting the node environment so that a pool is not added by the node command?)

tests/st/test_ipip.py
```diff
@@ -29,3 +32,56 @@ class TestIPIP(TestBase):
...
+            # Set up first pool before Node is started, to ensure we get tunl IP on boot
+            ipv4_pool = "10.0.1.0/24"
+            host.calicoctl("pool add %s --ipip" % ipv4_pool)
```
robbrockbank: We should probably add a non-IPIP pool too somewhere and/or multiple pools.

matthewdupre:
FYI / is read-only on GCI, which Google seem to be pushing for Kubernetes (used on GKE, and by default kube-up).

I doubt anyone's installing uncontainerised on there currently, but it could be a problem in future.

robbrockbank:
@djosborne what's the status of this one?

djosborne:
@robbrockbank if you agree that the written STs are sufficient, this is ready to merge

robbrockbank:
@djosborne :  Perhaps they just expired, but it doesn't look like the tests ran for this.  If this needs rebasing can you do that and re-push.  That way it'll run the tests.

djosborne:
Closing in favor of #1097 which includes this commit. and adds teardown which removes the ip from tunl, fixing the STs.

PR #1029, opened 2016-06-27
Bumped ICMP validation test to use 256

djosborne:
Since we've increased the range of valid ICMP types to include 255 in [libcalico#117](https://github.com/projectcalico/libcalico/pull/117).


caseydavenport:
LGTM.

PR #1030, opened 2016-06-28
Remove version header from building.md

tomdee:
It's about building from master so doesn't need a version header


djosborne:
LGTM

PR #1038, opened 2016-07-05
Replace Metaswitch copyright statements with Tigera.

fasaxc:
According to the lawyers, we need to update the existing copyrights too.


fasaxc:
@tomdee Not sure how I can have caused the circle CI failure.  Is that a known intermittent one?

djosborne:
@fasaxc I believe that's just an issue caused by a change upstream in libcalico. We've made a downstream fix in this repo to resolve it. A quick rebase off later master should pass. I know the docs are unrelated, so I don't mind merging, but it should also be pretty quick to get this passing

fasaxc:
Rebased...

PR #1039, opened 2016-07-05
Pinned alpine to fix setuptools issue.

djosborne:
This PR temporarily fixes the build by pinning the alpine image calico-node is built from to 3.3. 

See #1040 for more information on what may be causing failures.

### DIFF
--- a/calico_node/build.sh
+++ b/calico_node/build.sh
@@ -16,13 +16,14 @@
 set -e
 set -x
 
-# Ensure the main and testing repros are present. Needed for runit
-echo "http://alpine.gliderlabs.com/alpine/edge/testing" >> /etc/apk/repositories
-
 # These packages make it into the final image.
-apk -U add runit python py-setuptools libffi ip6tables ipset iputils iproute2 yajl conntrack-tools
+# Install runit from the testing repository, as its not yet available in global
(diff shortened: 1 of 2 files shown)


calico_node/build.sh
```diff
@@ -16,13 +16,14 @@
 set -e
 set -x
 
-# Ensure the main and testing repros are present. Needed for runit
-echo "http://alpine.gliderlabs.com/alpine/edge/testing" >> /etc/apk/repositories
-
 # These packages make it into the final image.
-apk -U add runit python py-setuptools libffi ip6tables ipset iputils iproute2 yajl conntrack-tools
+# Install runit from the testing repository, as its not yet available in global
+apk add runit --repository "http://alpine.gliderlabs.com/alpine/edge/testing" --update-cache
```
tomdee: I prefer having the package name after all the arguments

calico_node/build.sh
```diff
@@ -16,13 +16,14 @@
...
+# Install remaining deps from the global repository
+apk add python py-setuptools libffi ip6tables ipset iputils iproute2 yajl conntrack-tools --update-cache
```
tomdee: As above, I think it's better to have update-cache first.

tomdee:
One small nit, otherwise LGTM

PR #1047, opened 2016-07-07
Update k8s GCE guide to 1.3.0

caseydavenport:
This PR does the following:

- Removes the documentation on annotation based + v1alpha1 network policy
- Adds v1beta1 policy to the integration guide
- Updates the GCE guide to use v1.3.0 k8s and v1beta1 NetworkPolicy.


tomdee:
In the integration guide - 
` - An existing Kubernetes cluster running Kubernetes >= v1.1`
But that's not going to work for network policy

tomdee:
You missed a reference to v1.1.4 in the intergration guide

tomdee:
I think its worth saying that iptables mode in kube-proxy is the default in both the integration guide and the readme

tomdee:
As we discussed, I think it might be worth leaving some reference to annotation based policy, but that's it not recommended for new deployments.

tomdee:
And finally, jazzing up the main readme would be nice - adding a link to a blog post or embedding a video would be nice.

PR #1049, opened 2016-07-07
Update AWS guide to install k8s v1.3.0, v1beta1 policy.

caseydavenport:
These are simply config changes:

- Bump the kubernetes version.
- Add the policy controller.


caseydavenport:
I've run through this and verified this works, and I've run the stars demo against it to make sure policy works.

PR #1053, opened 2016-07-09
Fix reference to policy-controller.yaml

### LINKED ISSUE #1052: Outdated pointer to policy-controller.yaml
https://github.com/projectcalico/calico-containers/blob/master/docs/cni/kubernetes/KubernetesIntegration.md refers to https://raw.githubusercontent.com/projectcalico/k8s-policy/v0.2.0/examples/policy-controller.yaml, which says
```
      image: calico/k8s-policy-controller
```
--- which is wrong, the correct image is referenced from https://raw.githubusercontent.com/projectcalico/k8s-policy/master/examples/policy-controller.yaml, which says the image is `calico/kube-policy-controller:latest`.

MikeSpreitzer:
Updated docs/cni/kubernetes/KubernetesIntegration.md to reference the
policy-controller.yaml file from the master branch rather than the
v0.2.0 tag because the latter has an outdated name for the policy
controller image.

Fixes #1052

### DIFF
--- a/docs/cni/kubernetes/KubernetesIntegration.md
+++ b/docs/cni/kubernetes/KubernetesIntegration.md
@@ -126,7 +126,7 @@ To install the policy controller:
 kubectl create ns calico-system
 ```
 
-- Place [this manifest](https://raw.githubusercontent.com/projectcalico/k8s-policy/v0.2.0/examples/policy-controller.yaml) in the kubelet's config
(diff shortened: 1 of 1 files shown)


caseydavenport:
I'll make a note to update this link when a new `k8s-policy` is released, because ideally links like this are versioned, but for now this works.  Thanks!

PR #1059, opened 2016-07-11
Update stars demo files and kubedns

caseydavenport:
The DNS manifest is used across the AWS, GCE, and Vagrant demos.  I've also checked that the stars demo works on each.

- Updates to latest kubedns
- Updates stars demo manifests to use correct domain names.
- Updates to use `calico/` versions of the stars Docker images rather than `caseydavenport/`


caseydavenport:
OK I've tested this works.  @tomdee could you give a look?

tomdee:
Looks very nice

PR #1065, opened 2016-07-14
Identified step 4 in the stars demo

MikeSpreitzer:
The identification of step 4 was missing.  The step was there, but it
was not called out as a distinct step.


caseydavenport:
LGTM - Thanks!

PR #1077, opened 2016-07-25
Make it clearer that libnetwork users must set --cluster-store

matthewdupre:
Been seeing a lot of people miss this recently, so I've tried to make it a bit clearer that it probably does apply to the reader.

### DIFF
--- a/docs/calico-with-docker/docker-network-plugin/ManualSetup.md
+++ b/docs/calico-with-docker/docker-network-plugin/ManualSetup.md
@@ -37,19 +37,19 @@ up a cluster.
 > NOTE: If you are running etcd with SSL/TLS, see the (Etcd Secure Cluster)[../../EtcdSecureCluster.md]
 > documentation.
 
-### Docker
+### Docker (with multi-host networking)
 
 Follow the instructions for installing [Docker][docker].  A version of 1.9 or
 greater is required.
 
-To use the multi-host native networking feature of Docker, the Docker daemon
-needs to be run specifying a cluster store.  If using etcd as a cluster store,
-run Docker daemon with the following additional parameter:
+To use Calico as a Docker network plugin, the Docker daemon
+needs to run specifying a cluster store.  If using etcd as a cluster store,
+configure the `cluster-store` on the Docker daemon to `etcd://<ETCD_IP>:<ETCD_PORT>`,
+replacing `<ETCD IP>` and <ETCD_PORT> with the appropriate address and client
+port for your etcd cluster.
 
-    --cluster-store=etcd://<ETCD IP>:2379
-
-Replacing `<ETCD IP>` with the appropriate address for your etcd cluster.  This
-also assumes your cluster uses the standard etcd port 2379.
+> For Docker 1.10+, you can use the [daemon configuration file][daemon-config-file],
+> or for 1.9 see the appropriate 'Configuring Docker' section in [configuring docker][configuring-docker-1.9].
 
 #### Docker permissions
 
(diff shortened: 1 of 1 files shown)


djosborne:
Cool! I was thinking the same thing. 

I also think it might be a good idea to change the title from just "Docker" to "Docker configured with a Cluster Store". My theory is that people just see "docker" and assume we're just saying to have it installed, without bothering to read the fine print. Thoughts?

docs/calico-with-docker/docker-network-plugin/ManualSetup.md
```diff
@@ -42,8 +42,8 @@ up a cluster.
 Follow the instructions for installing [Docker][docker].  A version of 1.9 or
 greater is required.
 
-To use the multi-host native networking feature of Docker, the Docker daemon
-needs to be run specifying a cluster store.  If using etcd as a cluster store,
+To use Calico as a libnetwork plugin, the Docker daemon
```
robbrockbank: Our docs describe this as a "Docker network plugin" - so we should probably stick with that.  I'm loathe to talk about libnetwork plugin because to me that seems like an implementation detail for the network plugin provider.

matthewdupre:
Updated

djosborne:
Hate to nit, but I strongly feel that "**configured** with multi-host networking" would really drive the point across.

docs/calico-with-docker/docker-network-plugin/ManualSetup.md
```diff
@@ -37,13 +37,13 @@ up a cluster.
...
 
-### Docker
+### Docker (with multi-host networking)
 
 Follow the instructions for installing [Docker][docker].  A version of 1.9 or
 greater is required.
 
-To use the multi-host native networking feature of Docker, the Docker daemon
-needs to be run specifying a cluster store.  If using etcd as a cluster store,
+To use Calico as a Docker network plugin, the Docker daemon
```
robbrockbank: I think we should also update this to mentioned daemon.json.   How about something like:

To use Calico as a Docker network plugin, the Docker daemon must be configured to run using a cluster store.  For simplicity you can use the same etcd cluster that is used by Calico, although it is not necessary to do so.  

There are two ways to configure Docker to use a cluster store:

1) Add the following additional parameter to your Docker daemon run command.

  --cluster-store=etcd://<ETCD_IP>:2379

2) Add the cluster-store details to your daemon.json file.  By default, Docker checks for the file `/etc/docker/daemon.json`.  If this file exists, add an entry for "cluster-store".  If the file does not exist, create a new one with the following contents.

    {
      "cluster-store": "etcd://<ETCD_IP>:<PORT>"
    }

matthewdupre: I'm not convinced adding the alternative suggestion adds enough value to outweigh the increase in doc length / complexity.  Anyone who knows about daemon.json will be able to use it easily enough.

djosborne: Normally I would agree with @matthewdupre  - I'm a fan of cutting down the redundant information from our docs. But this _is_ the docker-libnetwork guide. Its the only place I can understand going a little verbose on the docker-daemon configuration instructions.

robbrockbank: Yes I think we do need to talk about using daemon.json somewhere - quite often people aren't aware of it because in general they don't need to provide additional docker parms.  This seems like the right place for this information.  If you wanted to keep it simple then it would be better to remove the docker run parm - since I think daemon.json is the preferred way to go now... only problem there is that we'd probably want to update anywhere we mention the --cluster-store parameter.

matthewdupre:
So I looked it up, and daemon.json is only supported in 1.10+; whereas we support 1.9+.

I don't think it's worth it: DOCKER_OPTS is a perfectly good way of doing this; and it's not our place to teach people how to use docker (I don't like that we have the permissions section below, for example).

djosborne:
How about adding a link to wherever you found that^ information?

matthewdupre:
Updated.

robbrockbank:
LGTM

PR #1078, opened 2016-07-26
calicoctl diags: gather ipset list from inside container if it fails outside

### LINKED ISSUE #840: calicoctl diags fails to grab ipsets when ipset isn't installed
Tom: "We could try to run ipset in the calico-node container (or even run a calico/diags container). Could you raise a calico-containers bug to track that?"

However we go about it, it would be nice to see the ipsets.

matthewdupre:
Fixes #840.


tomdee:
LGTM - merging

Issue #1044, opened 2016-07-07
vagrant-ubuntu failing with '--advertise-client-urls' not found or does not exist..


fasaxc:
Looks like `calico-containers/docs/calico-with-docker/without-docker-networking/vagrant-ubuntu` is out-of-date.  Presumably, the :latest tag on the etcd image has been updated to a too-new release:

```
==> calico-01: -- Container: quay.io/coreos/etcd
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!


          rm -f /var/lib/vagrant/cids/79e5abb9b63993e510d9c9b46401c5b39425c119
          docker run --cidfile=/var/lib/vagrant/cids/79e5abb9b63993e510d9c9b46401c5b39425c119 -d --name quay.io-coreos-etcd --restart=always --net=host quay.io/coreos/etcd --advertise-client-urls http://172.17.8.101:2379 --listen-client-urls http://0.0.0.0:2379


Stdout from the command:

ec9f6282f845f52921e3b1eec02ab01fc1dff3a2f74a5a4dfa5ba42297374e40


Stderr from the command:

Unable to find image 'quay.io/coreos/etcd:latest' locally
latest: Pulling from coreos/etcd
e110a4a17941: Pulling fs layer
b9b0b8b25fab: Pulling fs layer
7cac2d6de03d: Pulling fs layer
8ea649a70577: Pulling fs layer
8ea649a70577: Waiting
e110a4a17941: Verifying Checksum
e110a4a17941: Download complete
e110a4a17941: Pull complete
7cac2d6de03d: Verifying Checksum
7cac2d6de03d: Download complete
8ea649a70577: Verifying Checksum
8ea649a70577: Download complete
b9b0b8b25fab: Verifying Checksum
b9b0b8b25fab: Download complete
b9b0b8b25fab: Pull complete
7cac2d6de03d: Pull complete
8ea649a70577: Pull complete
Digest: sha256:6dc8369bb1021b53df36f7cd4388170d4a36e0e3a4d34cbfa347dd0e31ccdf14
Status: Downloaded newer image for quay.io/coreos/etcd:latest
docker: Error response from daemon: Container command '--advertise-client-urls' not found or does not exist..

alagalah@thing2:vagrant-ubuntu (master)$
```

matthewdupre:
Yeah, this worked with etcd 2.3.7 but doesn't with 3.0.0.

matthewdupre:
[...]

PR #1079, opened 2016-07-27
Fix Ubuntu Vagrant demo

matthewdupre:
Fixes https://github.com/projectcalico/calico-containers/issues/1044

Sticking with v3.0.3 because it's shiny.


tomdee:
I can't see anywhere else that this needs changing so LGTM

Issue #1073, opened 2016-07-19
docker image tag and version are not in sync


leodotcloud:
I pulled the `v0.20.0` tag and `./calicoctl version` shows `0.19.0-dev`
```
core@calico-k8s-node-01 ~ $ docker run -v /var/run/calico:/var/run/calico -v /var/run/docker.sock:/var/run/docker.sock --net=host --privileged=true --rm -e ETCD_AUTHORITY=172.18.18.101:2379 --entrypoint=sh -it calico/ctl:v0.20.0
Unable to find image 'calico/ctl:v0.20.0' locally
v0.20.0: Pulling from calico/ctl
4d06f2521e4f: Pull complete
a3ed95caeb02: Pull complete
72de78712521: Pull complete
1b61d7ea5bcb: Pull complete
7c44d05d8b9d: Pull complete
Digest: sha256:6ce66a2483d89865801f2f2530a61f772b8de721e7c0a1a4ff90ae465675c7ab
Status: Downloaded newer image for calico/ctl:v0.20.0
/ # ./calicoctl
/ # calicoctl status
sh: calicoctl: not found
/ # ./calicoctl status
calico-node container is running. Status: Up 25 hours
Running felix version 1.4.0rc1

IPv4 BGP status
IP: 172.18.18.102    AS Number: 64511 (inherited)
+---------------+-------------------+-------+----------+-------------+
|  Peer address |     Peer type     | State |  Since   |     Info    |
+---------------+-------------------+-------+----------+-------------+
| 172.18.18.103 | node-to-node mesh |   up  | 12:35:43 | Established |
| 172.18.18.101 | node-to-node mesh |   up  | 18:31:12 | Established |
+---------------+-------------------+-------+----------+-------------+

IPv6 BGP status
No IPv6 address configured.

/ # ./calicoctl version
0.19.0-dev
/ #
```

matthewdupre:
Oops!  Thanks for reporting.

I'll add a note to the release process to check the version.  I'm not sure it's worth going back now though: we'll make sure to get it right for 0.21.

PR #1080, opened 2016-07-27
Ensure versions are correct when doing releases

matthewdupre:
Avoid repeats of #1073.  The release scripts looked right to me, there's manual steps involved, and 0.19 was right (and used the same scripts).

I think it's a one-off, so just adding a note to make sure :).


matthewdupre:
(I still think we might as well merge this, but there's also a concrete problem with the process of getting the calico/ctl docker build; which we'll fix when we move to caligoctl.)

tomdee:
LGTM

Issue #1084, opened 2016-08-02
Calico/node: auto-detect hostname if not provided


caseydavenport:
It would be useful if one didn't need to pass an explicit hostname to `calico/node`, and it instead defaulted to the value of `socket.gethostname()`.

Currently, there is no default: https://github.com/projectcalico/calico-containers/blob/master/calico_node/filesystem/startup.py#L306

My use case is running in a DaemonSet under Kubernetes, where the hostname is not readily available through the [Downward API](http://kubernetes.io/docs/user-guide/downward-api/).

caseydavenport:
Fixed by #1085

PR #1085, opened 2016-08-02
calico/node: default to socket.gethostname()

caseydavenport:
Fixes issue #1084 

I still need to test that this works for the DaemonSet use-case in the issue.

### DIFF
--- a/calico_node/filesystem/startup.py
+++ b/calico_node/filesystem/startup.py
@@ -14,6 +14,7 @@
 
 import os
 import sys
+import socket
 
 import netaddr
 from netaddr import AddrFormatError, IPAddress
@@ -259,9 +260,10 @@ def main():
     # Write a startup environment file containing the IP address that may have
     # just been detected.
     # This is required because the confd templates expect to be able to fill in
-    #  the IP by fetching it from the environment.
+    # some templates by fetching them from the environment.
     with open('startup.env', 'w') as f:
-        f.write("IP=" + ip)
+        f.write("IP=%s\n" % ip)
+        f.write("HOSTNAME=%s\n" % hostname)
 
     warn_if_hostname_conflict(ip)
 
@@ -303,7 +305,12 @@ def main():
         _remove_host_tunnel_addr()
 
 
+# Try the HOSTNAME environment variable, but default to
+# the socket.gethostname() value if unset.
 hostname = os.getenv("HOSTNAME")
+if not hostname:
+    hostname = socket.gethostname()
+
 client = IPAMClient()
 
 if __name__ == "__main__":


matthewdupre:
LGTM - feel free to merge once you've tested, and backed out the caseydavenport/ part.

calico_node/filesystem/startup.py
```diff
@@ -259,9 +260,10 @@ def main():
     # Write a startup environment file containing the IP address that may have
     # just been detected.
     # This is required because the confd templates expect to be able to fill in
-    #  the IP by fetching it from the environment.
+    # some templates by fetching them from the environment.
     with open('startup.env', 'w') as f:
-        f.write("IP=" + ip)
+        f.write("export IP=%s\n" % ip)
```
caseydavenport: Various scripts source this fie.  However, since the values are not exported they don't propagate to child processes, meaning things like `confd` which rely on these values being set in the environment don't see them.

@matthewdupre WDYT?

matthewdupre: I just ran a `calicoctl node --ip=<not default selection>` with a recent calico-node and confd picked up the IP variable fine.

What makes you say "this just didn't work before"?

matthewdupre: Looks like this file is sourced by start_runit; which afaik does propagate (and certainly looks like it does)?

caseydavenport: Yeah, I'm looking into it.  I'm not sure what's going on.

`calicoctl node --ip=<something>` will set the `IP` env var when starting the container (via a `-e` flag)

The case I'm talking about is when no `-e IP` flag is passed to Docker (i.e `calico/node` must auto-detect).

It looks like when I start the container with `-e IP` ( butno address specified), the auto-detection works just fine.  When I start it with no `-e` flag for the `IP`var, the auto-detection works but the value is not propagated to confd, which fails when trying to fill in the bird template here https://github.com/projectcalico/calico-containers/blob/master/calico_node/filesystem/etc/calico/confd/templates/bird.cfg.mesh.template#L5

matthewdupre: All sounds a bit weird...  Do you expect we'll rely on autodetection for the daemonsetted calico-node then?  We could specify it as an env option in kubernetes using a valueFrom:fieldRef:fieldPath:status.podIP instead (is there a hostname equivalent?)?

In a way I don't like calico-node doing any detection itself: I'd prefer it was pushed back to parameters (e.g. -e) rather than relying on side effects from the container inspecting the host networking.

caseydavenport: Having `calico/node` do the auto-detection is very useful for cases just like this.  I think we will rely on autodetection for the DaemonSet deployment.

I looked at `valueFrom:fieldRef:fieldPath:status.podIP`, but the value there seems less reliable than our auto-detection and we don't have control over it.  Could be that we patch k8s to get a better value here, or get the `nodeIP` to be available in the downward API (but that might never happen because the k8s devs don't like the idea too much).

matthewdupre: Is podIP ever different to nodeIP for a host networked pod?  If so when?  We may well end up depending on this for etcd anyway; so I'd like to understand any concrete problems with it.

calico-node detection may be useful, but it's also horrible for debugability: anything where we're having to guess what calico-node is doing inside is unpleasant, as you've found above.  Having parameters be auto-filled deep in the bowels of a container is just dubious design: they should be pinned down early - calico-node should depend as little as possible on the host it's running on.

caseydavenport:
OK - I've tested and backed out the caseydavenport bit as well as the other bits that weren't necessary for the immediate issue (hostname detection).

PR #1089, opened 2016-08-03
CA can be specified separately

### LINKED ISSUE #1046: calicoctl node requires all of ETCD_CA_CERT_FILE, ETCD_KEY_FILE, ETCD_CERT_FILE
but if etcd doesn't require authentication, only ETCD_CA_CERT_FILE should be needed.

see https://github.com/projectcalico/calico-containers/blob/master/calicoctl/calico_ctl/node.py#L285

robbrockbank:
Fixes #1046 

Both Felix and libcalico accept CA being specified without the Key and Cert files.


matthewdupre:
I assume you've tested this?

robbrockbank:
I tested in as much as I checked by hand what was passed across - but not actually with a live running etcd....   We should really add an ST for this scenario.   I have added a couple of UTs now though.

matthewdupre:
Wow that test is long!

PR #1090, opened 2016-08-03
Always create runtime dir

### LINKED ISSUE #951: calico-node is doing something which is deprecated?
I'm seeing this in my logs:

`dockerd[1141]: time="2016-04-28T09:39:42.608567851Z" level=warning msg="Auto-creating non-
existent volume host path /var/run/calico, this is deprecated and will be removed soon"`

I think this means that calico-node is doing something which it will soon be unable to do?

robbrockbank:
Fixes #951

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -258,9 +258,16 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
         if not etcd_ok or (using_docker and not docker_ok):
             sys.exit(1)
 
-    # Ensure log directory exists
-    if not os.path.exists(log_dir):
+    # Ensure log directory and /var/run/calico exist so that they can be
+    # mounted into the containers.
+    try:
         os.makedirs(log_dir)
+    except OSError:
+        pass
+    try:
+        os.makedirs("/var/run/calico")
+    except OSError:
+        pass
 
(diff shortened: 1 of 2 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -258,9 +258,12 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
         if not etcd_ok or (using_docker and not docker_ok):
             sys.exit(1)
 
-    # Ensure log directory exists
+    # Ensure log directory and /var/run/calico exist so that they can be
+    # mounted into the containers.
     if not os.path.exists(log_dir):
         os.makedirs(log_dir)
+    if not os.path.exists("/var/run/calico"):
+        os.makedirs("/var/run/calico")
```
matthewdupre: It's better (no race condition) and more Pythonic (forgiveness vs permission) to catch an exception here (although the verbosity does make me sad).

Sadly the exist_ok parameter isn't available until 3.2.

matthewdupre:
I'm not too fussed about changing to an exception - it's more intended as feedback for next time: the race here isn't likely.  If you can't be bothered, just hit merge instead :).

calicoctl/calico_ctl/node.py
robbrockbank: Yes good point!

PR #1094, opened 2016-08-04
Fix calico-mesos-docker demo

djosborne:
Fixed the docker systemd dropin as v1.12.0 of docker doesn't need fd:// in centos (apparently).

A better fix would be to change this so instead of using a systemd dropin, we set `daemon.json`...

### DIFF
--- a/docs/mesos/vagrant-centos/units/docker.service.d/docker.conf
+++ b/docs/mesos/vagrant-centos/units/docker.service.d/docker.conf
@@ -1,4 +1,4 @@
 [Service]
 EnvironmentFile=/etc/sysconfig/calico
 ExecStart=
-ExecStart=/usr/bin/docker daemon -H fd:// --cluster-store=etcd://${ETCD_AUTHORITY}
+ExecStart=/usr/bin/dockerd --cluster-store=etcd://${ETCD_AUTHORITY}


docs/mesos/vagrant-centos/units/docker.service.d/docker.conf
```diff
@@ -1,4 +1,4 @@
 [Service]
 EnvironmentFile=/etc/sysconfig/calico
 ExecStart=
-ExecStart=/usr/bin/docker daemon -H fd:// --cluster-store=etcd://${ETCD_AUTHORITY}
\ No newline at end of file
+ExecStart=/usr/bin/dockerd --cluster-store=etcd://${ETCD_AUTHORITY}
```
matthewdupre: I don't understand why there are two ExecStarts here - seems a bit weird.

djosborne: So I recall that the blank execstart is actually required to "clear"  the one you are overriding. 

I did some googling to back that up, but besides a few folks doing the same, I couldn't find any definitive proof that it's necessary. 

I'd really like to restore functionality of this demo. @matthewdupre are you comfortable merging this?

robbrockbank: Yeah I couldn't find details on what that did, but i remember seeing it in some Docker or CoreOS documentation somewhere, and I remember that not including it stops it from working -- so it was needed.

matthewdupre:
Sure, it's at worst superfluous.

PR #1101, opened 2016-08-09
Show etcd connection failure error message

alexbrand:
In the case of an etcd connection failure, the error message that might help the user in diagnosing the problem is being swallowed.


robbrockbank:
Thanks Alex - looks like a good fix!  Merging.

alexbrand:
Cheers!

PR #1104, opened 2016-08-10
Remove profile ID from rules

robbrockbank:


tomdee:
Looks fine to me. Do we know how/why this was there in the first place?

robbrockbank:
@tomdee - no idea I'm afraid - it makes the rules json a little nicer in that it also includes details about which profile it belongs to - but it's a minor point.  That's all that I can think of.

robbrockbank:
fixed in the latest release

PR #1106, opened 2016-08-12
Map HOSTNAME to FELIX_FELIXHOSTNAME in calico-node.

### LINKED ISSUE #1105: calico-node and calico-libnetwork HOSTNAME not matching actual hostname means workload routes don't get added
So I've been assigning the HOSTNAME environment var to `calico-node` and `calico-libnetwork` from DNS SRV records to configure my environment.

These hostnames differ from the machine hostname which is unconfigured (or set transiently from DHCP), and the result seems to be that when I launch a workload (using the libnetwork plugin and docker), the blackhole entry for a network gets created ala
```
blackhole 192.168.125.0/26  proto bird
```
but the corresponding workload interface route does not get created and thus nothing can reach the workload nor can it communicate.

wrouesnel:
Setting `-e HOSTNAME` on the calico-container did not update the hostname
Felix receives, which leads to surprise breakages (blackhole routes work, but
interface routes don't).

Closes #1105

Note: this is untested just yet. I'm wondering if adding automated container integration tests is possible?

### DIFF
--- a/calico_node/filesystem/etc/service/felix/run
+++ b/calico_node/filesystem/etc/service/felix/run
@@ -1,5 +1,10 @@
 #!/bin/sh
 exec 2>&1
+# Felix doesn't understand HOSTNAME, but the container exports it as a common
+# interface. This ensures Felix gets the right hostname.
+if [ ! -z $HOSTNAME ]; then
+    export FELIX_FELIXHOSTNAME=$HOSTNAME
+fi
 export FELIX_ETCDADDR=$ETCD_AUTHORITY
 export FELIX_ETCDENDPOINTS=$ETCD_ENDPOINTS
 export FELIX_ETCDSCHEME=$ETCD_SCHEME


calico_node/filesystem/etc/service/felix/run
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 exec 2>&1
+# Felix doesn't understand HOSTNAME, but the container exports it as a common
+# interface. This ensures Felix gets the right hostname.
+if [ ! -z $HOSTNAME ]; then
+    export FELIX_FELIXHOSTNAME=$HOSTNAME
```
robbrockbank: You can remove the if statement and just do the assignment - simplifies the change.

robbrockbank:
Thanks @wrouesnel for this fix.  One minor comment in the PR.

Regarding testing - I've made some libcalico/calico-containers changes that test this particular use case and it does indeed fix the problem!  I'm mid-release so I'll get the tests updated _after_ the release, but I can get the actual fix into the release I'm working on.

calico_node/filesystem/etc/service/felix/run
wrouesnel: I considered that, but since Felix evolves independently it didn't seem wise to potentially set the environment variable to an empty string rather then leaving it unset when we're not trying to control it.

robbrockbank: That is true - and I'm happy to merge as is - that said Felix has to be resilient to empty strings for these parameters and it's unlikely that we'd change that behavior (i.e. treat "" differently from missing).

I've put up PRs to test this change in libcalico and calico-containers.

robbrockbank:
lgtm - margin - will be in the next release.

PR #1111, opened 2016-08-13
Add test for hostname override

robbrockbank:
Requires #1106 

This is a test for #1106


djosborne:
For my own knowledge, how does this even work when we "lie" about the hostname? Do we only  use hostname as uuid, and not for resolution of any kind?

robbrockbank:
@djosborne: The hostname is used by Calico just as a way to uniquely identify the host, so that the various components know where to read/write host specific configuration from the etcd datastore.

djosborne:
Nice, LGTM

PR #1117, opened 2016-08-17
Update policy-controller + calico/node version in k8s docs

caseydavenport:


djosborne:
@caseydavenport Code looks good, but does this need a runthrough test?

caseydavenport:
@djosborne I don't think it needs a full test.  Might be worth someone running through the vagrant guide + stars demo or something to do a smoke test, but we shouldn't need to test _all_ of the guides.

@gunjan5 Care to run a quick Calico + Kubernetes test? I can help.

gunjan5:
@caseydavenport : just saw the mention, I'll try it out when I get home

gunjan5:
I tested it and it worked fine! LGTM :)

PR #1119, opened 2016-08-18
Replace all etcd nodes in confd args

### LINKED ISSUE #1118: /etc/rc.local for the calico-node container has a sed bug
This line 

```
    ETCD_ENDPOINTS_CONFD=`echo "-node=$ETCD_NODE" | sed -e 's/,/ -node=/'`
```
in https://github.com/projectcalico/calico-containers/blob/master/calico_node/filesystem/etc/rc.local#L34

should read

```
ETCD_ENDPOINTS_CONFD=`echo "-node=$ETCD_NODE" | sed -e 's/,/ -node=/g'`
```

so that the substitution works for more than two etcd nodes.

caseydavenport:
Fixes #1118


matthewdupre:
LGTM

PR #1125, opened 2016-08-22
Get runit from community rather than testing

caseydavenport:
4 Days ago, runit was moved from `testing` to `community`: https://patchwork.alpinelinux.org/patch/2333/

This broke our builds for `calico/node`.


tomdee:
LGTM

Issue #1066, opened 2016-07-14
calicoctl node can fail to detect successful docker image startup


DreadPirateShawn:
We occasionally hit a scenario where "calicoctl node" appears to hang, but we can see that the calico-node image has in fact started successfully.

We run `calicoctl node --ip=192.168.0.68 --detach=true` and we see the Calico process start just fine, we see the "calico-node" appear in "docker ps", but the `calicoctl node` process itself never completes.

Calico processes:

```
$ ps aux | grep calico
highland  2693  0.0  0.0   4408   608 pts/1    Ss+  08:03   0:00 sh -c sudo /opt/skytap/bin/skytap-service calico start
root      2694  0.0  0.1 109824  3172 pts/1    S+   08:03   0:00 sudo /opt/skytap/bin/skytap-service calico start
root      2695  0.0  0.0  12588  1796 pts/1    S+   08:03   0:00 /bin/bash /etc/init.d/calicoctl start
root      2870  0.0  0.0   9716  1788 pts/1    S+   08:03   0:00 /opt/calico/bin/calicoctl node --ip=192.168.0.68 --detach=true
root      2872  0.0  0.8 107076 17524 pts/1    S+   08:03   0:00 /opt/calico/bin/calicoctl node --ip=192.168.0.68 --detach=true
root      2948  0.0  0.0    748     4 ?        S    08:03   0:00 svlogd -tt /var/log/calico/bird6
root      2949  0.0  0.0    640   252 ?        S    08:03   0:05 bird6 -R -s /var/run/calico/bird6.ctl -d -c /etc/calico/confd/config/bird6.cfg
root      2951  0.0  0.0    752     4 ?        S    08:03   0:00 svlogd /var/log/calico/felix
root      2952  0.0  0.0    748     4 ?        S    08:03   0:00 svlogd -tt /var/log/calico/bird
root      2953  0.0  0.0    748     4 ?        S    08:03   0:00 svlogd /var/log/calico/confd
root      2956  0.0  0.0    716   516 ?        S    08:03   0:07 bird -R -s /var/run/calico/bird.ctl -d -c /etc/calico/confd/config/bird.cfg
root      2957  0.1  1.2 115028 24656 ?        Sl   08:03   0:53 /usr/bin/python /usr/bin/calico-felix
[...]

PR #1137, opened 2016-08-31
Buffer line when looking for startup message.

fasaxc:
Speculative fix for #1066.


matthewdupre:
@fasaxc the linked issue describes a tentative patch not ready for merge.  Is this that, or intended to be merged?

fasaxc:
@matthewdupre Sorry, poor communication on my part.  This is the cleaned up tentative patch.  If the hunch in the issue was correct, this should fix it.  If the hunch wasn't correct, this should make no difference.

matthewdupre:
OK thanks: merging.  Let's see if it makes it better.

PR #1144, opened 2016-09-07
Wait for etcd before continuing in node startup.

caseydavenport:
This gates `calico/node` startup on etcd.  If etcd isn't present when the node container is started, then we wait for it.  This doesn't affect the case where etcd goes down after a while of normal execution.

The idea here is to prevent a number of race condition scenarios when starting a cluster where etcd is not yet available when the node container is started (but will be soon after).


matthewdupre:
LGTM

PR #1145, opened 2016-09-07
adjustments to improve the tutorial experience

mikev:
These were changes necessary to pass manual testing for the release.

Reference to coreos release.  Coreos alpha and stable are quite near in time, but the stable release works with Calico and will provide a better customer experience.
https://coreos.com/releases/


matthewdupre:
LGTM

Issue #1138, opened 2016-08-31
Ability to customize export filters is needed


adidenko:
Right now export filter is quite restrictive as it allows to export only IP pools defined in Calico. As a cloud operator I may want to reuse existing Bird from calico/node to also export some additional routes needed for my infrastructure. For instance, export routes for all 'tap*' interfaces or for some specific IP addresses which are not a part of Calico pool.

caseydavenport:
@neiljerram I believe you've been involved here, so assigning to you!

robbrockbank:
I believe this has been merged some time ago, and custom filters can now be applied writing to the relevant etcd location.  

@adidenko :  Please re-open if you believe this is not the case.

Regarding doc's/calicoctl control of this - I think this is a broader question about how we handle additional BGP configuration, and is dependent on the BGP backend that we use.  Needs some thought - but I'll raise a different issue for that.

PR #1151, opened 2016-09-13
Add support for custom export filters

adidenko:
This patch allows operators to put custom export filters into
`/calico/bgp/v1/global/custom_filters/v4` and
`/calico/bgp/v1/global/custom_filters/v6` directories as a plain text
keys. Those keys will be included (as plain text) into
custom_filters() functions.

Closes issue #1138


neiljerram:
@adidenko This PR code looks plausible to me, but there are some ST failures in the semaphoreci build above.  Have you looked into those?

adidenko:
Not yet, I want to build a custom image with this patch and test it on the lab. I'll post an update here.

cutwater:
I've built container based on latest `calico-containers` master with this patch applied and custom filters seems to be working.

adidenko:
OK, we've found the problem with this patch. Working on it, please do not merge.

adidenko:
I've also added https://github.com/projectcalico/libcalico/pull/150
Guys, pls review.

calico_node/filesystem/etc/calico/confd/config/custom_filters.cfg
```diff
@@ -0,0 +1,5 @@
+# Generated by confd
```
neiljerram: There aren't any other files in calico_node/filesystem/etc/calico/confd/config/, so I would guess that this file should not be checked into Git.  (Unless there is a reason why different handling is needed for this file, than all of the other confd-generated files?  In which case please explain that.)

calico_node/filesystem/etc/calico/confd/config/custom_filters6.cfg
```diff
@@ -0,0 +1,5 @@
+# Generated by confd
```
neiljerram: (Same for this file.)

calico_node/filesystem/etc/calico/confd/config/custom_filters.cfg
adidenko: This file is a placeholder to not break BIRD if we don't yet have needed paths/dirs in Etcd. So this placeholder will be shipped along with container. But if/when those paths/dirs are created in Etcd, then this file is going to be re-written by Confd.

adidenko: @neiljerram here's what I'm talking about (errors stopped when I created needed directories in etcd):
http://pastebin.com/Btu2Ethf

So without placeholders that would cause error in BIRD since `/etc/calico/confd/config/custom_filters.cfg` and `/etc/calico/confd/config/custom_filters6.cfg` do not exist.

neiljerram: OK, so according to the discussion at https://github.com/kelseyhightower/confd/issues/216, it seems you should use the `exists` keyword to handle this case.  Could you look at doing that instead?

adidenko: @neiljerram it's possible to use `exists` in templates. But this errors comes from `.toml` config file, not from template. AFAIK, it's not possible to bypass this in toml.

neiljerram: It's possible that I do not fully understand the details here, but I think there are now 2 reasons why we should not need to have `calico_node/filesystem/etc/calico/confd/config/custom_filters.cfg` and `calico_node/filesystem/etc/calico/confd/config/custom_filters6.cfg` in this repo.

1. We have merged https://github.com/projectcalico/libcalico/pull/150 - which means that the etcd paths `/calico/bgp/v1/global/custom_filters/v4` and `/calico/bgp/v1/global/custom_filters/v6` _will_ exist.

2. If (1) was not true, it should still be possible in principle to modify `calico_node/filesystem/etc/calico/confd/conf.d/custom_filters.toml` and `calico_node/filesystem/etc/calico/confd/templates/custom_filters.cfg.template` (and the 6 versions) to use `exists` for the `custom_filters` dir.  (I guess you would need to change `prefix` from `/calico/bgp/v1/global/custom_filters` to `/calico/bgp/v1/global/`.)

With either (1) or (2), execution of `calico_node/filesystem/etc/calico/confd/conf.d/custom_filters.toml` should proceed without error and generate `/etc/calico/confd/config/custom_filters.cfg`, so it should not be necessary to have a copy of `/etc/calico/confd/config/custom_filters.cfg` in this repo.

Am I missing something?

adidenko: You're right about (2). But this will be not an optimal solution from the performance point of view, since confd will have to watch a wider Etcd subtree that it actually needs for custom filters. That's why I used more precise keys. It may be not a problem if you don't expect a lot of changes in `/calico/bgp/v1/global/` though. So if you confirm it's OK, I'll update the patch.

As for (1) is merged. Let me update this patch and see how CI goes.

neiljerram:
@adidenko Sorry about the semaphoreci 'no space left on device' issue - I've reported it to my team

lwr20:
Ok - we've got a speculative fix in for the 'no space left on device' issue.

lwr20:
retest this please

adidenko:
Rebased.

neiljerram:
Thanks @adidenko, LGTM now.

neiljerram:
@zeca-pereira I'm sorry, I'm afraid Calico temporarily lost its support for custom filters when we transitioned to the v3 data model (aka Calico v3).  I presume you are using Calico v3 here?

We are planning to reinstate equivalent support, but don't have a clear schedule for that right now.  In the meantime, you could try editing the installed templates to re-add the necessary `include "custom_filters.cfg";` and `custom_filters();` lines.  (To be clear though, I haven't tried that, so am not sure if that would be sufficient to get this working again.)

zeca-pereira:
@neiljerram I have that in place:

/etc/calico/confd/templates # `cat /etc/calico/confd/templates/bird.cfg.mesh.template | grep custom`
include "custom_filters.cfg";

/etc/calico/confd/templates # `cat /etc/calico/confd/config/custom_filters.cfg`
# Generated by confd
 function custom_filters ()
 {
 
 }

neiljerram:
@zeca-pereira D'oh, yes, that won't work because confd doesn't directly read the whole content of etcd any more.  It only sees particular Calico v3 data model keys that it needs for generating BGP peerings.

You could experiment with generating `/etc/calico/confd/config/custom_filters.cfg` in some different way.  That could be with a parallel, upstream (i.e. kelseyhightower) confd instance - although then you'd need something to avoid the two confd instances picking up each other's config - or it could be any other mechanism that will just keep `/etc/calico/confd/config/custom_filters.cfg` up to date with the CIDRs that you need.

zeca-pereira:
@neiljerram Thank you.  In my opinion, all that really matters is Kubernetes services.  It's great to advertise pod addresses directly, but consuming services is the ultimate production goal.  Having the ability to open up and advertise the ClusterIP range is very valuable, especially if it's dynamic (BGP based, node specific based on pod backing the service), allowing you to bypass legacy exposure modes (NodePort, etc).  It's a shame this hasn't been a priority for Calico, as I'm a big fan.  Cillium and others are now doing this.

Thank you again for the help!

neiljerram:
@zeca-pereira Thanks.  I've passed this on, and hope that we will be able to address this soon.

PR #1163, opened 2016-09-21
Improve etcd deployment

caseydavenport:


tomdee:
Squash and check if tolerations are still needed. Then LGTM

PR #1164, opened 2016-09-22
Remove cluster-service label, which doesn't jive with the addon manager

caseydavenport:


caseydavenport:
LGTT (Looks good to tom)

PR #1180, opened 2016-09-29
Libnetwork plugin bundled in calico/node container

gunjan5:
- The new behavior is if calico/node is started with `calicoctl` and `--libnetwork` flag is passed then instead of running libnetwork-plugin in a separate container, we run it inside calico/node container.

- if calico/node is not started with `calicoctl` (like in case of k8s or manually starting the container), libnetwork can be enabled by passing an env var `CALICO_LIBNETWORK_ENABLED`, which is `false` by default and doesn't run libnetwork unless it's set to `true`

### DIFF
--- a/calicoctl/calico_ctl/node.py
+++ b/calicoctl/calico_ctl/node.py
@@ -68,9 +68,8 @@ Options:
   --ip6=<IP6>               The local IPv6 management address to use.
   --ipv4                    Show IPv4 information only.
   --ipv6                    Show IPv6 information only.
-  --libnetwork              Use the libnetwork plugin.
-  --libnetwork-image=<LIBNETWORK_IMAGE_NAME>    Docker image to use for
-                            Calico's libnetwork driver.
+  --libnetwork              (Deprecated) Use the libnetwork plugin.
+  --libnetwork-image=<LIBNETWORK_IMAGE_NAME>    (Deprecated) This flag will be ignored.
                             [default: calico/node-libnetwork:latest]
   --log-dir=<LOG_DIR>       The directory for logs [default: /var/log/calico]
   --no-pull                 Prevent from pulling the Calico node Docker images.
@@ -233,8 +232,9 @@ def node(arguments):
         assert arguments.get("--detach") in ["true", "false"]
         detach = arguments.get("--detach") == "true"
 
-        libnetwork_image = None if not arguments.get("--libnetwork") \
-                                else arguments.get("--libnetwork-image")
+        # Set libnetwork_enabled to False if --libnetwork flag is not passed 
+        libnetwork_enabled = False if not arguments.get("--libnetwork") else True
+
         node_start(ip=arguments.get("--ip"),
                    node_image=arguments.get('--node-image'),
                    runtime=arguments.get("--runtime"),
@@ -242,13 +242,13 @@ def node(arguments):
                    ip6=arguments.get("--ip6"),
                    as_num=as_num,
                    detach=detach,
-                   libnetwork_image=libnetwork_image,
+                   libnetwork_enabled=libnetwork_enabled,
                    no_pull=arguments.get("--no-pull"),
                    backend=backend)
 
 
 def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
-               libnetwork_image, no_pull, backend):
+               libnetwork_enabled, no_pull, backend):
     """
     Create the calico-node container and establish Calico networking on this
     host.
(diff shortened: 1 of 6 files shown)


calicoctl/calico_ctl/node.py
```diff
@@ -68,8 +68,8 @@
   --ip6=<IP6>               The local IPv6 management address to use.
   --ipv4                    Show IPv4 information only.
   --ipv6                    Show IPv6 information only.
-  --libnetwork              Use the libnetwork plugin.
-  --libnetwork-image=<LIBNETWORK_IMAGE_NAME>    Docker image to use for
+  --libnetwork              (Deprecated) Use the libnetwork plugin.
+  --libnetwork-image=<LIBNETWORK_IMAGE_NAME>    (Deprecated) Docker image to use for
```
tomdee: Does this have any effect any more? If not then the docopt should reflect this

calicoctl/calico_ctl/node.py
```diff
@@ -260,8 +261,8 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
     the global default value will be used.
     :param detach: True to run in Docker's "detached" mode, False to run
     attached.
-    :param libnetwork_image: The name of the Calico libnetwork driver image to
-    use.  None, if not using libnetwork.
+    :param libnetwork_enabled: True to run libnetwork plugin inside calico-node.
+    False by default. None, if not using libnetwork.
```
tomdee: Remind me what the reasoning was behind defaulting this to false

calicoctl/calico_ctl/node.py
```diff
@@ -348,25 +349,24 @@ def node_start(node_image, runtime, log_dir, ip, ip6, as_num, detach,
...
     """
     Start the main Calico node container.
 
     :param ip:  The IPv4 address of the host.
     :param ip6:  The IPv6 address of the host (or None if not configured)
     :param as_num: The AS number for the host
     :param log_dir:  The log directory to use.
-    :param node_image:  The calico-node image to use.
+    :param libnetwork_enabled: True to run libnetwork plugin inside calico-node.
+    False by default. None, if not using libnetwork.
```
tomdee: This line is bad. You've already said it's defaulted to false at the point where it's defaulted. It's wrong to repeat that here.

calicoctl/calico_ctl/node.py
```diff
@@ -391,13 +391,20 @@ def _start_node_container_docker(ip, ip6, as_num, log_dir, node_image, detach, e
...
+
     environment = [
         "HOSTNAME=%s" % hostname,
         "IP=%s" % (ip or ""),
         "IP6=%s" % (ip6 or ""),
         "CALICO_NETWORKING_BACKEND=%s" % backend,
         "AS=%s" % (as_num or ""),
-        "NO_DEFAULT_POOLS=%s" % (no_default_pools or "")
+        "NO_DEFAULT_POOLS=%s" % (no_default_pools or ""),
+        "CALICO_LIBNETWORK_ENABLED=%s" % libnetwork_flag_str
```
tomdee: can't you just use %b or something and avoid the if/else above?

calicoctl/calico_ctl/node.py
gunjan5: It is ignored. Does appending `This flag will be ignored` to the docopts string sound sufficient?

calicoctl/calico_ctl/node.py
gunjan5: So it doesn't start libnetwork for the systems like kubernetes where libnetwork is not used, and the flag is not passed

calicoctl/calico_ctl/node.py
gunjan5: fixed

calicoctl/calico_ctl/node.py
gunjan5: python doesn't seem to have a `%b`, it has `%r` to format to binary `True` or `False` but it won't be consistent with the ENV vars we are setting (in all lower case). See flag `CALICO_NETWORKING` for example which also has to converted into `false` (string) from `False`(bool)

calicoctl/calico_ctl/node.py
tomdee: I think replacing "Docker image to use for.." with "This flag will be ignored" is the right thing to do

calicoctl/calico_ctl/node.py
tomdee: OK, sounds OK to me.

calicoctl/calico_ctl/node.py
tomdee: I don't see any change? I think line 369 should be removed.

calicoctl/calico_ctl/node.py
tomdee: OK, it's only a few lines so I don't feel too strongly.

tomdee:
Two more minor things - can you fix and squash and let me have another look?

calicoctl/calico_ctl/node.py
gunjan5: sorry I didn't push it. Wanted to get confirmation before pushing

gunjan5:
made those changes and squashed

tomdee:
LGTM - merging

PR #1181, opened 2016-09-30
First pass at changes to support etcd TLS for hosted install

caseydavenport:
- [x] Replace caseydavenport/cni image with calico/cni release.
- [x] Test this against real, secured etcd.

Relies on https://github.com/projectcalico/calico-cni/pull/182/files


docs/cni/kubernetes/manifests/calico-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+            "type": "k8s",
+            "k8s_api_root": "https://__KUBERNETES_SERVICE_HOST__:__KUBERNETES_SERVICE_PORT__",
+            "k8s_auth_token": "__SERVICEACCOUNT_TOKEN__"
+        },
+        "kubernetes": {
+            "kubeconfig": "__KUBECONFIG_FILEPATH__"
+        }
+    }
+
+  # If you're using TLS enabled etcd uncomment the following.
```
caseydavenport: @tomdee These are the parts someone will need to modify to make TLS happen.

docs/cni/kubernetes/manifests/calico-config.yaml
```diff
@@ -0,0 +1,60 @@
...
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: calico-etcd-secrets
+  namespace: kube-system
+data:
+  # Populate the following files with etcd TLS configuration if desired.
+  # This self-hosted install expects three files with the following names.
+  # etcd-key: null
```
tomdee: Why the use of null here? Are there better examples that you could give?

docs/cni/kubernetes/manifests/calico-hosted.yaml
```diff
@@ -52,10 +70,12 @@ spec:
             - mountPath: /var/run/calico
               name: var-run-calico
               readOnly: false
+            - mountPath: /calico-secrets
+              name: etcd-certs
         # This container installs the Calico CNI binaries
         # and CNI network config file on each node.
         - name: install-cni
-          image: calico/cni:latest
+          image: caseydavenport/cni:latest
```
tomdee: I know you already know you're going to change this before merge

docs/cni/kubernetes/manifests/calico-config.yaml
caseydavenport: The things that go here are base64 encoded cert files, so not super useful to stick an example in.

I could stick the default value for each as something like "replace-with-etcd-key" or similar.  WDYT?

tomdee: I'm happy with this as is, but as discussed a little more commentary might be nice. Up to you.

tomdee:
LGTM

PR #1187, opened 2016-10-05
[WIP] Add golang FVs

lwr20:
Relies on https://github.com/projectcalico/libcalico/pull/154


tests/st/test_pool.py
```diff
@@ -0,0 +1,112 @@
...
+    def test_pool_crud(self):
+        """
+        Test that a basic CRUD flow for pool commands.
+        """
+        with DockerHost('host', dind=False, start_calico=False) as host:
+            # Set up the ipv4 and ipv6 pools to use
+            ipv4_pool = "10.0.1.0/24"
+            ipv6_pool = "fed0:8001::/64"
+
+            ipv4_yaml = """
```
robbrockbank: I would probably define these as python dictionaries and write out the dictionaries as json.   The YAML parser handles JSON and YAML so using JSON should suffice for the most part.

tests/st/test_pool.py
```diff
@@ -0,0 +1,112 @@
...
+            host.calicoctl("create -f ipv6.yaml", new=True)
+            # And read it back out using the python calicoctl
+            pool_out = host.calicoctl("pool show")
+            # Assert output contains the ipv6 pool, but not the ipv4
+            self.assertNotIn(ipv4_pool, pool_out)
+            self.assertIn(ipv6_pool, pool_out)
+            self.assertNotIn("ipip", pool_out)
+
+            # Now read it out with the Go calicoctl too:
+            pool_out = host.calicoctl("get pool", new=True)
```
robbrockbank: I'd recommend using JSON output format and parsing into a dictionary - that way you can do more precise searches than just a simple text search.

tests/st/test_pool.py
lwr20: We've already got pyyaml as a requirement - makes sense to use dictionaries, but write out using pyyaml.

tests/st/test_pool.py
lwr20: Or yaml format and read in with pyyaml :)  Agree in principle though.

PR #1189, opened 2016-10-05
Allow to provide custom URL path for calico binaries

SergK:
We need to have an ability to provide custom URLs for calico
binaries to be downloaded from.
This patch allows to redefine the below variables:
  CONFD_URL
  BIRD_URL
  BIRD6_URL
  BIRDCL_URL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,15 @@ test: st ut              ## Run all the tests
 ssl-certs: certs/.certificates.created ## Generate self-signed SSL certificates
 
 ###############################################################################
+# URL for Calico binaries
+# confd binary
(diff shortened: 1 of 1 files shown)


neiljerram:
Why are these custom URLs needed?  In the long term it would be better to incorporate any fixes needed into our own builds - or even better into the upstream versions.

SergK:
Sometime we need to test binaries (bird, confd, etc) which were built from the code that is not yet merged and is under CI, so we need to have an ability to provide the path to custom URLs. Definitely this pull-request doesn't change the default behavior.

Sure we can make wget outside of Makefile, but this not correct way from make system, since modification time of targets can be "older" and Makefile will re-build target, using hard-coded urls

tomdee:
LGTM

PR #1192, opened 2016-10-05
Add ability to build calico/ctl image

SergK:
Add new make target ctl_image, which builds calico_ctl image
consuming the results from binary target

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ birdcl
 .vagrant
 gobgp
 calico_node/bin
+calicoctl/calicoctl
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,11 @@ clean_calico_node:
 ###############################################################################
 CALICOCTL_DIR=calicoctl
 CALICOCTL_FILE=$(CALICOCTL_DIR)/calicoctl.py $(wildcard $(CALICOCTL_DIR)/calico_ctl/*.py) calicoctl.spec
+CALICOCTL_URL?=http://www.projectcalico.org/builds/calicoctl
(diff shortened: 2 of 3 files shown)


neiljerram:
@tomdee Please could you review this?  As I understand it, there are a couple of things changing in this PR:

- It adds a Makefile target for building the calico/ctl container.  (For which I believe there is currently no equivalent, even though we do already have a Dockerfile for this.)

- The content of the calico/ctl container will be whatever is in the Git repo where the build happens.  Whereas previously it did a download from http://www.projectcalico.org/builds/calicoctl - so we had a two-step process where a build of the calico/ctl container might use older calicoctl code.

Do those changes make sense?  Please let us know your thoughts.

SergK:
@neiljerram thank you for commenting on this. Let me a little bit describe, why we can win from this change.

* Yes we add new target `ctl_image` which builds calico/ctl container, for now afaik we don't have this target and the only way to build is to run:

`docker build -t calico/ctl calicoctl`

but in this case we are not able to provide some custom `calicoctl` file (built from code that is not yet merged and tested on ci) because we always download it from http://www.projectcalico.org/builds/calicoctl

* Probably we can add some flag (if condition) in Makefile which *by default* downloads by wget calicoctl from http://www.projectcalico.org/builds/calicoctl
so in this case there will be no impact on default behavior

SergK:
@neiljerram @tomdee I preserved the default behavior for make, please see my latest changes

tomdee:
This is a really useful change, thanks @SergK I'm super busy right now but I'll try to give this a proper review later today.

tomdee:
I just pulled in the first commit to a new PR and merged that #1202

PR #1204, opened 2016-10-07
Remove demo files, add redirects to new site.

djosborne:


docs/cni/kubernetes/vagrant-coreos/VagrantCoreOS.md
```diff
@@ -8,110 +8,9 @@
```
matthewdupre: This master only banner references a non-existent image.  We should probably just remove the "latest version of docs" banner(s, if there are any more).

docs/cni/kubernetes/vagrant-coreos/VagrantCoreOS.md
```diff
@@ -8,110 +8,9 @@
...
-## 3. Next Steps
-Now that you have a cluster with `kubectl` configured, you can [install Calico and other cluster addons](../InstallAddons.md).
-
-[calico-networking]: https://github.com/projectcalico/calico-containers
-[calico-cni]: https://github.com/projectcalico/calico-cni
-[virtualbox]: https://www.virtualbox.org/
-[vagrant]: https://www.vagrantup.com/downloads.html
-[git]: http://git-scm.com/
 
 [![Analytics](https://calico-ga-beacon.appspot.com/UA-52125893-3/calico-containers/docs/cni/kubernetes/VagrantCoreOS.md?pixel)](https://github.com/igrigorik/ga-beacon)
```
matthewdupre: The path in this analytics URL is wrong.  Do we need analytics URLs in any of these anyway?

djosborne: URL looks correct to me. It isn't supposed to be a full, valid URL to the file - analytics just reads it as a string and logs the view.

docs/cni/kubernetes/vagrant-coreos/VagrantCoreOS.md
djosborne: I didn't remove / modify any images. This was broken before I go to it. Let's keep it out of scope for this PR.

djosborne:
@matthewdupre 

>but what about the user-data / Vagrant files (docs/calico-with-docker/docker-network-plugin/cloud-config and such)?
>
>Is calico-containers/docs the right home for them?

No, which is why I moved all of them to the new `projectcalico/calico` site. Did I miss some?

-------------------------------------------------------------------------------------------------------------------------

I don't know which images are / aren't used in the docs, and don't really have the time to go searching, so can we just leave them in?

djosborne:
ping @matthewdupre

docs/cni/kubernetes/vagrant-coreos/VagrantCoreOS.md
matthewdupre: You see the extra 'vagrant-coreos' directory right?  If it's a string then that's fine for now, but we should be keeping these correct; even if just to save the time of going through this conversation every time someone discovers it (which is at least every containers release).

matthewdupre:
On cloud config files:
- I gave an example above: docs/calico-with-docker/docker-network-plugin/cloud-config/*.  If you want to find the rest, what I did was `ls -Rl | less` and look for files with size >800 and not 4096 (i.e. unmoved documents).

As for which images are used in the docs (meaning here); surely the answer should be "None" (modulo Building.md), so we can remove all of them?

djosborne:
@matthewdupre 

Removed remaining files / images. LGTM?

docs/cni/kubernetes/vagrant-coreos/VagrantCoreOS.md
matthewdupre: I promise you it wasn't intentional, but we can also remove it later.

PR #1228, opened 2016-10-21
Field renaming and docs tidy

robbrockbank:


calicoctl/commands/node/diags.go
```diff
@@ -84,7 +92,7 @@ func runDiags(logDir string) {
 	fmt.Println("Collecting diagnostics")
 
 	// Create a temp directory in /tmp
-	tmpDir, err := ioutil.TempDir("/tmp", "tmp")
+	tmpDir, err := ioutil.TempDir("", "calico")
```
gunjan5: good catch!

Makefile
```diff
@@ -191,7 +191,8 @@ test-containerized: run-etcd $(BUILD_CONTAINER_MARKER)
 	docker run -ti --rm --privileged --net=host \
 	-e PLUGIN=calico \
 	-v ${PWD}:/go/src/github.com/projectcalico/calico-containers:rw \
-	$(BUILD_CONTAINER_NAME) make ut
+	$(BUILD_CONTAINER_NAME) make ut; \
```
gunjan5: This line should be `$(BUILD_CONTAINER_NAME) bash -c 'make ut; \` 
That should fix the CircleCI build

robbrockbank:
The `--version` issue should really be another PR, but I've fixed up here since it was easy.  I have fixed up to only allow `--version` from the  root calicoctl command, not the sub commands.

It returns a version summary string, much like Dockers.

So, `calicoctl --version` will work, but `calicoctl apply --version` will print a usage statement.

PR #1229, opened 2016-10-24
Fix up IPAM commands to be hierarchical

robbrockbank:
As per node and calicoctl, I've moved the IPAM commands into a sub directory so that each command has it's own help.  It made sense to do this while I'm busy fixing up the docs.

I've put a couple of things in a common module in calicoctl - which I will probably eventually move to be in more "functionally named" modules - but IMHO that can wait for the time being - I needed to commonize some of the code to prevent circular imports.

### DIFF
--- a/calicoctl/commands/apply.go
+++ b/calicoctl/commands/apply.go
@@ -20,11 +20,11 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/docopt/docopt-go"
 
-	"github.com/projectcalico/calico-containers/calicoctl/commands/common"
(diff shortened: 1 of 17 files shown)


gunjan5:
BTW, as a side note: https://blog.golang.org/package-names
>Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused.

We should come up with a better package name in future :)

gunjan5:
LGTM

PR #1232, opened 2016-10-27
Fix release scripts

gunjan5:
Update release scripts to account for the fact that:
- `calico/node` doesn't use `build.sh` anymore
- `projectcalico/calico` was renamed to `projectcalico/felix`

### DIFF
--- a/release-scripts/utils.py
+++ b/release-scripts/utils.py
@@ -30,7 +30,6 @@ from calico_ctl import __version__
 # Path names relative to the root of the project
 PATH_CALICOCTL_NODE = path.join("calicoctl", "calico_ctl", "node.py")
 PATH_CALICOCTL_INIT = path.join("calicoctl", "calico_ctl", "__init__.py")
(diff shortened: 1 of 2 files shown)


djosborne:
As discussed, I'm going to take over this PR

release-scripts/utils.py
```diff
@@ -30,7 +30,6 @@
 # Path names relative to the root of the project
 PATH_CALICOCTL_NODE = path.join("calicoctl", "calico_ctl", "node.py")
 PATH_CALICOCTL_INIT = path.join("calicoctl", "calico_ctl", "__init__.py")
-PATH_CALICONODE_BUILD = path.join("calico_node", "build.sh")
```
djosborne: Oops. didn't notice that `PATH_MAKEFILE` is included in `UPDATE_FILES_STATIC`. So this script *does* already correctly update those version pins over in the Makefile.

Nice one @robbrockbank  :)

PR #1236, opened 2016-11-01
Add config commands and node as a resource

robbrockbank:
This PR adds:
-  Node as a resource
-  System wide configuration options (and low level node configuration options).

It's possible that we'll make config a "resource" as well and that some of the node-specific config options could move into the node resource - but I think that can be done later with pretty minor deprecation and phase out.

I've also gone through the various command line helps and fixed the width at 79 characters because it was too long and uneven.  Should make our docs nicer once I cut and paste it into our docs.

### DIFF
--- a/calicoctl/commands/config.go
+++ b/calicoctl/commands/config.go
@@ -19,19 +19,20 @@ import (
 	"os"
 	"strings"
 
-	"github.com/docopt/docopt-go"
-	"github.com/projectcalico/libcalico-go/lib/client"
-	"github.com/projectcalico/calico-containers/calicoctl/commands/constants"
-	"github.com/projectcalico/calico-containers/calicoctl/commands/clientmgr"
 	"fmt"
-	"strconv"
+
+	"github.com/docopt/docopt-go"
+	"github.com/projectcalico/calico-containers/calicoctl/commands/clientmgr"
+	"github.com/projectcalico/calico-containers/calicoctl/commands/constants"
+	"github.com/projectcalico/libcalico-go/lib/client"
+	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
 
 func Config(args []string) {
 	doc := constants.DatastoreIntro + `Usage:
   calicoctl config set <NAME> <VALUE> [--node=<NODE>] [--config=<CONFIG>]
   calicoctl config unset <NAME> [--node=<NODE>] [--config=<CONFIG>]
-  calicoctl config view <NAME> [--node=<NODE>] [--config=<CONFIG>]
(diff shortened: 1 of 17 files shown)


calicoctl/commands/config.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/projectcalico/calico-containers/calicoctl/commands/clientmgr"
+	"fmt"
+	"strconv"
+)
+
+func Config(args []string) {
+	doc := constants.DatastoreIntro + `Usage:
+  calicoctl config set <NAME> <VALUE> [--node=<NODE>] [--config=<CONFIG>]
+  calicoctl config unset <NAME> [--node=<NODE>] [--config=<CONFIG>]
+  calicoctl config view <NAME> [--node=<NODE>] [--config=<CONFIG>]
```
fasaxc: Feels inconsistent with "show" for other commands, can you check that we're matching k8s here or bring it into line with our other commands?

calicoctl/commands/config.go
```diff
@@ -0,0 +1,319 @@
...
+	c client.ConfigInterface
+}
+
+
+func (a asnum) set(value, node string) error {
+	if node != "" {
+		return errors.New("--node should not be specified")
+	}
+
+	asn, err := strconv.ParseUint(value, 10, 64)
```
fasaxc: WIBNI this allowed the nice format of AS number

calicoctl/commands/resources.go
```diff
@@ -100,6 +100,12 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
 	orchestrator := argStringOrBlank(args, "--orchestrator")
 	resScope := argStringOrBlank(args, "--scope")
 	switch strings.ToLower(kind) {
+	case "nodes":
```
fasaxc: Can't you do `case "nodes", "node"`, might look neater than a fallthrough.

calicoctl/commands/config.go
robbrockbank: k8s view returns the whole set of config (presumably as yaml) which you can then filter using a golang template.

How about we stick with get - like the other commands.   We can update later perhaps to take no parameters - in which case it'll return all config as a YAML doc.

calicoctl/commands/config.go
robbrockbank: I think that should be relatively easy to parse - so i'll see if i can knock that off.

calicoctl/commands/resources.go
robbrockbank: oooh lovely - if it works.

robbrockbank:
You can ignore semaphore warnings for now since we don't have a semaphore make target.

PR #1243, opened 2016-11-02
Rename pool -> ipPool

robbrockbank:
And also match interface changes in libcalico.


robbrockbank:
There is also a minor change to make the main calicoctl help 79 chars long (and it was missing the config option).  I've added this as a separate commit - so no need to merge commit.

PR #1251, opened 2016-11-03
libcalico is no longer requered for node image

SergK:
check related commit: 7d0f4abad31e8e3edc24915aa3efcc546f585345


SergK:
@tomdee could you please check this patch as well?

tomdee:
I think it's actually still used for the startup.py scripts - I can double check tomorrow.

PR #1254, opened 2016-11-03
Work around Felix hanging on empty DATASTORETYPE

matthewdupre:
https://github.com/projectcalico/felix/issues/1156 has the Felix side.  This fix alone should restart calico-node to normality.


caseydavenport:
One comment, but feel free to ignore and just merge it.

matthewdupre:
I think given the current time pressure, let's just merge.  Thanks

Issue #1271, opened 2016-11-05
calicoctl doesn't show global default ASN.


caseydavenport:
```
macncheese:calico-containers casey$ ./dist/calicoctl get nodes -o wide
NAME          ASN     IPV4            IPV6
k8s-master    <nil>   172.18.18.101   <nil>
k8s-node-01   <nil>   172.18.18.102   <nil>
k8s-node-02   <nil>   172.18.18.103   <nil>
```

It should populate ASN with the global default rather than <nil> when none is specified for a node.

caseydavenport:
CC @robbrockbank

caseydavenport:
repurposed this issue just for the ASN bug (now that the nil stuff is fixed by #1283)

robbrockbank:
I think this will be fixed as part of a larger rework of how we display inherited per-node configuration.

robbrockbank:
First iteration, print "using global value"    rather than blank
(maybe) Update node status to display effective value.

PR #1283, opened 2016-11-07
Don't display nil when viewing node resources with the --wide option

robbrockbank:
Fixes part of #1271 (but doesn't fix the displaying default ASN when ASN is not specified)

This also fixes the Makefile not rebuilding after changing calicoctl files.


caseydavenport:
LGTM

PR #1285, opened 2016-11-07
Fix the run help and typo in command

robbrockbank:


calicoctl/commands/node/run.go
```diff
@@ -89,15 +89,7 @@ Options:
 
 Description:
   This command is used to start a Calico node container instance.  The
```
caseydavenport: I think "Calico node" should be "calico/node" right?

robbrockbank: yup - i've tweaked slightly and reduced the text.

robbrockbank:
@caseydavenport since you've already looked - want to give it a sign off?

PR #1290, opened 2016-11-08
Tweaks to node run text

robbrockbank:


calicoctl/commands/node/run.go
```diff
@@ -69,8 +69,7 @@ Options:
      --log-dir=<LOG_DIR>   The directory containing Calico logs.
                            [default: /var/log/calico]
      --node-image=<DOCKER_IMAGE_NAME>
-                           Docker image to use for Calico's
-                           per-node container.
+                           Docker image to use for Calico's per-node container.
                            [default: calico/node:%s]
```
gunjan5: this should be `latest` not `%s`

gunjan5:
@robbrockbank I think the node run command is trying to pull an image that doesn't exist now since the latest image was tagged this morning, breaking the demo
it's trying to pull `calico/node:v1.0.0-beta-8-gb43719b` and fails because that's not a valid tag

gunjan5:
as discussed, let's have a separate issue/PR for the image version issue, which is outside of the scope of this PR. This one is ok to merge

PR #1292, opened 2016-11-08
Add 'libltdl' library into 'calico/ctl' docker image

### LINKED ISSUE #1291: `calicoctl node run` fails while executing inside container: error while loading shared libraries: libltdl.so.7
When I try to run calico/node docker container using new Golang `calicoctl` (which is also running with docker) I get the following error:

> docker: error while loading shared libraries: libltdl.so.7: cannot open shared object file: No such file or directory

More info here: http://paste.openstack.org/show/588375/

To fix the issue I have to install **libltdl** into container:

`apk update && apk add libltdl`

Most probably this bug will be gone if https://github.com/projectcalico/calico-containers/issues/1280 is fixed.

artem-panchenko:
Closes #1291

### DIFF
--- a/Dockerfile.calicoctl
+++ b/Dockerfile.calicoctl
@@ -5,4 +5,7 @@ ADD dist/calicoctl ./calicoctl
 
 ENV CALICO_CTL_CONTAINER=TRUE
 
+# libltdl.so is needed by docker command line tool
+RUN apk add --no-cache libltdl
+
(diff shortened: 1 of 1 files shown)


tomdee:
Can you add a comment above the new line you added to explain why it's needed? Otherwise someone will just remove it in future if they don't understand what it's for.

PR #1306, opened 2016-11-15
Standardize alpine image bases, use single etcd image and use quay.io more

tomdee:


tomdee:
Closing - replaced by #1409 and #1408

PR #1311, opened 2016-11-17
Add a selection of resource aliases

### LINKED ISSUE #1308: Support shorthand for calicoctl resources
Would be really nice to support shorthand for at least the longer calicoctl resources.

e.g

```
calicoctl get workloadendpoint
```
vs
```
calicoctl get wep
calicoctl get hep
calicoctl get pro
calicoctl get pol
calicoctl get bp
calicoctl get no
...
```

As much as I'd like this to be v1.0.0 worthy, I don't think it actually is (though it's really not that much work.. so if someone wants to make me really happy...)

robbrockbank:
Fixes #1308

Two things to note:
-  This will need documenting - probably just in the resources description sections (and perhaps a mention in our usage guides) ... i'm treating this as low priority for now.
-  This does not affect the reading/writing of YAMl files - the exact resource name is required for that.  Does k8s allow for alternate names in the YAML for the "kind" - if so we probably will want to have another issue - or just reject this PR (since I'll implement it in a different way - but deffo not in time for 2.0.0)

### DIFF
--- a/calicoctl/commands/resources.go
+++ b/calicoctl/commands/resources.go
@@ -101,31 +101,31 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
 	orchestrator := argutils.ArgStringOrBlank(args, "--orchestrator")
 	resScope := argutils.ArgStringOrBlank(args, "--scope")
 	switch strings.ToLower(kind) {
-	case "node", "nodes":
+	case "node", "nodes", "no", "nos":
 		p := api.NewNode()
 		p.Metadata.Name = name
 		return *p, nil
-	case "hostendpoint", "hostendpoints":
+	case "hostendpoint", "hostendpoints", "hep", "heps":
 		h := api.NewHostEndpoint()
 		h.Metadata.Name = name
 		h.Metadata.Node = node
 		return *h, nil
-	case "workloadendpoint", "workloadendpoints":
+	case "workloadendpoint", "workloadendpoints", "wep", "weps":
 		h := api.NewWorkloadEndpoint()
 		h.Metadata.Name = name
 		h.Metadata.Orchestrator = orchestrator
 		h.Metadata.Workload = workload
 		h.Metadata.Node = node
 		return *h, nil
-	case "profile", "profiles":
+	case "profile", "profiles", "pro", "pros":
 		p := api.NewProfile()
 		p.Metadata.Name = name
 		return *p, nil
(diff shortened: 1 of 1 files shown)


calicoctl/commands/resources.go
```diff
@@ -101,31 +101,32 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
...
 		p := api.NewProfile()
 		p.Metadata.Name = name
 		return *p, nil
-	case "policy", "policies":
+	case "policy", "policies", "pol", "pols":
 		p := api.NewPolicy()
 		p.Metadata.Name = name
 		return *p, nil
-	case "ippool", "ippools":
+	case "ippool", "ippools", "ipp", "ipps", "ip", "ips":
```
caseydavenport: I think `ip` and `ips` aren't good choices here given that an IP address is a real thing that might exist in our data model.

Maybe `pool` though?

caseydavenport:
> This does not affect the reading/writing of YAMl files - the exact resource name is required for that. Does k8s allow for alternate names in the YAML for the "kind" - if so we probably will want to have another issue - or just reject this PR (since I'll implement it in a different way - but deffo not in time for 2.0.0)


No, the `kind` is part of an API and should be a single value.

These are just convenient aliases.

calicoctl/commands/resources.go
```diff
@@ -101,31 +101,32 @@ func getResourceFromArguments(args map[string]interface{}) (unversioned.Resource
...
 		p.Metadata.Name = name
 		return *p, nil
-	case "hostendpoint", "hostendpoints":
+	case "hostendpoint", "hostendpoints","hostep", "hosteps", "hep", "heps":
 		h := api.NewHostEndpoint()
 		h.Metadata.Name = name
 		h.Metadata.Node = node
 		return *h, nil
-	case "workloadendpoint", "workloadendpoints":
+	case "workloadendpoint", "workloadendpoints", "workloadep", "workloadeps",
```
caseydavenport: Personally I don't see the value in the only partially abbreviated versions (workloadep, workloadeps).

I'd be happy with `wep` and `weps`, but that's just me.

caseydavenport: I guess it doesn't hurt to have more!

robbrockbank: ok i'll keep it simple.

caseydavenport:
LGTM

PR #1313, opened 2016-11-18
pin libcalico-go latest

gunjan5:
for bugfix: https://github.com/projectcalico/libcalico-go/issues/258
"IP's allocated from original pool after deletion"

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -19,6 +19,6 @@ import:
 - package: github.com/termie/go-shutil
 - package: github.com/mitchellh/go-ps
 - package: github.com/projectcalico/libcalico-go
-  version: 1.0.0-beta
+  version: c3038033665942c39cf958a6e145267eb1157ca3
   subpackages:
   - lib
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -19,6 +19,6 @@ import:
 - package: github.com/termie/go-shutil
 - package: github.com/mitchellh/go-ps
 - package: github.com/projectcalico/libcalico-go
-  version: 1.0.0-beta
```
robbrockbank: For master is it not better to remove the version lock completely so that we can easily regenerate the lock file without needing to look up the latest release?

robbrockbank:
lgtm, but would be nice to address comment (not necessary for this PR though if you're in a hurry)

glide.yaml
gunjan5: hmmm it makes sense, not sure if we do that for our other repos?

gunjan5:
closing this in favor of #1314

PR #1314, opened 2016-11-18
Use latest libcalico-go

robbrockbank:
In particular this gets:

-  Profile tags moved to Metadata
-  Updated IP pool and Node deletion to remove associated IPAM data


glide.lock
```diff
@@ -247,7 +245,7 @@ imports:
   - pkg/apis/extensions/v1beta1
   - pkg/apis/policy
   - pkg/apis/policy/install
-  - pkg/apis/policy/v1alpha1
+  - pkg/apis/policy/v1beta1
```
gunjan5: just curious, why is there an update for the policy version?

robbrockbank: No idea - the lock file was re-generated from scratch

PR #1316, opened 2016-11-20
Reinstate STs, update existing STs and node run improvements

### LINKED ISSUE #1257: `calicoctl node run` cmd should honor CALICO_LIBNETWORK_ENABLED ENV var
command hardcodes `CALICO_LIBNETWORK_ENABLED` to `true` right now

For post `v2.0.0-beta`

### LINKED ISSUE #1294: calicoctl starts node even without etcd connection
Seems `calicoctl` no longer fails to start node if it can't connect to etcd. When launched, we're left with a immediately dead container, and a return code of `0`.

```
turk:calico/ (master✗) $ sudo calicoctl node run                                              [10:17:52]
Running command to load modules: modprobe -a xt_set ip6_tables
Enabling IPv4 forwarding
Enabling IPv6 forwarding
Increasing contrack limit
Running the following command:

docker run -d --net=host --privileged --name=calico-node -e HOSTNAME=durk -e IP6= -e CALICO_NETWORKING_BACKEND=bird -e ETCD_SCHEME=http -e CALICO_LIBNETWORK_ENABLED=true -e IP= -e AS= -e NO_DEFAULT_POOLS= -e ETCD_AUTHORITY=127.0.0.1:2379 -e ETCD_ENDPOINTS= -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock -v /var/log/calico:/var/log/calico -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules calico/node:v1.0.0-beta

turk:calico/ (master✗) $ echo $?                                                              [10:17:54]
0
turk:calico/ (master✗) $ docker ps -a                                                         [10:17:56]
CONTAINER ID        IMAGE                     COMMAND             CREATED             STATUS                     PORTS               NAMES
630d945c26cf        calico/node:v1.0.0-beta   "start_runit"       5 seconds ago       Exited (1) 4 seconds ago                       calico-node
turk:calico/ (master✗) $ docker logs calico-node                                              [10:17:59]
  File "<string>", line 346, in <module>
  File "<string>", line 297, in main
Waiting for etcd connection...
No IP provided. Using detected IP: 172.18.0.1
  File "<string>", line 210, in warn_if_hostname_conflict
  File "site-packages/pycalico/datastore.py", line 137, in wrapped
pycalico.datastore_errors.DataStoreError: get_host_bgp_ips: Error accessing etcd (Connection to etcd failed due to MaxRetryError("HTTPConnectionPool(host='127.0.0.1', port=2379): Max retries exceeded with url: /v2/keys/calico/bgp/v1/host/durk/ip_addr_v4 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fdd67685110>: Failed to establish a new connection: [Errno 111] Connection refused',))",)).  Is etcd running?
startup returned -1
Calico node failed to start
```

### LINKED ISSUE #1315: Reinstate old STs to cover general calicoctl/config testing
Checklist to ensure we have at least the same ST coverage that we had before

- [ ] IPIP configuration
- [ ] Container to host pings
- [ ] Hostname override
- [ ] Manage pools
- [ ] Single host mainline
- [ ] Multi host mainline
- [ ] BGP configuration
- [ ] Global peering
- [ ] Node specific peering
- [ ] Route reflector cluster
- [ ] Single route reflector
- [ ] Node status test
- [ ] Node diags test
- [ ] IPAM release/show
- [ ] Endpoint show

- [ ] SSL tests

robbrockbank:
This PR re-instates the old calico-container STs.  Functionally these are basically testing the same stuff as before, but the tests were re-worked to use docker networking.

Fixes #1294 
Fixes #1257 
Fixes #1315

The original commit is the original ST files.  The second commit contains the rework plus some bug fixes highlighted by the tests, the final commit contains the updates to `calicoctl node run`.

Requires: https://github.com/projectcalico/libcalico/pull/170
Requires: https://github.com/projectcalico/libcalico-go/pull/266


calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+	if err != nil {
+		fmt.Println("Error querying calico/node logs:")
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	// Read stdout until the node fails, or until we see the output
+	// indicating success.
+	stdout := ""
+	txt := make([]byte, 1000)
```
gunjan5: might be better to use the `bufio` package here to create a read buffer since `outPipe` implements the `ReadCloser` interface `bufReader := bufio.NewReader(outPipe)` and `ReadLine()` that way you don't have to create a buffer of 1000 size which might not fit the need, and won't have to do `txt[:n]`

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+	var n int
+	for {
```
gunjan5: shouldn't we have a timout of some sort here? This loop could go on forever if we simply rely on docker process to fail or finish as docker daemon hangs many times and for those cases this would be in the loop forever

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+		if err != nil {
+			break
+		}
+		if n == 0 {
+			time.Sleep(10 * time.Millisecond)
+			continue
+		}
+
+		outStr := string(txt[:n])
+		stdout += outStr
```
gunjan5: why is this string concat needed? can't you check `if strings.Contains(outStr, "Calico node started successfully\n")` ?

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+		fmt.Print(outStr)
+
+		if strings.Contains(stdout, "Calico node started successfully\n") {
+			break
+		}
+	}
+
+	// Kill the process if it is still running.
+	logCmd.Process.Kill()
```
gunjan5: why are we ignoring the returned error here?

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+	logCmd.Wait()
```
gunjan5: same here. Also would be nice to stick a comment saying `// Waiting for the command to exit.`

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+
+	if err != nil {
```
gunjan5: which error is this for? I don't see any unhandled `err` variable before this

calicoctl/commands/node/run.go
```diff
@@ -256,6 +274,58 @@ Description:
...
+
+	// Create the command to follow the docker logs for the calico/node
+	logCmd := exec.Command("docker", "logs", "calico-node", "--follow")
+
+	// Get the stdout pipe
+	outPipe, err := logCmd.StdoutPipe()
+	if err != nil {
+		fmt.Println("Error querying calico/node logs:")
+		fmt.Println(err)
+		os.Exit(1)
```
robbrockbank: Well I don't expect it to fail, so an error querying the logs is an odd thing to occur and certainly something is not working correctly.

calicoctl/commands/node/run.go
robbrockbank: No not really - for the same reason that your status command didn't work.   A read only returns what is currently in the buffer - so you might get a partial read of a line.

Switching to use ReadLine() would sort this out.

calicoctl/commands/node/run.go
robbrockbank: The only reason this would fail is the process is already dead.   I'm not sure what use the error message would be here.  Perhaps we should just exit in this case rather than waiting.

calicoctl/commands/node/run.go
robbrockbank: Yes - if we can't read Stdout then there is no point in us getting the logs since we won't be able to see them.

robbrockbank: TBH, I suspect a non-nil error here is a panic worthy situation as I don't think we should ever hit it.... but better safe than sorry.

calicoctl/commands/node/run.go
robbrockbank: I don't think that is necessary here - by killing the process we know that it'll be dead pretty quickly.

robbrockbank:
@gunjan:  Regarding this branch:

```
	// Read stdout until the node fails, or until we see the output
	// indicating success.
	started := false
	for outScanner.Scan() {
		line := outScanner.Text()
		fmt.Println(line)
		if line == "Calico node started successfully" {
			started = true
			break
		}
	}

	// If we didn't successfully start then notify the user.
	if !started {
		fmt.Println("Error executing command: calico/node has terminated, check logs for details")
	}
```

I was explicitly not checking the outScanner.Error() because below, if we don't get the "started" message then we instruct the user to explicitly check the logs - so I don't think displaying the scanner error will provide any more useful context.

PR #1317, opened 2016-11-21
Restore SSL system tests

djosborne:
Restore the necessary SSL files removed by [this commit](https://github.com/projectcalico/calico-containers/commit/9ce8e39a8bb0e973589a582557246718795570fe) and add `st-ssl` to the semaphore make step.


robbrockbank:
LGTM

robbrockbank:
We may want to consider reducing the number of tests that are run as part of the SSL tests - but let's do that as a different PR - good to get this in ASAP.

PR #1318, opened 2016-11-21
do not detect IP addresses on dummy* devices

klynch:
I currently have some devices with the device name matching `dummy*` that I do not want automatically detected. I have a hard time believing someone else would want such a device detected either.


caseydavenport:
@klynch looks good to me - I'll merge right away once the cla ball goes green :)

klynch:
LGTM!

PR #1319, opened 2016-11-21
add flag to use predefined host ASN

klynch:
In some setups, it may be necessary to have peer using different ASNs (for
example when using unique ASN per rack). When the Calico node container starts,
the preconfigured ASN is overwritten and the node is no longer able to peer. By
providing a new environment variable USE_HOST_AS we are able to use the
preconfigured value and survive container restarts.

This is particularly important when using a DaemonSet to deploy the nodes to all
hosts.

### DIFF
--- a/calico_node/startup.py
+++ b/calico_node/startup.py
@@ -266,7 +266,9 @@ def main():
         sys.exit(1)
 
     as_num = os.getenv("AS")
-    as_num = as_num or None
+    if not as_num:
+        as_num = client.get_host_as(hostname)
+
     if as_num and not validate_asn(as_num):
         print "AS environment (%s) is not a AS number." % as_num
         sys.exit(1)


calico_node/startup.py
```diff
@@ -266,7 +266,11 @@ def main():
         sys.exit(1)
 
     as_num = os.getenv("AS")
-    as_num = as_num or None
+    if os.getenv("USE_HOST_AS") == "true":
```
caseydavenport: I think we could do this without introducing a new configuration option:

```python
as_num_env = os.getenv("AS")
if as_num_env:
  as_num = as_num_env
else:
  as_num = client.get_host_as(hostname)
```

The behavior is I think a little nicer:
- If "AS" is defined, use that value.  Otherwise,
- If the hostname has been configured with an AS, use that value.  Otherwise,
- No AS is used

caseydavenport: @klynch does that meet your requirement?

robbrockbank: @caseydavenport I don't think that works.  An empty AS Number means "inherit the global default value", so I think this would be the required approach.

caseydavenport: @robbrockbank doesn't this make more sense?

- Use the provided AS if one exists.  Otherwise,
- Use the host AS, if it is configured.  Otherwise,
- Use the global default value.

klynch: @caseydavenport Yes that should work for my use case. However, won't your suggestion slightly change existing behavior? Currently, if calico is restarted with an empty ASN, `client.create_host` will explicitly delete the entry in etcd. With your suggestion, if there is an ASN already defined and the client is restarted with `AS=` then the existing value will be used.

On a fresh configuration, either approach will work for me, however. What do you guys suggest?

caseydavenport: > Currently, if calico is restarted with an empty ASN, client.create_host will explicitly delete the entry in etcd

yikes, that sounds like a bug to me! Shouldn't we be respecting any config that has been set when we start calico/node rather than blatting over it?

I'll leave up to @robbrockbank to make the final call, but I like being able to do this all off of one `AS` env var, and it feels like we should be respecting whatever config exists for the node when it is started.

klynch:
Yup. Let me know if I should change it.

calico_node/startup.py
robbrockbank: Earlier releases didn't provide mechanisms for explicitly setting the ASN outside of passing the  environment when starting the node thus a blank ASN was used to remove the node-specific value and mean "use the global value".  I don't think we should break that.

However, I think I'd be happy using the same AS environment and have a "USE_NODE_AS" value to mean use the value set in the node configuration (e.g. by creating when modifying a node resource through calicoctl).

robbrockbank: One other detail.   Strictly speaking, if we are using the value configured on the node, then we should really do this by updating the client.create_host() method to not set the AS number in this case.   At the moment there is a potential timing window between reading the current value and then passing it in to create_host() where it will set it again.....

That said, I'd be happy to accept this update as it stands (with the update to use the single environment variable) and I'll raise an issue to close that timing window in a future release.

klynch: @robbrockbank do you mean switch to the single environment variable version @caseydavenport suggested?

robbrockbank:
@klynch   Hi - sorry to backtrack a little.   Been chatting this over with Casey and we both think that we can afford a minor change in the behavior (that is in fact unlikely to affect any real deployment scenario) and improves the overall use case.

Could you make it so that it follows @caseydavenport suggestion of using an empty or missing value to "use the current configured value".

robbrockbank:
@klynch :  Sorry that I dragged my heels on this one.  I was wondering though if you'd be able to get an update today as I'd quite like it included in the next cut that we will be doing.

klynch:
@robbrockbank sorry for the delay. It's merged in and tests are running now.

calico_node/startup.py
```diff
@@ -266,7 +266,12 @@ def main():
         sys.exit(1)
 
     as_num = os.getenv("AS")
-    as_num = as_num or None
+    as_num_env = os.getenv("AS")
```
robbrockbank: Looks like you don't need the extra `as_num_env` variable.   You could just do:

```
if not as_num:
    as_num = client.get_host_as(hostname)
```

robbrockbank:
LGTM

PR #1323, opened 2016-11-21
[WIP] Read from bird socket should handle timing issues

### LINKED ISSUE #1281: calicoctl node status sometimes doesn't show the BGP sessions
Hi,

I'm testing the v1.0.0-beta of calicoctl, and after setting up a BGP session with my router, the commando 'calicoctl node status' sometimes doesn't show the BGP session.

Output as can be seen:

```
[root@local-mesos-slave-001 ~]# calicoctl node status
Calico process is running.

IPv4 BGP status
+---------------+-----------+-------+----------+-------------+
| PEER ADDRESS  | PEER TYPE | STATE |  SINCE   |    INFO     |
+---------------+-----------+-------+----------+-------------+
| 192.168.56.15 | global    | up    | 12:58:40 | Established |
+---------------+-----------+-------+----------+-------------+

IPv6 BGP status
No IPv6 peers found.

[root@local-mesos-slave-001 ~]# calicoctl node status
Calico process is running.

IPv4 BGP status
No IPv4 peers found.

IPv6 BGP status
No IPv6 peers found.
```

My setup is as follows:
- 1 mesos-master
- 2 mesos-agents
- 1 router

I'm running this within vagrant / virtualbox on my laptop, based on RHEL7.

I disabled the node-to-node BGP mesh option, and configured the BGP session with the router with this yaml file:

```
- apiVersion: v1
  kind: bgpPeer
  metadata:
    scope: global
    peerIP: 192.168.56.15
  spec:
    asNumber: 64520
```

Attached is the output of `calico node diags`.
[diags-20161107_142526.tar.gz](https://github.com/projectcalico/calico-containers/files/575433/diags-20161107_142526.tar.gz)

gunjan5:
Fixes #1281

### DIFF
--- a/calicoctl/commands/node/status.go
+++ b/calicoctl/commands/node/status.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/docopt/docopt-go"
 	gops "github.com/mitchellh/go-ps"
 	"github.com/olekukonko/tablewriter"
@@ -135,7 +136,7 @@ func printBGPPeers(ipv string) {
 	timeOut := 2 * time.Second
 	bufReader := bufio.NewReader(c)
 
-	birdOut := ""
+	birdOut := []string{}
 
 	for {
 		// Set a time-out for reading from the socket connection.
@@ -146,17 +147,17 @@ func printBGPPeers(ipv string) {
 		str, err := bufReader.ReadString('\n')
 		if err != nil {
 			fmt.Println(err)
-			return
+			break
 		}
 
-		fmt.Println("[DEBUG] in str", str)
+		log.Debugf("Reading line from BIRD sock: %s\n", str)
 
 		// "0000" output from BIRD means end of output.
-		if strings.Contains(str, "0000") {
+		if str == "0000 \n" {
 			break
 		} else {
-			// Append the line to birdOut string if it's not the end of the output.
-			birdOut = birdOut + str
+			// Append the line to birdOut slice of strings if it's not the end of the output.
+			birdOut = append(birdOut, str)
 		}
 	}
 
@@ -165,7 +166,7 @@ func printBGPPeers(ipv string) {
 	table := tablewriter.NewWriter(os.Stdout)
 	table.SetHeader([]string{"Peer address", "Peer type", "State", "Since", "Info"})
 
-	for _, line := range strings.Split(birdOut, "\n") {
+	for _, line := range birdOut {
 
 		ipString := bgpPeerRegex.FindString(line)
 
--- a/calicoctl/commands/node/status.go
+++ b/calicoctl/commands/node/status.go
@@ -33,6 +33,7 @@ import (
 // Check for Word_<IP> where every octate is seperated by "_", regardless of IP protocols
 // Example match: "Mesh_192_168_56_101" or "Mesh_fd80_24e2_f998_72d7__2"
 var bgpPeerRegex, _ = regexp.Compile(`[A-Za-z]+\_\w+\b`)
+var birdCodeRegex, _ = regexp.Compile(`^\d00\d[a-zA-Z0-9\s-]+`)
 
 // Status prings status of the node and returns error (if any)
 func Status(args []string) {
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/status.go
```diff
@@ -129,20 +131,40 @@ func printBGPPeers(ipv string) {
...
+		c.SetReadDeadline(time.Now().Add(timeOut))
+
+		// Read string with \n as delim.
+		str, err := bufReader.ReadString('\n')
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+
+		fmt.Println("[DEBUG] in str", str)
```
robbrockbank: I think you should just put some real debug logs in here.  You can turn on debug logging when you run calicoctl commands.

calicoctl/commands/node/status.go
```diff
@@ -129,20 +131,40 @@ func printBGPPeers(ipv string) {
...
 
-	n, err := c.Read(buf[:])
-	if err != nil {
-		fmt.Printf("Error: %v\n", err)
+	birdOut := ""
+
+	for {
+		// Set a time-out for reading from the socket connection.
+		// Read operation will fail if no data is received until the timeout.
+		c.SetReadDeadline(time.Now().Add(timeOut))
```
robbrockbank: Do we need to keep setting the timeout?  Surely we can do this once outside the for loop.

calicoctl/commands/node/status.go
```diff
@@ -129,20 +131,40 @@ func printBGPPeers(ipv string) {
...
+
+		// Read string with \n as delim.
+		str, err := bufReader.ReadString('\n')
```
robbrockbank: Very inefficient looping here.  We are reading a string and then joining the strings together after they have been read.  And then below we split them by /n again and process each line separately.

Why not store a slice of strings and append each line to that slice.

calicoctl/commands/node/status.go
```diff
@@ -129,20 +131,40 @@ func printBGPPeers(ipv string) {
...
+		if err != nil {
+			fmt.Println(err)
+			return
+		}
+
+		fmt.Println("[DEBUG] in str", str)
+
+		// "0000" output from BIRD means end of output.
+		if strings.Contains(str, "0000") {
```
robbrockbank: I think this should be str == "0000\n"

calicoctl/commands/node/status.go
```diff
@@ -129,20 +131,40 @@ func printBGPPeers(ipv string) {
...
+			break
+		} else {
+			// Append the line to birdOut string if it's not the end of the output.
+			birdOut = birdOut + str
```
robbrockbank: Just append to a slice of strings and loop through the slice of strings below.

calicoctl/commands/node/status.go
```diff
@@ -129,21 +132,41 @@ func printBGPPeers(ipv string) {
...
+	// BIRD socket read timeout.
+	timeOut := 2 * time.Second
+	bufReader := bufio.NewReader(c)
 
-	n, err := c.Read(buf[:])
-	if err != nil {
-		fmt.Printf("Error: %v\n", err)
+	birdOut := []string{}
+
+	for {
```
robbrockbank: I think a Scanner would be nicer here.  Then this becomes:

```
scanner = bufio.NewScanner(c)
for scanner.Scan() {
    line = scanner.Text()
}
```

robbrockbank: you don't need to worry about errors within the loop

calicoctl/commands/node/status.go
```diff
@@ -129,21 +132,41 @@ func printBGPPeers(ipv string) {
...
+		str, err := bufReader.ReadString('\n')
+		if err != nil {
+			fmt.Println(err)
+			break
+		}
+
+		log.Debugf("Reading line from BIRD sock: %s\n", str)
+
+		// "0000 \n" output from BIRD means end of output.
+		if str == "0000 \n" {
```
robbrockbank: You aren't handling error codes are you?

I think this loop, i.e. the one that is reading the output from stdout should also be checking the msg codes in the output.  So this code should check for the heading line, the start of the data block and the end of the data block.  It should store the heading line separately and store the data lines as a slice, and remove the msg codes and space padding.  It is better done here that in the post processing.

robbrockbank: I think you should put above this scanning some sample output from BIRD.

robbrockbank: Better to have the following sort of logic:

-  Read line
-  Line starts with 4-digit code:
   -  Is it the heading code?  Yes, strip off and store the headings (or "assert" that they are what we expect)
   - Is it the end marker?  Yes, exit the loop
   - Is it the start of the data?  Yes?  "assert" we have read the heading.   Strip off code and store the data
  - It's an unexpected code.... Error
-  Line starts with a "space"
   -  Have we started collecting data?  Yes, this is more data, strip the line and store it.
   - We haven't started collecting data, this is unexpected.   Error.


So now we have stored all the data we expect, with no odd 4-digit codes etc.

calicoctl/commands/node/status.go
```diff
@@ -129,21 +132,41 @@ func printBGPPeers(ipv string) {
...
+		c.SetReadDeadline(time.Now().Add(timeOut))
+
+		// Read string with \n as delim.
+		str, err := bufReader.ReadString('\n')
+		if err != nil {
+			fmt.Println(err)
+			break
+		}
+
+		log.Debugf("Reading line from BIRD sock: %s\n", str)
```
robbrockbank: Shorted to "Read: %s\n"

And maybe put a debug log outside the for loop saying that you're reading from the BIRD socket, and put some debug trace in the branch where you exit the loop and some trace after you exit the loop.

calicoctl/commands/node/status.go
```diff
@@ -129,21 +132,41 @@ func printBGPPeers(ipv string) {
...
 
 	table := tablewriter.NewWriter(os.Stdout)
 	table.SetHeader([]string{"Peer address", "Peer type", "State", "Since", "Info"})
 
-	birdOut := string(buf[:n])
-
-	for _, line := range strings.Split(birdOut, "\n") {
+	for _, line := range birdOut {
 
 		ipString := bgpPeerRegex.FindString(line)
```
robbrockbank: What is the separator between columns in the output data?  Is it a tab, or just some whitespace?   If it is a tab then you shouldn't use Fields(), you should split on tab instead ... since some of the info parameters might contain spaces.

robbrockbank: If the columns are separated by tabs, are there always the same number of columns?  I'm wondering if the scanning should also split the columns and check the number of columns is as expected....erroring if not.

calicoctl/commands/node/status.go
```diff
@@ -129,21 +132,41 @@ func printBGPPeers(ipv string) {
...
+			fmt.Println(err)
+			break
+		}
+
+		log.Debugf("Reading line from BIRD sock: %s\n", str)
+
+		// "0000 \n" output from BIRD means end of output.
+		if str == "0000 \n" {
+			break
+		} else {
```
robbrockbank: Don't really need the else here, since you are breaking in the previous branch

calicoctl/commands/node/status.go
gunjan5: 1. Curious what value does scanner add vs using bufio bufferedReader? I've mostly seen scanner used for reading files and bufio reader for reading from socket.
2. I think we need to handle the error inside the loop, I looked at golang official example and they handle it inside for{} loop

calicoctl/commands/node/status.go
gunjan5: the issue is the separators are spaces, and to handle the case where some of the info parameters might contain spaces, I'm concatenating rest of the fields in the info tab, I've tested this when I originally wrote this code with different data outputs

gunjan5:
closing in favor of #1326

PR #1326, opened 2016-11-23
Rework of the node status command changes

robbrockbank:


robbrockbank:
I will fix the Scan() issue as part of #1316 (which is where the bug is).

Once tests pass I'll merge the PR.

PR #1327, opened 2016-11-23
[DO NOT MERGE] Integration of node status fix up and STs

robbrockbank:
Integrated the ST PR and the node status fix just to see if the two are playing nicely.  Looking good to me!


calicoctl/commands/node/run.go
```diff
@@ -249,10 +283,55 @@ Description:
...
+	// Protect against calico/node taking too long to start, or docker
+	// logs hanging without output.
+	time.AfterFunc(checkLogTimeout, func() {
+		logCmd.Process.Kill()
+	})
+
+	// Read stdout until the node fails, or until we see the output
+	// indicating success.
+	started := false
+	for outScanner.Scan() {
```
robbrockbank: Good point!  I'll fix.

PR #1329, opened 2016-11-23
curl should fail if smth is missing to download

SergK:
By default (HTTP) Fail silently (no output at all) on server errors.
This is mostly done to better enable scripts etc to better
deal with failed attempts.
So we must fail if some resources are unavailable


heschlie:
LGTM

PR #1333, opened 2016-11-23
Add improved error messages and use strict unmarshaling of data

robbrockbank:
Fixes https://github.com/projectcalico/libcalico-go/issues/230
Fixes https://github.com/projectcalico/libcalico-go/issues/248

This PR relies on forks of three other libraries, these forks provide improved error details, swap Float32 with Float64 support, strict validation of unexpected fields.   Please also see the following PRs:
 
https://github.com/projectcalico/go-yaml/pull/1
https://github.com/projectcalico/go-yaml-wrapper/pull/1

### DIFF
--- a/calicoctl/commands/get.go
+++ b/calicoctl/commands/get.go
@@ -164,8 +164,11 @@ Description:
 	results := executeConfigCommand(parsedArgs, actionList)
 	log.Infof("results: %+v", results)
 
-	if results.err != nil {
+	if results.fileInvalid {
 		fmt.Printf("Failed to execute command: %v\n", results.err)
+		os.Exit(1)
+	} else if results.err != nil {
+		fmt.Printf("Failed to get resources: %v\n", results.err)
 		os.Exit(1)
 	}
 
--- a/calicoctl/commands/get.go
+++ b/calicoctl/commands/get.go
@@ -165,7 +165,7 @@ Description:
 	log.Infof("results: %+v", results)
 
 	if results.err != nil {
-		fmt.Printf("Error getting resources: %v\n", results.err)
+		fmt.Printf("Failed to execute command: %v\n", results.err)
 		os.Exit(1)
 	}
 
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -29,7 +29,7 @@ import (
 	"bytes"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/ghodss/yaml"
+	"github.com/projectcalico/go-yaml-wrapper"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/validator"
 )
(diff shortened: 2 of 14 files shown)


calicoctl/commands/get.go
```diff
@@ -165,7 +165,7 @@ Description:
 	log.Infof("results: %+v", results)
 
 	if results.err != nil {
-		fmt.Printf("Error getting resources: %v\n", results.err)
+		fmt.Printf("Failed to execute command: %v\n", results.err)
```
gunjan5: does the new format still print "Error getting resources" ?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -171,7 +171,7 @@ func unmarshalResource(tm unversioned.TypeMetadata, b []byte) ([]unversioned.Res
 		return nil, err
 	}
 
-	if err = yaml.Unmarshal(b, unpacked); err != nil {
+	if err = yaml.UnmarshalStrict(b, unpacked); err != nil {
```
gunjan5: are we utilizing the special error(`UnmarshalUnknownFieldError`) this func provides to print a specific error? Just wondering where that happens since I don't see a special case that matches `UnmarshalUnknownFieldError` to print a specific message in this PR

robbrockbank: That is a good question, and at the moment I am kind of cheating.  
The UnmarshalUnknownFieldError displays a useful message that is geared towards our requirements, and I've updated the other JSON parsing/unparsing error messages to also be geared towards our requirements (i.e. to display text in a way that makes more sense to us).

That is not my long term approach though.  I intend to (but not for v1.0.0) return the error messages in the json/yaml packages back to their more json/yamlly format and in our code we'll switch on the different error types and craft our own error message out of them.   I guess that is where the wrapped error package will come in handy - and we should take the time to understand that and start doing our error handling properly.

But for v1.0.0 I wanted something that was quick and easy to test and to implement.

calicoctl/commands/get.go
robbrockbank: Oooh good spot - actually I think I meant to re-word this to be Failed to get resources.

robbrockbank: I've got two branches here now.  If the input file is invalid then it writes "Failed to execute command:  <msg about the file>"  whereas if it fails for another reason then it writes "Failed to get resources"

That'll be more consistent with the error messages for the create/apply/delete/update commands.

caseydavenport:
This LGTM.

glide.yaml
```diff
@@ -6,7 +6,6 @@ owners:
   email: rob@tigera.io
 import:
 - package: github.com/docopt/docopt-go
-- package: github.com/ghodss/yaml
```
caseydavenport: I think instead of doing this, glide lets you mask one as the other.

Not sure if that's preferred. Something like this:

```
- package: github.com/ghodss/yaml
   version: calico
   repo: github.com/projectcalico/go-yaml-wrapper
```

robbrockbank: I think given that there is an interface change and you can't switch between the two, I don't  think that would be a preferred approach.

PR #1337, opened 2016-11-24
Build our own golang-build container

SergK:
Use our own golang-build container for building go code into a
binary. The original idea came from felix project [1]

[1]. https://github.com/projectcalico/felix/blob/master/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,18 @@ CURL=curl -sSf
 # Default value: directory with Makefile
 SOURCE_DIR?=$(dir $(lastword $(MAKEFILE_LIST)))
 SOURCE_DIR:=$(abspath $(SOURCE_DIR))
+
+# Figure out the users UID/GID.  These are needed to run docker containers
+# as the current user and ensure that files built inside containers are
+# owned by the current user.
+MY_UID:=$(shell id -u)
+MY_GID:=$(shell id -g)
+
+# Pre-configured docker run command that runs as this user with the repo
+# checked out to /code, uses the --rm flag to avoid leaving the container
+# around afterwards.
+DOCKER_RUN_RM:=docker run --rm --user $(MY_UID):$(MY_GID)
+DOCKER_RUN_RM_ROOT:=docker run --rm
 ###############################################################################
 # URL for Calico binaries
 # confd binary
@@ -68,7 +80,7 @@ $(NODE_CONTAINER_CREATED): $(NODE_CONTAINER_DIR)/Dockerfile $(NODE_CONTAINER_FIL
 $(NODE_CONTAINER_BIN_DIR)/%: $(NODE_CONTAINER_DIR)/%.py
 	-docker run -v $(SOURCE_DIR):/code --rm \
 	 $(PYTHON_BUILD_CONTAINER_NAME) \
-	 sh -c 'pyinstaller -ayF --specpath /tmp/spec --workpath /tmp/build --distpath $(@D) $< && chown $(shell id -u):$(shell id -g) -R $(@D)'
+	 sh -c 'pyinstaller -ayF --specpath /tmp/spec --workpath /tmp/build --distpath $(@D) $< && chown $(MY_UID):$(MY_GID) -R $(@D)'
 
 # Get felix binaries
 $(NODE_CONTAINER_BIN_DIR)/calico-felix:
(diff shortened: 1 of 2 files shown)


SergK:
So --build-args in docker goes from 1.9.0 version, but on circleci it seems
```
Docker
We're using a build of Docker which is patched to work in the CircleCI build environment.
The current version is 1.8.2.
We also pre-install Docker Compose 1.5.1
```
the question is should I re-work my patch, or there is some change to update circleci test image

neiljerram:
@SergK Does this enhancement solve a specific problem?

I agree that it is sensible for Calico repositories to build a golang build container in similar ways; but on the other hand we don't want to enforce uniformity across our repos.  So I'd like to understand if you are addressing a specific problem here.

SergK:
For now one can have build issues for different calico components running on the same node, because of incorrect permissions in `$HOME/.glide` directory:
1. Try to build calico-containers, which for now build code under the root (https://github.com/projectcalico/calico-containers/blob/master/Makefile#L359)
2. Then try to build containers from felix repo and it fails, because in felix we run build under CI user
https://github.com/projectcalico/felix/blob/master/Makefile#L284

Of course, if CI user is root - no issue exists ... but on desktop, the issue is still exists, since in the most
cases users don't work under root
Definitely we can add simple `chown UID:GID -R /.glide` command in container

The rest issues are less important but I'd like to name them:
1. For example, we have a rather good feature https://github.com/projectcalico/calico-containers/pull/1342 and it can be implemented on other calico projects without any additional efforts, since we will have the same `go build way`
2. Moreover, probably we can move/create some component like calico/build-go-code container (put it for example in felix) and deliver to all calico components but simple pulling, so we will have the single point to configure calico/build-go-code
3. We don't need rely on upstream dockerepo/glide and can have our own fully customizable (with openssh out of the box, etc) container

matthewdupre:
@tomdee this sounds like your area of expertise?

robbrockbank:
I have a couple of comments, but I this more @tomdee and @fasacx area of expertise.

-  Fixing up the permissions issue with the $HOME/.glide is a good thing - it's certainly something I've run into in the past!

-  We generally name our containers calico/<name>  so calico/build-golang or something akin to that would be preferable.
-   This container would also be used for testing I believe, so we'd need to get rid of the test container that is also in this repo (See TEST_CALICOCTL_CONTAINER_NAME in the Makefile)
-  Is it necessary to specify the WORKDIR in the docker file since you are now passing that in as a Docker run argument?

fasaxc:
@robbrockbank FYI, we're using a convention of `calico-build/<name>` for build containers in felix to distinguish them from containers that we push.  That may or may not be a good distinction.

fasaxc:
In case anyone needs the background as to why I made the Felix build jump through some hoops to run its docker containers as non-root in the first place:

- It avoids leaking root-owned files and breaking permissions (i.e. breaking the glide cache apparently).  I didn't know about the glide cache explicitly, but I knew that if I ran the build as non-root I wouldn't run into that type of issue.
- It protects you against some mistakes (`rm -rf` is a tad more powerful as root)
- It reduces your attack surface (principle of least privilege); `go build` doesn't need root permissions so we shouldn't give it to it.  This is a bit academic, if someone's managed to sneak code into our build scripts then they could run as root via docker anyway.

tomdee:
The glide permission issue is definitely a problem that we should fix ASAP.  But I'm not sure (at least in the short term) whether than means we just stop mapping the `.glide` dir into builds in the short term, or fix permissions with a `chown` or move to a build container with correct users.

I'm cautiously in support of a single build container, but if it's going to be shared but a bunch of repos it's important that we get it right. 
- Where do we build it? In each repo or in one and then push it?
- Do we use it just for building? Or for vendoring too? Or for running tests too? What about when different repos need different test utilities or libraries, will it end up getting bloated?

I'll keep pondering the best way forward but I am keep to keep this moving forward.

tomdee:
So AFAICT, this PR is just focused on fixing the vendoring permissions issue whereas there seems to be an expectation that the "build" container will be used not just for vendoring but all building and testing.

SergK:
Ok folks, try to summarize from what we have.

The issue I'm trying to resolve:

1. Permission issues in `~/.glide` cache
2. Allow to provide pre-populated `/vendor` directory. In isolated environment (no access to 3rd party repos) `/vendor` directory can be pupulated
outside of calico build process, then we can just put it and update target `touch vendor/.up-to-date` manually,e.g.:

```
cp -r $CALICO_PROJECT/prepopulated_vendor $CALICO_PROJECT/vendor
touch $CALICO_PROJECT/vendor/.up-to-date
cd $CALICO_PROJECT
make build-containerized
```

Simple way:
1. Add simple chown
2. Update targets, to support `vendor/.up-to-date`


Long way:
1. Implement single build-go container for all calico projects, but this looks for me not so simple and obvious, since for example GID and UID can be different for CI user in different systems.
2. Implement build-go container per project.

So if Simple way is acceptable for now I can prepare small change that will have less impact on current `build-go` way, but will resolve the above issues

SergK:
I've created `less destructive` change https://github.com/projectcalico/calico-containers/pull/1387

SergK:
I think we can close this PR in favor of https://github.com/projectcalico/calico-containers/pull/1387

PR #1342, opened 2016-11-25
Build and use a glide container with SSH

neiljerram:


neiljerram:
@robbrockbank Please could you be the main reviewer here?
@matthewdupre I've also added you because I expect you'll be interested, as this allows having `repo:` fields in glide.yaml that point to Git repositories that need an SSH key.  The assumption in that case is that the developer (or CI environment) will have an SSH agent running, and will have added any required keys into that agent.

SergK:
Folks, can we have apply this patch to https://github.com/projectcalico/calico-containers/pull/1337
with this PR I'm trying to align the mainstream of golang-build container usage. The same approach we used in felix project as well the related patch was introduced in calico-cni (https://github.com/projectcalico/calico-cni/pull/222)
Actually https://github.com/projectcalico/calico-containers/pull/1337 failed because we are using ubuntu 12.04 image which doesn't support `--build-args`

Makefile
```diff
@@ -340,7 +340,8 @@ LDFLAGS=-ldflags "-X github.com/projectcalico/calico-containers/calicoctl/comman
 	-X github.com/projectcalico/calico-containers/calicoctl/commands.BUILD_DATE=$(CALICOCTL_BUILD_DATE) \
 	-X github.com/projectcalico/calico-containers/calicoctl/commands.GIT_REVISION=$(CALICOCTL_GIT_REVISION) -s -w"
 
-GLIDE_CONTAINER_NAME?=dockerepo/glide
+GLIDE_CONTAINER_NAME=calico-build/glide
+GLIDE_CONTAINER_MARKER=calico-build/glide_container.created
```
robbrockbank: At least in the non-Felix repos we usually just have a calico/ prefix for all of our container names .... so calico/glide would be preferable here.

robbrockbank: The clean target also needs updating to remove this container

Makefile
```diff
@@ -350,14 +351,18 @@ calico/ctl: $(CTL_CONTAINER_CREATED)      ## Create the calico/ctl image
 
 ## Use this to populate the vendor directory after checking out the repository.
 ## To update upstream dependencies, delete the glide.lock file first.
-vendor: glide.lock
+vendor: $(GLIDE_CONTAINER_MARKER) glide.lock
 	# To build without Docker just run "glide install -strip-vendor"
 	if [ "$(LIBCALICOGO_PATH)" != "none" ]; then \
           EXTRA_DOCKER_BIND="-v $(LIBCALICOGO_PATH):/go/src/github.com/projectcalico/libcalico-go:ro"; \
 	fi; \
+	if [ -n "$$SSH_AUTH_SOCK" ]; then \
```
robbrockbank: We need to document how to use this parm somewhere

robbrockbank:
If we are in the realms of creating our own vendoring container, then I'd be more inclined to have a single container for vendoring, compiling and running the go tests... to that end merging this into #1337 might be a good idea - probably worth Shaun taking a look over 1337 first.

Makefile
neiljerram: Thanks, I'll make those changes.

Makefile
neiljerram: Will do.

neiljerram:
@robbrockbank I've made the markups that you requested, and think this is good to merge now - WDYT?

(Re "If we are in the realms of creating our own vendoring container" and possible overlap with #1337, I'd say that _this_ PR is not in that realm; rather I'm just trying to solve a problem of how we can use a private repo when we want to.  #1337 certainly looks valuable also to me, but I think it can be progressed and merged independently of this, so I recommend that we do that and that it shouldn't block this PR.)

calico-build/Dockerfile.glide
```diff
@@ -0,0 +1,7 @@
+FROM dockerepo/glide
+MAINTAINER Neil Jerram <neil@tigera.io>
+
+RUN apk update
+RUN apk add openssh
```
tomdee: Better to use --no-cache than the update line above

calico-build/Dockerfile.glide
```diff
@@ -0,0 +1,7 @@
+FROM dockerepo/glide
```
tomdee: Add you adding a build job to dockerhub and quay.io to do autobuilds of this?

tomdee: One of the benefits of using our own image for this is so we can control which version of glide that we use - using this as the base image doesn't give us that benefit. Would it be easy to just inline this base image?

neiljerram: No, but should I? (Is it in some sense a regression from the status quo if I don't do that?)

neiljerram: Yes, I guess that would be straightforward, and would also encourage eventual convergence between the golang/glide containers that our different repos use. 

I'm not sure it should be mixed into this PR though.

calico-build/Dockerfile.glide
neiljerram: OK, will update that in about 30 mins.

calico-build/Dockerfile.glide
tomdee: You should at least be considering it. If you don't, will it be obvious to people how to update this image, e.g. to get an updated version of glide?

tomdee:
I'm still happy for this to be merged, but it might be better to just reimplement it on the centralized go-build container projectcalico/gobuild

neiljerram:
Abandoning this as we actually have no immediate need for it now.

Issue #873, opened 2016-03-15
Improve calico/node IP address detection


tomdee:
Our current method fails in too many cases.

It is however a really useful feature - e.g. for deploying as a kubernetes daemonset

wrouesnel:
A middleground option might be to allow specifying a DNS A/AAAA record which lists ALL public IP addresses a host could have. This would help resolve ambiguity, and in (at least my) deployment scenarios isn't too hard to setup in advance with shared images.

So `nodes.cloud.com` resolves to something like 192.168.1.10 and 192.168.1.11 - `calicoctl` retrieves all records, and checks all the interface IPs it has for one in the list.

olavmrk:
I also just hit this problem -- calico-node picked the wrong address on a node with multiple interfaces.

A simple way to pick the source address may be to just create a UDP socket and connect it to a public address. Since this is a UDP socket, no packets will be sent, but Linux will still bind the socket to the local IP it would use for sending packets:

```
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.connect( ('8.8.8.8', 0) )
default_src_ip = s.getsockname()[0]
```

matthewdupre:
It would be helpful if we could get the specific failure cases: perhaps `ip addr` output, and an indication of what was picked, and what should have been picked.

caseydavenport:
One case I've hit numerous times -

A vagrant setup with an eth0 which is NAT'd for internet access, and then an eth1 which is host-local and is meant for cluster traffic. All the eth0 interfaces have the same `10.0.2.15` address, and the eth1 interfaces have `172.18.18.X` addresses.

Calico always picks `10.0.2.15`.  For this, being able to specific which interface to pick an address from would probably be enough (this is also how flannel does it).

Without thinking __too_ much about it, this might work:
- An option to use a specific, exact IP address.
- An option limit chosen address to a specific interface (provided one exists on the node, error otherwise)
- An option limit chosen address from a specific CIDR (provided one exists on the node, error otherwise.)
- Default behavior of picking the first address on the first interface.

caseydavenport:
Sticking this at P2 since I think people will start to feel the pain here with Kubernetes self-hosted.

caseydavenport:
@tomdee @matthewdupre What do you think of the proposal in my comment above?

dkoshkin:
[...]

PR #1351, opened 2016-11-28
Allow calico/node to not override node resource parms if AS, IP or IP6 options are not specified

### LINKED ISSUE #1350: Update calicoctl node run to optionally overwrite the node resource parameters
Currently the node run command takes --as, --ip and --ip6 parameters to allow you to configure the AS number and IP addresses on the node.

However, these are also configurable through the node resource.

We should make it so that if the parameters are omitted then we will use the currently configured values.  If a node resource has not been explicitly created and set these values then we will:
-  Auto detect an IPv4 address
-  Use the global default AS number

robbrockbank:
Fixes #1350 

Requires #1319

### DIFF
--- a/calico_node/startup.py
+++ b/calico_node/startup.py
@@ -252,28 +252,28 @@ def main():
             print "Connected to etcd"
             break
 
-    # Start node.
-    ip = os.getenv("IP")
-    ip = ip or None
-    if ip and not netaddr.valid_ipv4(ip):
-        print "IP environment (%s) is not a valid IPv4 address." % ip
-        sys.exit(1)
+    # Query the currently configured host IPs
+    try:
+        current_ip, current_ip6 = client.get_host_bgp_ips(hostname)
+    except KeyError:
+        current_ip, current_ip6 = None, None
 
-    ip6 = os.getenv("IP6")
-    ip6 = ip6 or None
-    if ip6 and not netaddr.valid_ipv6(ip6):
-        print "IP6 environment (%s) is not a valid IPv6 address." % ip6
-        sys.exit(1)
+    # Determine the IP addresses and AS Number to use
+    ip = os.getenv("IP") or None
+    if ip == "autodetect":
+        # If explicitly requesting auto-detection, set the ip to None to force
+        # auto-detection.  We print below if we are auto-detecting the IP.
+        ip = None
+    elif ip:
+        if not netaddr.valid_ipv4(ip):
+            print "IP environment (%s) is not a valid IPv4 address." % ip
+            sys.exit(1)
+        print "Using IPv4 address from IP environment: %s" % ip
+    elif current_ip:
+        print "Using configured IPv4 address: %s" % current_ip
+        ip = current_ip
 
-    as_num = os.getenv("AS")
-    if not as_num:
-        as_num = client.get_host_as(hostname)
-
-    if as_num and not validate_asn(as_num):
-        print "AS environment (%s) is not a AS number." % as_num
-        sys.exit(1)
-
-    # Get IP address of host, if none was specified
+    # Get IP address of host, if none was specified or previously configured.
     if not ip:
         ips = get_host_ips(exclude=["^docker.*", "^cbr.*", "dummy.*",
                                     "virbr.*", "lxcbr.*", "veth.*",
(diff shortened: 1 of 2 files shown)


calico_node/startup.py
```diff
@@ -252,15 +252,19 @@ def main():
             print "Connected to etcd"
             break
 
+    # Query the currently configured host IPs
+    try:
+        ip, ip6 = client.get_host_bgp_ips(hostname)
```
caseydavenport: We should log out that we found an IP address in the datastore.

caseydavenport: I guess we've only got "print" here, so maybe that's too verbose...

calicoctl/commands/node/run.go
```diff
@@ -181,15 +194,23 @@ Description:
 
 	// Create a mapping of environment variables to values.
 	envs := map[string]string{
-		"HOSTNAME": name,
```
caseydavenport: The Felix runit job expects HOSTNAME to be set I think.  It sets the FELIX_HOSTNAME value based on that.

caseydavenport: We might want to keep both around for a little while, so that users can use the new calicoctl to launch old calico/node images?

robbrockbank: I don't believe we need to pass in HOSTNAME. 

The startup script honors both HOSTNAME and NODENAME and writes the name into a startup environments file that is used by rc.local  (and therefore by Felix and confd)

calico_node/startup.py
```diff
@@ -252,39 +252,58 @@ def main():
...
-    if ip6 and not netaddr.valid_ipv6(ip6):
-        print "IP6 environment (%s) is not a valid IPv6 address." % ip6
-        sys.exit(1)
+    ip6 = os.getenv("IP6") or None
+    if ip6:
+        if not netaddr.valid_ipv6(ip6):
+            print "IP6 environment (%s) is not a valid IPv6 address." % ip6
+            sys.exit(1)
+        print "Using IPv6 address from IP6 environment: %s" % ip
+    elif current_ip:
```
caseydavenport: should be current_ip6

calico_node/startup.py
```diff
@@ -280,11 +284,12 @@ def main():
...
             print "Couldn't autodetect a management IP address. Please " \
-                  "provide an IP address by rerunning the container with the" \
-                  " IP environment variable set."
+                  "provide an IP address either by configuring one in the " \
+                  "node resource, or by rerunning the container with the " \
+                  "IP environment variable set."
             sys.exit(1)
         else:
-            print "No IP provided. Using detected IP: %s" % ip
+            print "No IP configured or provided. Using detected IP: %s" % ip
```
robbrockbank: That is a valid fear - so I've added an "autodetect" option to the node run and to the container.

calico_node/startup.py
robbrockbank: ha!  already fixed :-)

caseydavenport:
Looking good to me.

Something about the auto-detect stuff makes me a little uneasy - it might be related to #873 though.

robbrockbank:
Yes the auto-detect stuff makes me nervous too.

We could add a auto-discovery method option that indicates which auto-discovery method to use, but that does sound like a separate PR to me.

caseydavenport:
yeah, definitely separate PR.  I just want to convince myself that an IP=auto option doesn't mean we're shooting ourselves in the foot somehow.

calicoctl/commands/node/run.go
```diff
@@ -63,17 +63,29 @@ func Run(args []string) {
...
+     --as=<AS_NUM>         Set the AS number for this node.  If omitted, the
+                           node will use the currently configured value and
+                           you can use the node resource to modify the AS
+                           number.  If there is no configured value and the
+                           the --as option is omitted, the node will inherit
+                           the global AS number (see 'calicoctl config' for
+                           details).
+     --ip=<IP>             Set the local IPv4 routing address for this node.
+                           If omitted, the node will use the currently
+                           configured value and you can use the node resource
```
caseydavenport: I'd reword this to "If omitted, will use the value configured on the node object."

caseydavenport:
LGTM

calicoctl/commands/node/run.go
```diff
@@ -65,27 +65,23 @@ Options:
...
-                           configured value and you can use the node resource
-                           to modify the address.  If there is no configured
-                           value and the --ip option is omitted, the node will
+                           If omitted, it will use the value configured on the
+                           node resource.  If there is no configured value
+                           and the --ip option is omitted, the node will
                            attempt to autodetect an IP address to use.  Use a
                            value of 'autodetect' to always force autodetection
-                           of the IP everytime the node starts.
+                           of the IP each the node starts.
```
gunjan5: did you mean to write `each time` here?

calico_node/startup.py
```diff
@@ -307,6 +289,31 @@ def main():
...
+    elif current_ip6:
+        print "Using configured IPv6 address: %s" % current_ip6
+        ip6 = current_ip6
+    else:
+        print "No IPv6 address configured"
+
+    as_num = os.getenv("AS")
+    if as_num:
+        if not validate_asn(as_num):
+            print "AS environment (%s) is not an AS number." % as_num
```
gunjan5: is not *a valid* AS number

PR #1354, opened 2016-11-29
Test global AS updates

robbrockbank:
Tests that changing the global AS number will trigger config regeneration


robbrockbank:
Not required.

PR #1359, opened 2016-11-29
ST the BIRD and gobgp backends

robbrockbank:


Makefile
```diff
@@ -95,16 +95,17 @@ $(NODE_CONTAINER_BIN_DIR)/confd:
 
 # Get the calico-bgp-daemon binary
 $(NODE_CONTAINER_BIN_DIR)/calico-bgp-daemon:
+	$(CURL) -L $(GOBGP_URL) -o $(@D)/gobgp
```
gunjan5: why are we adding this? Don't think we need this guy here since we're using gRPC to get the status as opposed to GoBGP CLI. Doesn't seem like it's being used for the tests either, this would just make the image havier

robbrockbank: A couple of reasons.
-  Consistency with BIRD, we include the CLI to aide debugging.
-  We don't currently have a working status command that is checked in do we .. and if we do, are we ST-ing it, because we should be.
-  It's useful if you want to poke around (like I needed to last night).

I guess it boils down to what % it adds to the image.  I wonder if we should produce a gobgp and a BIRD calico/node image so it has reduced size

robbrockbank: I would suggest leaving in now, and possibly removing if you get gobgp status working.   WDYT?

PR #1360, opened 2016-11-30
Print global ASN rather than blank

### LINKED ISSUE #1271: calicoctl doesn't show global default ASN.

robbrockbank:
Fixes #1271

### DIFF
--- a/calicoctl/commands/resources.go
+++ b/calicoctl/commands/resources.go
@@ -182,6 +182,10 @@ type commandResults struct {
 
 	// The results returned from each invocation
 	resources []unversioned.Resource
+
+	// The Calico API client used for the requests (useful if required
+	// again).
+	client *client.Client
 }
 
 // executeConfigCommand is main function called by all of the resource management commands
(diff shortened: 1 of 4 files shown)


calicoctl/commands/resources.go
```diff
@@ -182,6 +182,10 @@ type commandResults struct {
 
 	// The results returned from each invocation
 	resources []unversioned.Resource
+
+	// The Calico API client used for the requests (useful if required
+	// again).
```
gunjan5: not sure I understand this: "useful if required again"

gunjan5:
LGTM assuming you have tested the output since templating can be weird sometimes, and I can't tell by just looking at the code. :)

robbrockbank:
Yup I tested.   The default value appears in parens, whereas the configured value appears normally.

calicoctl/commands/resources.go
robbrockbank: Just that there is no real reason to return the client in the response data ... but in this case it is useful to have access to it so that we don't have to create another client.

PR #1361, opened 2016-11-30
Add a warning when downloading the docker image

### LINKED ISSUE #1287: `calicoctl node run` hangs without warning if it needs to download image
The new `calicoctl node run` command hangs for a long time while the image is being downloaded, without telling the user why.

We should just  always print an info message that states `Pulling calico/node:xyz if it doesn't exist locally...`

robbrockbank:
Fixes #1287

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -298,10 +298,21 @@ Description:
 		setNFConntrackMax()
 	}
 
-	// Run the docker command.
-	fmt.Println("Running the following command:")
-	fmt.Printf("\n%s\n\n", strings.Join(cmd, " "))
+	// Make sure the calico-node is not already running before we attempt
+	// to start the node.
+	fmt.Println("Removing old calico-node container (if running).")
+	err = exec.Command("docker", "rm", "-f", "calico-node").Run()
+	if err != nil {
+		log.WithError(err).Debug("Unable to remove calico-node container (ok if container was not running)")
+	}
 
+	// Run the docker command.
+	fmt.Println("Running the following command to start calico-node:")
+	fmt.Printf("\n%s\n\n", strings.Join(cmd, " "))
+	fmt.Println("Image may take a short time to download if it is not available locally.")
+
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/run.go
```diff
@@ -298,17 +298,30 @@ Description:
 		setNFConntrackMax()
 	}
 
+	// Make sure the calico-node is not already running before we attempt
+	// to start the node.
+	fmt.Println("Removing old calico-node container (if running).")
+	err = exec.Command("docker", "rm", "-f", "calico-node").Run()
+	if err != nil {
+		log.WithError(err).Debug("Unable to remove calico-node container")
```
gunjan5: we should add `ignore this if calico-node is not running` so users don't get confused when they see that message in the logs

calicoctl/commands/node/run.go
```diff
@@ -298,17 +298,30 @@ Description:
...
+	}
+
 	// Run the docker command.
-	fmt.Println("Running the following command:")
+	fmt.Println("Running the following command to start calico-node:")
 	fmt.Printf("\n%s\n\n", strings.Join(cmd, " "))
+	fmt.Println("If the container image is not available locally, this may take a short amount")
+	fmt.Println("of time to complete.\n")
```
gunjan5: hmmm.. not sure how i feel about having 2 separate lines for a single message, what if my terminal size is 60 chars and it wraps, first line would be 2 lines (1 and a half-ish) and then `of time to complete.` in a new line leaving the second half of the line empty. It might look like this:
```
If the container image is not available locally, this may
take a short amount
of time to complete.
```
might be better to just have it in one Println, WDYT?

calicoctl/commands/node/run.go
robbrockbank: users generally don't look in the logs.

calicoctl/commands/node/run.go
robbrockbank: kk

PR #1376, opened 2016-12-01
Add envs to containerized binary build

### LINKED ISSUE #1375: RC1 calicoctl runs the wrong calico/node
The RC1 calicoctl runs the following when doing `calicoctl node run`:  
```$ sudo ./calicoctl node run
Running command to load modules: modprobe -a xt_set ip6_tables
Enabling IPv4 forwarding
Enabling IPv6 forwarding
Increasing conntrack limit
Removing old calico-node container (if running).
Running the following command to start calico-node:

docker run --net=host --privileged --name=calico-node -d --restart=always -e CALICO_NETWORKING_BACKEND=bird -e NO_DEFAULT_POOLS= -e CALICO_LIBNETWORK_ENABLED=true -e ETCD_AUTHORITY=127.0.0.1:2379 -e ETCD_SCHEME=http -e ETCD_ENDPOINTS= -e NODENAME=drogon -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules -v /var/log/calico:/var/log/calico -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock calico/node:latest
```
So its running `calico/node:latest`.  Is that correct?  Should it not be a specific calico/node version?

robbrockbank:
Fixes #1375 

Build process uses containerized builds which did not pass in the "overridable" environments.  These are now all being passed through.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,14 +263,19 @@ semaphore: clean
 	bash -c 'rm -rf /home/runner/{.npm,.phpbrew,.phpunit,.kerl,.kiex,.lein,.nvm,.npm,.phpbrew,.rbenv}'
 
 	# Actually run the tests (refreshing the images as required), we only run a
-	# small subset of the tests for testing SSL support.
-	CALICOCTL_NODE_VERSION=$$BRANCH_NAME $(MAKE) calico/ctl calico/node st
-	CALICOCTL_NODE_VERSION=$$BRANCH_NAME ST_TO_RUN=tests/st/policy $(MAKE) st-ssl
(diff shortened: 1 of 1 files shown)


robbrockbank:
The tagging of the release in the .tar make target seems a little out-of-place.

I'm wondering if we should tag the versioned and quay images as part of the make calico/ctl and calico/node targets.   

@tomdee wdyt?

tomdee:
As discussed, you're going to make some tweaks to the semaphore target in the Makefile

tomdee:
LGTM

PR #1381, opened 2016-12-04
Start moving to startup.go

caseydavenport:
Initial intent is to create a new startup.go and migrate pieces of startup.py over to it.

First bit to migrate is the ClusterGUID.

### DIFF
--- a/calico_node/startup.go
+++ b/calico_node/startup.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"github.com/satori/go.uuid"
+
+	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


calico_node/startup.go
```diff
@@ -0,0 +1,73 @@
...
+		panic(fmt.Sprintf("Error initializing datastore: %s", err))
+	}
+	log.Info("Datastore is initialized")
+
+	// Make sure we have a global cluster ID set.
+	ensureClusterGuid(c)
+}
+
+// ensureClusterGuid assigns a cluster GUID if one doesn't exist.
+func ensureClusterGuid(c api.Client) {
```
caseydavenport: :(

For some reason Felix isn't picking up the value that's being assigned here.

robbrockbank:
Need to fix up merge conflicts.

I'm okay with this as is for the next RC, but we should move the GUID stuff into libcalico-go - it is already in libcalico-go within the node creation branch.

In the longer term I was always imagining that the startup would effectively create or update a node resource and that the node resource would do much of the etcd init stuff  -- that way it's useful for environments that don't use calico/node..... 

But for now... LGTM

calico_node/startup.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// main performs startup operations for a node.
+// For now, this only creates the ClusterGUID.  Ultimately,
+// all of the function from startup.py will be moved here.
+func main() {
+	// Build a Calico client.
+	log.Info("Creating Calico client")
+	cfg, err := client.LoadClientConfig("")
+	if err != nil {
```
robbrockbank: Should some of  these logs be print statements so that we are outputting to screen (like we do with the python startup?)

caseydavenport:
> In the longer term I was always imagining that the startup would effectively create or update a node resource and that the node resource would do much of the etcd init stuff -- that way it's useful for environments that don't use calico/node.....

yes, this is exactly what I'd like as well. Starting calico/node should just be a shortcut for running `calicoctl create -f mynode.yaml`

I couldn't get all the way there in this PR because I didn't want to mess too much with the IP detection, etc in the v2.0 timeframe.  Will be doing that after v2.0 though!

PR #1387, opened 2016-12-06
Use glide cache to build go code

SergK:
* use cache as much as possible from CI user (~/.glide) also with
  fixing permissions
* add $MY_UID $MY_GID variables

We need to have an ability to run calico build in isolated environment
(without access to 3rd part repos) and cache usage is one step to this
direction


SergK:
Hi, @tomdee @fasaxc could you please review this PR?

tomdee:
Hi @SergK this was fixed as part of #1461 so closing.

PR #1396, opened 2016-12-07
Fix calicoctl cross-platform builds

caseydavenport:


calicoctl/commands/node_darwin.go
```diff
@@ -0,0 +1,26 @@
...
+package commands
+
+import (
+	"fmt"
+	"os"
+)
+
+// Node function is a switch to node related sub-commands
+func Node(args []string) {
+	fmt.Println("node sub-command is not available on darwin")
```
robbrockbank: Could we make this:   Error executing command: calicoctl node commands are not available on this OS.

robbrockbank:
One comment then LGTM

PR #1402, opened 2016-12-09
Rev felix to 2.0.0-rc5.

fasaxc:


fasaxc:
@tomdee This re-integrates the commit that I built RC3 from.  Note: I thought calico-containers wasn't having checked-in pins but I saw rc4 was in there so thought I'd better be consistent.

PR #1406, opened 2016-12-10
Better resource error message

### LINKED ISSUE #1405: calicoctl apply returns "No resources specified in file"
**Version**: `v1.0.0-rc1`, `v1.0.0-rc2`, `v1.0.0-rc3`
**OS**: CentOS

```
[root@ip-10-0-3-122 calico]# ./calicoctl apply -f ipPool.yaml
No resources specified in file
[root@ip-10-0-3-122 calico]# cat ipPool.yaml
apiVersion: v1
kind: ipPool
metadata:
  cidr: "172.16.0.0/16"
spec:
  ipip:
    enabled: true
  nat-outgoing: true
  disabled: false
[root@ip-10-0-3-122 calico]# ./calicoctl apply -f ipPool.yaml
No resources specified in file
[root@ip-10-0-3-122 calico]# ./calicoctl create -f ipPool.yaml
No resources specified in file
```
I've also tried the example from the [docs](http://docs.projectcalico.org/v2.0/reference/calicoctl/resources/ippool) with the same error.

Same file works with calicoctl `v1.0.0-beta`

gunjan5:
Fixes #1405 
Pre-req: https://github.com/projectcalico/libcalico-go/pull/302

### DIFF
--- a/calicoctl/commands/resources.go
+++ b/calicoctl/commands/resources.go
@@ -17,14 +17,15 @@ package commands
 import (
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/projectcalico/go-yaml-wrapper"
 	"github.com/projectcalico/calico-containers/calicoctl/commands/argutils"
 	"github.com/projectcalico/calico-containers/calicoctl/commands/clientmgr"
 	"github.com/projectcalico/calico-containers/calicoctl/resourcemgr"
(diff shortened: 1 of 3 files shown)


gunjan5:
@caseydavenport wondering if I should pin libcalico-go since the message is only useful if calicoctl uses the code that was merged with  projectcalico/libcalico-go#302 (calicoctl is currently pinned to v1.0.0-rc4 which doesn't include the libcalico-go fix)

caseydavenport:
@gunjan5 yeah, updating the pin probably makes sense.

PR #1408, opened 2016-12-12
Makefile: Use latest etcd (in three places) for tests

tomdee:
Use a single version, and the latest so we can spot problems with etcd
updates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,8 @@ run-etcd-st:
 	$(MAKE) stop-etcd
 	docker run --detach \
 	--net=host \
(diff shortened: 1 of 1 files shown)


djosborne:
In an attempt to make it easier to diagnose if a silent etcd version bump is causing build failures in the future, I'm wondering if we should:

1. Print the version of etcd that is used
2. Make it overridable via environment variable

But it's not obvious to me right now if either of those will help. So I'll give it an LGTM for now, and we'll see what improvements are needed later on when it starts catching / causing issues :)

PR #1410, opened 2016-12-12
windows build with .exe

### LINKED ISSUE #1397: windows version of calicoctl needs to end in .exe
Otherwise windows wants to open it in another application...

gunjan5:
Fixes #1397

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ default: help
 all: test                                 ## Run all the tests
 test: st test-containerized               ## Run all the tests
 ssl-certs: certs/.certificates.created    ## Generate self-signed SSL certificates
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -414,6 +414,7 @@ dist/calicoctl-darwin-amd64: $(CALICOCTL_FILES) vendor
 
 dist/calicoctl-windows-amd64: $(CALICOCTL_FILES) vendor
```
caseydavenport: You should update the make target to match so that running `make dist/calicoctl-windows-amd64` twice doesn't rebuild the same binary.

caseydavenport:
LGTM

PR #1411, opened 2016-12-12
Convert startup.py to Go

2blesd2bstresd:
Added function to detect and filter interfaces based on interface name, and return valid IPV4 addresses.


calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+
+	for _, iface := range ifaces {
+
+		valid_ip := true
+
+		// Check each interface name against 
+		// list of interfaces to ignore.
+		for _, v := range ignoreInterfaces {
+			regex, err := regexp.Compile(v)
+
```
caseydavenport: I think generally it's good to keep the error checking close to the thing that's generating the error, so I'd remove this newline in between.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+			if err != nil {
+				continue
+			}
+
+			if regex.MatchString(iface.Name) {
+				valid_ip = false
```
caseydavenport: Given that we're checking if the interface is valid, I think a better name for this flag would be "valid_interface" or "valid_iface".

However, I think we can get rid of the flag completely with a little bit of restructuring.

e.g

```
if !regex.MatchString(iface.Name) {
  // This is a valid interface.  Get all the IPs from it.
  ips = append(ips, ipsForInterface(iface))
}
```

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"net"
+	"regexp"
+	"strings"
+)
+
+func detectIPs(ignoreInterfaces []string) ([]string, error) {
```
caseydavenport: where does this code get called from? :)

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+
+			if regex.MatchString(iface.Name) {
+				valid_ip = false
+			}
+		}
+
+		if valid_ip == true {
+			addrs, err := iface.Addrs()
+
```
caseydavenport: Same here - I'd remove the extra newline so it's clear which error we're handling.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+			if err != nil {
+				continue
+			}
+			
+			for _, addr := range addrs {
+
+				ip := strings.Split(addr.String(), "/")[0]
+				valid_ip := net.ParseIP(ip)
+
+				if valid_ip.To4() != nil{
```
caseydavenport: According to this, https://golang.org/pkg/net/#ParseIP, ParseIP can return nil, in which case this will be a nil pointer exception.  We probably need to check if valid_ip is nil first.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+		if valid_ip == true {
+			addrs, err := iface.Addrs()
+
+			if err != nil {
+				continue
+			}
+			
+			for _, addr := range addrs {
+
+				ip := strings.Split(addr.String(), "/")[0]
```
caseydavenport: I think a better way to do this would be to parse the whole String() output as a CIDR and then use the `.IP` field of the returned object to get the IP address. That way we're not messing around with string splits / array indexes.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
+package main
```
caseydavenport: Can we make a new package `autodetection` for this code?  Maybe in `calico_node/autodetection`?

Keeping it separate will be better for adding alternative auto detection mechanisms down the road.

calico_node/ip_detection.go
caseydavenport: I think a better signature for this would be:

```
// detectIPs returns a list of IPv4 address and a list of IPv6 addresses assigned across
// all interfaces on the host which don't match any string in ignoreInterfaces, or an error.
func detectIPs(ignoreInterfaces []string) ([]net.IPNet, []net.IPNet error)
```

It's more flexible to pass around IPNet objects as opposed to strings, and it's useful to pass back both IPv4 and IPv6 addresses.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"net"
+	"regexp"
+	"strings"
+)
+
+func detectIPs(ignoreInterfaces []string) ([]string, error) {
+
+	ips := []string{}
+	ifaces, err := net.Interfaces()
```
caseydavenport: We're not handling the error returned here but we should be :)

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+
+		valid_ip := true
+
+		// Check each interface name against 
+		// list of interfaces to ignore.
+		for _, v := range ignoreInterfaces {
+			regex, err := regexp.Compile(v)
+
+			if err != nil {
+				continue
```
caseydavenport: There's no indication of an error when we hit this - we should at least be logging something urgent here, and at most be panicking.

calico_node/ip_detection.go
```diff
@@ -0,0 +1,53 @@
...
+			for _, addr := range addrs {
+
+				ip := strings.Split(addr.String(), "/")[0]
+				valid_ip := net.ParseIP(ip)
+
+				if valid_ip.To4() != nil{
+					ips = append(ips, ip)
+				}
+					
+				}
```
caseydavenport: I think you need to run `gofmt` over this code.  If you haven't already, you can set up `gofmt` to run on save for most IDEs / vim.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+			}
+  			ipv6_meta = api.IPPoolMetadata{CIDR: cnet.IPNet{*cidr}}			
+  			c.Apply(api.IPPool(ipv6_meta))
+		}
+	}
+
+	ns := api.NodeSpec{api.NodeBGPSpec{ASNumber: as_num, IPv4Address: ip, IPv6Address, ip6}}
+	nm := api.NodeMetadata{Name: hostname}
+	n := api.Node{Metadata: nm, Spec: ns}
+	n, err : = c.Nodes().Create(n)
```
caseydavenport: Create() is the wrong operation here.  The Node object might already exist, in which case Create() will throw an error.

I think we want to use Apply() instead, since that will update it if it exists, or create it if it doesn't.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+	if err != nil{
+		panic(info.Warnf("Failed to create the node: %s", err))
```
caseydavenport: We should replace all the `panic(...)` statements in this code with `log.Fatalf(...)`

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+	}
+
```
caseydavenport: Weird spacing here.  Make sure you're running `gofmt` over this.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+		ipv4_present, ipv6_present := false, false
+		for p := range poolList {
+			s := strings.Split(p.IPPoolMetadata.CIDR.String())
+			valid_ip := net.ParseIP(ip)
+			if valid_ip.to4() != nil {
+				ipv4_present = true
+			} else {
+				ipv6_present = true
+			}
+			if ipv4_present == true && ipv6_present_present == true {
```
caseydavenport: present_present

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+                   node resource, or by re-running the container with the
+                   IP environment variable set.")
+		os.exit(1)
+	} 
+
+    // Write a startup environment file containing the IP address that may have
+    // just been detected.
+    // This is required because the confd templates expect to be able to fill in
+    // some templates by fetching them from the environment.
+	f, err := os.OpenFile("startup.env", os.O_APPEND|os.O_WRONLY, 0600)
```
caseydavenport: Weird indentation here.  `gofmt` should fix it up.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+	as_num := os.Getenv("AS")
+	if as_num != nil {
+		as, err := numorstring.ASNumberFromString(as_num)
+		if err != nil {
+			as_num = nil
+		}
+	}
+
+	// If ip, ip6 and as_num haven't been set yet, get them from bgp
+	ns, err := c.Nodes().Get(meta)
```
caseydavenport: I think `ns` should be called `node` instead to make this a bit more readable.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+				break
+			}
+		}
+
+		if ipv4_present == false {
+			_, cidr, err := net.ParseCIDR("192.168.0.0/16")
+			if err != nil {
+				panic(info.Warnf("Unable to create ipv4 cidr: %s", err))
+			}
+  			ipv4_meta = api.IPPoolMetadata{CIDR: cnet.IPNet{*cidr}}
```
caseydavenport: You're missing the Spec section of the IPPool here.  https://godoc.org/github.com/projectcalico/libcalico-go/lib/api#IPPool

We don't do this in the old code, but I think we should be setting the Spec.NATOutgoing field to be true.

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+		if err != nil {
+			panic(fmt.Sprintf("Error loading config: %s", err))
+		}
+		f.WriteString("export DATASTORE_TYPE=kubernetes\n")
+		f.WriteString("export HOSTNAME=%s\n" % hostname)
+	}
+
+	meta := api.NodeMetaData{Name: hostname}
+
+	fmt.Printf("Waiting for etcd connection...")
```
caseydavenport: We also need to move this WAIT_FOR_DATASTORE loop up above any code that actually uses the datastore.

There are a few bits above that depend on a datastore connection. e.g `ensureClusterGuid`

Would also make the code a bit cleaner to encapsulate this into a function call. e.g `waitForDatastore()`

calico_node/startup.go
```diff
@@ -51,8 +59,239 @@ func main() {
...
+	for v := os.Getenv("WAIT_FOR_DATASTORE"); v != "false"; {
+		_, err := c.Nodes().Get(meta)
```
caseydavenport: As far as I can tell, `c` is an instance of backend.Client, https://godoc.org/github.com/projectcalico/libcalico-go/lib/backend/api#Client, which means it doesn't actually have a Nodes() method.  We need to instantiate one of these instead: https://godoc.org/github.com/projectcalico/libcalico-go/lib/client#Client

caseydavenport:
> Tigera committed 6 days ago

^ Not sure what's going on but your commits seem a bit weird

PR #1420, opened 2016-12-14
[WIP] Perform IPIP only off host subnet

robbrockbank:
This PR adds the ability to selectively use IPIP based on the hosts subnet.
-  If the IP Pool has IPIP enabled, IPIP is only used if the originating router is off subnet
-  IP subnet detection is added to the node startup processing
-  The ps wide format output for the nodes is updated to include the subnets.
-  Remove the use of HOSTNAME within the calico/node since it was deprecated in favor of NODENAME
-  Update confd templates to use the node IP addresses configured in the node resource rather than the ones in the environment variables - this allows a user to modify the node resource IPs and have the changes reflected in the node.

### DIFF
--- a/calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
+++ b/calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
@@ -2,7 +2,7 @@
 filter calico_pools {
   calico_aggr();
   custom_filters();
-{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
   if ( net ~ {{$data.cidr}} ) then {
     accept;
   }
@@ -10,12 +10,23 @@ filter calico_pools {
   reject;
 }
 
+{{$network_key := printf "/bgp/v1/host/%s/network_v4" (getenv "NODENAME")}}{{$network := getv $network_key}}
 filter calico_ipip {
-{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}{{if $data.ipip}}
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
   if ( net ~ {{$data.cidr}} ) then {
+{{if $network}}
+    if ( from ~ {{$network}} ) then
+      krt_tunnel = "";
+    else
+      krt_tunnel = "{{$data.ipip}}";
+
+    accept;
+  }
+{{else}}
     krt_tunnel = "{{$data.ipip}}";
     accept;
   }
 {{end}}{{end}}
   accept;
 }
+
--- a/calico_node/startup.go
+++ b/calico_node/startup.go
@@ -16,15 +16,19 @@ package main
 import (
 	"encoding/hex"
 	"fmt"
-
-	"github.com/satori/go.uuid"
+	"io/ioutil"
+	"net/http"
+	"os"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/backend"
-	"github.com/projectcalico/libcalico-go/lib/backend/api"
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/satori/go.uuid"
 )
 
 // main performs startup operations for a node.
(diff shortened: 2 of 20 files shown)


calico_node/startup.go
```diff
@@ -51,22 +53,71 @@ func main() {
...
+	log.Infof("Current node config: %v", node)
+
+	// Check to see if we are on AWS, and if so get the IPv4 and IPv6
+	// subnets (if they have not yet been configured).
+	if node.Spec.BGP == nil || node.Spec.BGP.IPv4Network == nil {
+		log.Infof("Attempt to query AWS IPv4 network")
+
+
+		// query /mac to get MAC of eth0
+		// query /network/interfaces/macs/<MAC>/vpc-ipv4-cidr-block to get the CIDR block of the VPC
```
seanturner83: Hi Rob - my notes say "curl http://169.254.169.254/latest/meta-data/network/interfaces/macs/DEADBEEF/subnet-ipv4-cidr-block"

calico_node/startup.go
```diff
@@ -51,22 +53,71 @@ func main() {
...
+		} else {
+			log.WithField("Name", nm.Name).Info("Node has not yet been configured")
+			node = &api.Node{Metadata: nm}
+			modified = true
+		}
+	}
+	log.Infof("Current node config: %v", node)
+
+	// Check to see if we are on AWS, and if so get the IPv4 and IPv6
+	// subnets (if they have not yet been configured).
```
caseydavenport: Is this not available on the instance interface we're binding to?  i.e can we do something to remove the AWS specific bits of code?

calico_node/startup.go
```diff
@@ -51,22 +53,71 @@ func main() {
...
 			panic(err)
 		}
 		log.Infof("Using previously configured ClusterGUID")
 		return
 	}
 	log.Infof("Assigned ClusterGUID %s", guid)
 }
+
+// Update the current Node resource configuration (if it exists).
+func updateNodeResource() {
```
caseydavenport: This is going to overlap a lot with work that @2blesd2bstresd is working on.

Can you guys make sure that you're in sync?

calico_node/startup.go
seanturner83: I think so - it'll be the CIDR block for eth0 right? I don't know how to get this though.

calico_node/startup.go
caseydavenport: That said, I'd like to see the flow look more like this:

```
// Get ipip subnet if necessary.
ipipNet := determineIPIPSubnet()

// Do some other stuff ... get IP address, AS, etc etc
...

// Update the node resource in one go with all the stuff we've discovered above.
c.Nodes().Apply(node)
```

calico_node/startup.go
seanturner83: kind of begs the question why you'd choose to do this if you weren't on something as broken as AWS networking..

caseydavenport: You might still want to do this on bare metal, for example, if your TOR doesn't support BGP or if your network guys don't want to open up BGP.

Basically any time you've got a cluster that spans across multiple L3 subnets and you can't peer with the intermediate router, this would be super useful.

seanturner83: Variation on ip -4 route | awk '/^default/ { print $3 }'

calico_node/startup.go
robbrockbank: Hi @seanturner83 - yes the documentation for AWS metadata does indeed say that -- replacing the DEADBEEF with the appropriate MAC address which can be looked up with another metadata invocation.

calico_node/startup.go
robbrockbank: @caseydavenport:  I realize that this will overlap a lot with that work -- so I am expecting some overlap.  Be aware this work is WIP and a little premature for review because these is little prototyping and also a little more investigation required.   At the moment I am not sure if interface querying is sufficient or whether we should be querying the AWS metadata service.  I needed internet access to be able to do that which I didn't have at the time.

calico_node/startup.go
robbrockbank: @caseydavenport  I have not done interface based querying yet simply because it overlaps too much with the PR that @2blesd2bstresd is working on.   I don't think it makes sense to do that until that PR is in - the general flow will then be sorted out.

That said - the purpose of this current PR is to check our assumptions about what can be used for the IPIP decision.  See my above note about the metadata subnet vs vpc CIDRs

robbrockbank: Regarding bare metal - yes my plan is not to make this AWS specific ... the work I've done is AWS specific because I think it *might* still be useful and I don't want to tread on the PR that is currently being worked on for interface querying.

calico_node/startup.go
caseydavenport: Yep, sorry to prematurely review!

I was mostly wanting to make sure you guys are in sync on the two potentially conflicting PRs.

calico_node/startup.go
lxpollitt: +1 to removing any AWS specific bits from this - assuming the subnet is correctly there on the host interface.

calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -2,20 +2,34 @@
...
 filter calico_ipip {
-{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}{{if $data.ipip}}
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}{{if $data.ipip}}
   if ( net ~ {{$data.cidr}} ) then {
+{{if $network}}
+    krt_tunnel = "{{$data.ipip}}";
+    accept;
+  }
+{{else if $data.ipip_off_subnet}}
+    if ( gw ~ {{$network}} ) then {
```
seanturner83: Hi Rob - I couldn't get this filter to work - I ended up using "gw != <gateway address of subnet>" instead - BIRD complained about a type conversion? (e.g. from an address to a CIDR) What does the "network_v4" key contain in this version?

robbrockbank:
Replaced by #1453

PR #1427, opened 2016-12-22
Move start up processing from startup.py to startup.go

robbrockbank:
Porting of startup.py to go.

At the moment the detection of ip and network is based around having separate IP and IP network fields (although the latter doesn't actually exist).  When we rework the node resource interface this code will change a little - but I'd rather just get this in and tweak it alongside that work.

Careful reviewing as the startup.go file doesn't seem to auto diff - so it's easy to miss.


robbrockbank:
I need to get the libcalico-go changes in and then update  though and it seems to work ok.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	if errored {
+		terminate()
+	}
+}
+
+func ensureDefaultConfig(c *client.Client, node *api.Node) error {
+	// By default we set the global reporting interval to 0 - this is
+	// different from the defaults defined in Felix.
+	//
+	// Logging to file is disabled in the felix.cfg config file.  This
```
caseydavenport: Is this a change from before? I thought we used to log to file.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	//
+	// TODO: This is only required for the current BIRD and GoBGP integrations,
+	//       but should be removed once we switch over to a better watcher interface.
+	if err := ensureGlobalBGPConfig(c, "node_mesh", fmt.Sprintf("{\"enabled\": %v}", client.GlobalDefaultNodeToNodeMesh)); err != nil {
+		return err
+	} else if err := ensureGlobalBGPConfig(c, "as_num", strconv.Itoa(client.GlobalDefaultASNumber)); err != nil {
+		return err
+	} else if err = ensureGlobalBGPConfig(c, "loglevel", client.GlobalDefaultLogLevel); err != nil {
+		return err
+	} else if err = ensureGlobalBGPConfig(c, "LogSeverityScreen", client.GlobalDefaultLogLevel); err != nil {
```
caseydavenport: Huh, isn't this a Felix config not a BGP config? (LogSeverityScreen)

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+		return c.Config().SetBGPConfig(key, "", def)
+	} else {
+		log.Infof("Global BGP value already assigned %s=%s", key, val)
+		return nil
+	}
+}
+
+// message() prints a message to screen and to log.  A newline terminator is
+// not required in the format string.
+func message(format string, args ...interface{}) {
```
caseydavenport: So, I'm pretty sure this is going to show up as duplicate entries in the startup output to stdout, unless you've redirected log.X somewhere else.

I think we could just use `log.X` everywhere and we'd get the behavior we want, but I could be wrong.

Same for the ones below.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+
+// terminate() prints a terminate message and exists with status 1.
+func terminate() {
+	message("Terminating")
+	os.Exit(1)
+}
+
+// netEqual returns true if the two networks are equal.  Neither network
+// can be nil.
+func netEqual(net1, net2 *net.IPNet) bool {
```
caseydavenport: Shouldn't we use `reflect.DeepEqual` for this?

calico_node/tests/startup_test.py
```diff
@@ -1,153 +0,0 @@
-# Copyright (c) 2015-2016 Tigera, Inc. All rights reserved.
```
caseydavenport: We should probably have the test infrastructure in place to reinstate these tests as part of this PR, even if we don't get full coverage.

That way we can at least easily add tests in the future.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+
+// ensureGlobalFelixConfig ensures that the supplied global felix config value
+// is initialized, and if not initialize it with the supplied default.
+func ensureGlobalFelixConfig(c *client.Client, key, def string) error {
+	if val, assigned, err := c.Config().GetFelixConfig(key, ""); err != nil {
+		return err
+	} else if !assigned {
+		return c.Config().SetFelixConfig(key, "", def)
+	} else {
+		log.Infof("Global Felix value already assigned %s=%s", key, val)
```
caseydavenport: Can probably make this a debug log.

Also, do we have a way to configure the log level of the startup procedure itself?

calico_node/autodetection/basicdetection.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	// Find the first interface with a valid IP address and network.
+	// We initialise the IP with the first valid IP that we find just in
+	// case we don't find an IP *and* network.
+	var iface *Interface
+	var addr *Addr
+outer:
+	for _, i := range interfaces {
+		log.WithField("Name", i.Name).Info("Check interface")
```
caseydavenport: I think both of these should be Debug logs as well.

calico_node/autodetection/interfaces.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+
+		if ia.IPAddress == nil {
+			log.WithField("Address", addrStr).Info("Unable to parse IP address")
+			continue
+		}
+
+		if ia.IPAddress.Version() == version {
+			log.WithField("Addr", ia).Info("Storing IP and network in interface")
```
caseydavenport: I think the wording of this log is confusing (and also should be a Debug log).

calico_node/autodetection/interfaces.go
```diff
@@ -0,0 +1,99 @@
...
+		} else {
+			log.Debug("Parsing as IP")
+			ia.IPAddress = &cnet.IP{}
+			if err = ia.IPAddress.UnmarshalText([]byte(addrStr)); err != nil {
+				ia.IPAddress = nil
+			}
+		}
+
+		if ia.IPAddress == nil {
+			log.WithField("Address", addrStr).Info("Unable to parse IP address")
```
caseydavenport: Should this be a warn level? How often do we hit this?

calico_node/autodetection/interfaces.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (a Addr) String() string {
+	return fmt.Sprintf("Addr(IP=%s,Net=%s)", a.IPAddress, a.IPNetwork)
+}
+
+// GetInterfaces returns a list of all interfaces, skipping any interfaces whose
+// name matches any of the exclusion list regexes, and including those on the
+// inclusion list.
+func GetInterfaces(includeRegexes []string, excludeRegexes []string, version int) ([]Interface, error) {
+	// Create a single regex to perform the interface check.
```
robbrockbank: oops i moved the code but not the comment :-/

calico_node/autodetection/interfaces.go
robbrockbank: Yup probably should.

calico_node/autodetection/interfaces.go
robbrockbank: k

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	}
+
+	// Query the current Node resources.  We update our node resource with
+	// updated IP data and use the full list of nodes for validation.
+	node := getNode(client, nodeName)
+
+	// If running in policy only mode, apply the node resource and exit.
+	// This ensures we have an entry that can be queried, and also ensures
+	// the datastore is initialized.
+	if strings.ToLower(os.Getenv("CALICO_NETWORKING_BACKEND")) == "none" {
```
robbrockbank: Hmmm, I've no idea at all why I added that.  Odd.  Good catch.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+		terminate()
+	}
+
+	// Write the startup.env file now that we are ready to start other
+	// components.
+	ipv6Str := ""
+	if node.Spec.BGP.IPv6Address != nil {
+		ipv6Str = node.Spec.BGP.IPv6Address.String()
+	}
+	writeStartupEnv(nodeName, node.Spec.BGP.IPv4Address.String(), ipv6Str)
```
robbrockbank: At the moment we always assume an IPv4 address is specified.  That's how it used to be and we hadn't yet changed that behavior.  I'll update the code though so that both or either can be nil -- I think I hacked this in when I was testing, so it does need a little work.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	client, err := client.New(*cfg)
 	if err != nil {
-		panic(fmt.Sprintf("Error creating client: %s", err))
+		fatal("Error accessing the Calico datastore: %s", err)
+		terminate()
+	}
+
+	// An explicit value of false is required to skip waiting for the
+	// datastore.
+	if os.Getenv("WAIT_FOR_DATASTORE") == "false" {
```
robbrockbank: Good spot.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+		message("Skipping datastore connection test")
+		return cfg, client
+	}
+
+	message("Checking datastore connection")
+	for i := 0; i <= 120; i++ {
```
robbrockbank: I just couldn't bring myself to make it loop indefinitel, but you're probably right let's keep the scope tight.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	// Write out the startup.env file to ensure require environments are
+	// set (which they might not otherwise be).
+	if err := ioutil.WriteFile("startup.env", []byte(text), 0666); err != nil {
+		log.WithError(err).Info("Unable to write to startup.env")
+		fatal("Unable to write to local filesystem")
+		terminate()
+	}
+}
+
+// getNode returns the current node configuration and a list of all configured
```
robbrockbank: well I may have changed this at some point and not updated the comment..... whistle .. la la la

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+
+	} else {
+		node.Spec.BGP.IPv4Address, _ = fetchAndValidateIPAndNetwork(
+			ipv4Env, 4,
+			node.Spec.BGP.IPv4Address, nil)
+	}
+
+	ipv6Env := os.Getenv("IP6")
+	if ipv6Env == "autodetect" {
+		adm := os.Getenv("IP6_AUTODETECT_METHOD")
```
robbrockbank: I think we should not disallow that!  And if the method is "ping a particular address" rather than "ping" (with a separate address parm) then we deffo will.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	// Determine the autodetection type for IPv4 and IPv6.  Note that we
+	// only autodetect IPv4 when it has not been specified.  IPv6 must be
+	// explicitly requested using the "autodetect" value.
+	//
+	// If we aren't auto-detecting then we need to validate the configured
+	// value and possibly fix up missing subnet configuration.
+	ipv4Env := os.Getenv("IP")
+	if ipv4Env == "autodetect" || (ipv4Env == "" && node.Spec.BGP.IPv4Address == nil) {
+		adm := os.Getenv("IP_AUTODETECT_METHOD")
+		//node.Spec.BGP.IPv4Address, node.Spec.BGP.IPv4Network = autoDetectIPAndNetwork(adm, 4)
```
robbrockbank: indeed as can the similar line below.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+			node.Spec.BGP.IPv4Address, nil)
+	}
+
+	ipv6Env := os.Getenv("IP6")
+	if ipv6Env == "autodetect" {
+		adm := os.Getenv("IP6_AUTODETECT_METHOD")
+		//node.Spec.BGP.IPv6Address, node.Spec.BGP.IPv6Network = autoDetectIPAndNetwork(adm, 6)
+		node.Spec.BGP.IPv6Address, _ = autoDetectIPAndNetwork(adm, 6)
+	} else {
+		node.Spec.BGP.IPv6Address, _ = fetchAndValidateIPAndNetwork(
```
robbrockbank: I've reworked all the fetch/validate stuff.  I really wanted to do that before I left but I ran out of time... so it's much simplified now and does not muddy this PR with any subnet changes.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+			ipv6Env, 6,
+			node.Spec.BGP.IPv6Address, nil)
+	}
+
+}
+
+// fetchAndValidateIPAndNetwork fetches and validates the IP configuration from
+// either the environments, from pre-configured values.  In addition, missing
```
robbrockbank: As mentioned above.  I've reworked this to how I actually want it for this PR.  Much simplified and no muddying the waters with subnet stuff.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+}
+
+// configureASNumber configures the Node resource with the AS numnber specified
+// in the environment.
+func configureASNumber(node *api.Node) {
+	// Extract the AS number from the environment
+	asStr := os.Getenv("AS")
+	if asStr != "" {
+		if asNum, err := numorstring.ASNumberFromString(asStr); err != nil {
+			fatal("The AS number specified in the environment is not valid: %s", asStr)
```
robbrockbank: Yeah it might be useful.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+			terminate()
+		} else {
+			message("Using AS number specified in environment: %s", asNum)
+			node.Spec.BGP.ASNumber = &asNum
+		}
+	} else {
+		message("AS number not specified in environment, using current value")
```
robbrockbank: I was just thinking the same thing ...

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
-	log.Infof("Assigned ClusterGUID %s", guid)
+
+	// Get a list of all IP Pools
+	poolList, err := client.IPPools().List(api.IPPoolMetadata{})
+	if err != nil {
+		fatal("Unable to fetch IP pool list: %s", err)
+		terminate()
+	}
+
+	// Check for IPv4 and IPv6 pools and filter for IPIP pools.
```
robbrockbank: Ah no - we're not.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+		ipv4Present = ipv4Present || (version == 4)
+		ipv6Present = ipv6Present || (version == 6)
+		if ipv4Present && ipv6Present {
+			break
+		}
+	}
+
+	// Ensure there are pools created for each IP version.
+	if !ipv4Present {
+		createIPPool(client, DEFAULT_IPV4_POOL_CIDR)
```
robbrockbank: we could.... but as you say, not for this PR.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	}
+	if !ipv6Present && ipv6Supported() {
+		createIPPool(client, DEFAULT_IPV6_POOL_CIDR)
+	}
+}
+
+// ipv6Supported returns true if IPv6 is supported on this platform.  This performs
+// a simplistic check of /proc/sys/net/ipv6 since platforms that do not have IPv6
+// compiled in will not have this entry.
+func ipv6Supported() bool {
```
robbrockbank: Yeah that is probably not a bad idea.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	log.Infof("IPv6 supported on this platform: %v", supported)
+	return supported
+}
+
+// createIPPool creates an IP pool using the specified CIDR string.
+func createIPPool(client *client.Client, cs string) {
+	_, cidr, _ := net.ParseCIDR(cs)
+	version := cidr.Version()
+
+	log.Info("Creating default IPv%d pool", version)
```
robbrockbank: Not sure I'm too worried here.  We only call this function if we think we need to create a pool.  The only time we won't is when two nodes are concurrently trying to create the pool and the other node gets there first.   So in general we won't log this every time a node starts.

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+	pool := &api.IPPool{
+		Metadata: api.IPPoolMetadata{
+			CIDR: *cidr,
+		},
+		Spec: api.IPPoolSpec{
+			NATOutgoing: true,
+		},
+	}
+	if _, err := client.IPPools().Apply(pool); err != nil {
+		if _, ok := err.(errors.ErrorResourceAlreadyExists); !ok {
```
robbrockbank: Oooh!  Yes it should have been Create ... that's exactly why I was checking for that error type.  D'oh.  Ta.

calico_node/startup.go
robbrockbank: Dammit - it's gonna irritate me now that you mention it though.  lol

calico_node/startup.go
```diff
@@ -14,59 +14,657 @@
...
+		}
+		if theirIPv6 != nil && theirIPv6.Equal(ourIPv6.IP) {
+			message("Calico node '%s' is already using the IPv6 address %s:",
+				theirNode.Metadata.Name, ourIPv6)
+			message(" -  Check the node configuration to remove the IP address conflict")
+			errored = true
+		}
+	}
+
+	if errored {
```
robbrockbank: We can, but it means that we won't handle IPv4 and IPv6 conflicts if the conflicts are on different nodes - so we'll only log the first error and then exit.  I think how it is is probably the "correct" way.

calico_node/startup.go
robbrockbank: No we log to screen, but screen output goes to file.

calico_node/startup.go
robbrockbank: Ooooh that should not be there.... cut and paste error.

calico_node/startup.go
robbrockbank: Yup, and I should be using WithField() too.

calico_node/startup.go
robbrockbank: I don't think so actually.  There are a couple of different ways to store IP addresses since golang stores IPv4 as either 4-byte or 16-byte arrays....

Anyways - not required for this work, so I'm removing.

calico_node/startup.go
robbrockbank: I have tested this so I can state that the output from running calicoctl node run only outputs once and the the logs are not output to screen.

The reason is... by default logrus logs to stderr at info level.  So, the prints go to stdout, and it is stdout that we trace in calicoctl node run.   Logs go to stderr.  This means that the docker logs contain both stdout and stderr, so indeed the docker logs will contain duplicates.  I personally think this is OK as it's only a small handful of logs we output to screen.

I personally think this is a good bang-for-bucks solution.  Otherwise `calicoctl node run` would need to parse the logs and remove the timestamp and filter out certain log types etc.

caseydavenport: I'm not so sure.  Docker logs are important since that's the main way that we do logging in Kubernetes when using the self-hosted approach, so this means one of our most used deployment mechanisms will show duplicate logs.

robbrockbank:
Modified by #1447

PR #1430, opened 2016-12-26

### LINKED ISSUE #1429: `calicoctl node run` cmd fails: /usr/bin/docker: no such file or directory
The command fails since **calico/ctl** image was switched to use a common alpine docker image instead of custom one with **glibc**:

https://github.com/projectcalico/calicoctl/commit/6bceaa7b6dd82ae9fc8a6723d4c9931007766eea#diff-7c036e281666f41844cdd3a4e834e1ebR1

Example:

```
apanchenko@apanchenko-supermicro:~$ docker pull calico/ctl:v1.0.0-2-g57e67db
v1.0.0-2-g57e67db: Pulling from calico/ctl
3690ec4760f9: Already exists
10149d82994c: Pull complete
ce5e61ecd29c: Pull complete
Digest: sha256:f6fac6131b4236899c7448d2a01d9e4ab140488e6cc14357a03cadc220247167
Status: Downloaded newer image for calico/ctl:v1.0.0-2-g57e67db

apanchenko@apanchenko-supermicro:~$ /usr/bin/docker run -it --privileged --rm --net=host --pid=host --entrypoint /bin/sh -e ETCD_ENDPOINTS=https://node1:2379,https://node2:2379,https://node3:2379 -e ETCD_CA_CERT_FILE=/etc/calico/certs/ca_cert.crt -e ETCD_CERT_FILE=/etc/calico/certs/cert.crt -e ETCD_KEY_FILE=/etc/calico/certs/key.pem -v /usr/bin/docker:/usr/bin/docker -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/calico:/var/run/calico -v /etc/calico/certs:/etc/calico/certs:ro calico/ctl:v1.0.0-2-g57e67db
/ # docker ps
/bin/sh: docker: not found
/ # 
/ # ./calicoctl node run
Removing old calico-node container (if running).
Running the following command to start calico-node:

docker run --net=host --privileged --name=calico-node -d --restart=always -e CALICO_NETWORKING_BACKEND=bird -e CALICO_LIBNETWORK_IFPREFIX=cali -e ETCD_ENDPOINTS=https://node1:2379,https://node2:2379,https://node3:2379 -e ETCD_SCHEME= -e ETCD_CA_CERT_FILE=/etc/calico/certs/ca_cert.crt -e ETCD_CERT_FILE=/etc/calico/certs/cert.crt -e NODENAME=apanchenko-supermicro -e NO_DEFAULT_POOLS= -e CALICO_LIBNETWORK_ENABLED=true -e ETCD_AUTHORITY= -e ETCD_KEY_FILE=/etc/calico/certs/key.pem -v /var/log/calico:/var/log/calico -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock -v /etc/calico/certs/ca_cert.crt:/etc/calico/certs/ca_cert.crt -v /etc/calico/certs/key.pem:/etc/calico/certs/key.pem -v /etc/calico/certs/cert.crt:/etc/calico/certs/cert.crt -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules calico/node:v1.0.0-2-g57e67db

Image may take a short time to download if it is not available locally.
Error executing command: fork/exec /usr/bin/docker: no such file or directory
```

artem-panchenko:
glibc.

Closes #1429

### DIFF
--- a/calicoctl/Dockerfile.calicoctl
+++ b/calicoctl/Dockerfile.calicoctl
@@ -5,7 +5,15 @@ ADD dist/calicoctl ./calicoctl
 
 ENV CALICO_CTL_CONTAINER=TRUE
 
-# libltdl.so is needed by docker command line tool
-RUN apk add --no-cache libltdl
+# glibc and libltdl are needed by docker command line tool
+ENV GLIBC_VERSION 2.23-r3
+RUN apk add --no-cache libltdl 
+RUN apk add --no-cache --update wget openssl ca-certificates && \
+  wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
+  wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-${GLIBC_VERSION}.apk && \
+  wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-bin-${GLIBC_VERSION}.apk && \
+  apk add glibc-${GLIBC_VERSION}.apk glibc-bin-${GLIBC_VERSION}.apk && \
+  rm -f glibc-${GLIBC_VERSION}.apk glibc-bin-${GLIBC_VERSION}.apk && \
+  apk del openssl ca-certificates
 
 ENTRYPOINT ["./calicoctl"]


calicoctl/Dockerfile.calicoctl
```diff
@@ -5,7 +5,15 @@ ADD dist/calicoctl ./calicoctl
...
+# glibc and libltdl are needed by docker command line tool
+ENV GLIBC_VERSION 2.23-r3
+RUN apk add --no-cache libltdl 
+RUN apk add --no-cache --update wget openssl ca-certificates && \
+  wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
+  wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-${GLIBC_VERSION}.apk && \
+  wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-bin-${GLIBC_VERSION}.apk && \
+  apk add glibc-${GLIBC_VERSION}.apk glibc-bin-${GLIBC_VERSION}.apk && \
+  rm -f glibc-${GLIBC_VERSION}.apk glibc-bin-${GLIBC_VERSION}.apk && \
+  apk del openssl ca-certificates
```
robbrockbank: Can we also remove wget?

robbrockbank:
One minor comment/question.

I assume we should really be fixing this by using the golang Docker client rather than shelling out to the docker CLI - but this seems like a sensible temporary work around.

calicoctl/Dockerfile.calicoctl
artem-panchenko: Sure, we can if you want, but it's installed by default:

```
apanchenko@apanchenko-supermicro:~$ docker run -it --rm alpine:3.4 wget
BusyBox v1.24.2 (2016-08-12 14:38:34 GMT) multi-call binary.

Usage: wget [-csq] [-O FILE] [-Y on/off] [-P DIR] [-U AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

        -s      Spider mode - only check file existence
        -c      Continue retrieval of aborted transfer
        -q      Quiet
        -P DIR  Save to DIR (default .)
        -T SEC  Network read timeout is SEC seconds
        -O FILE Save to FILE ('-' for stdout)
        -U STR  Use STR for User-Agent header
        -Y      Use proxy ('on' or 'off')
```

That's why I don't know if `calicoctl` uses it somehow :)

artem-panchenko: @robbrockbank so should I remove wget or we can keep this as is? We can't use (with kubernetes) **calico/ctl** built from master for last 2 weeks because of #1429 and that's why we would like to fix it ASAP. Thanks in advance!

marvin-tigera:
Can one of the admins verify this patch?

robbrockbank:
So sorry about the delay in merging this.  LGTM.

robbrockbank:
Hi @artem-panchenko 

So after chatting with my colleagues, our preferred approach to adding glibc to the alpine image is documented here:

https://github.com/sgerrand/alpine-pkg-glibc

Regarding the removal of wget, whilst it is true that the base image includes wget - I believe it is a parred down version of wget - so doing the install will install a more feature rich and therefore larger version of wget.  Although it won't be a large difference, we probably should delete it once the install is finished.

calicoctl/Dockerfile.calicoctl
robbrockbank: Hi @artem-panchenko 

I'd like to get this checked in for you.  Couple of things:
-  For completeness could you remove wget.
-  Also, do we need the `glibc-bin` package?  From looking at https://github.com/sgerrand/alpine-pkg-glibc I'm wondering if we don't need that package unless we explicitly need some of the binaries.  If we don't need it can we also remove the glib-bin wget and api add

artem-panchenko: Hi @robbrockbank !

1. Done, I removed wget. BTW, the tool is still there after deletion, because busybox also provides it: http://paste.openstack.org/show/597976/
2. Yep, seems we (docker) need it http://paste.openstack.org/show/597975/

artem-panchenko:
I mistakenly removed the fork which this PR was created from (forked renamed calicoctl repo instead). So I had to create a new PR https://github.com/projectcalico/calicoctl/pull/1503 and closing this one

PR #1435, opened 2017-01-03
port allocateIPIPAddr from python

2blesd2bstresd:
Should allocateIPIPAddr be called from startup.go or is it a separate for a reason?


calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
+package allocateIPIPAddr
+
+import (
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/client"
+	"debug/pe"
```
robbrockbank: debug/pe :  what's this for?  I don't think it's used.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+func allocateIPIPAddr(client *client.Client, string nodename) {
+	meta := api.IPPoolMetadata{}
+	ipPoolList, err := client.IPPools().List(meta)
+	if err != nil {
+		log.Warnf("error retrieving pools: %s", err)
+	}
+    ipv4Pools := getIPv4Pools(ipPoolList.Items)
```
robbrockbank: Mixing tabs and spaces here.   Make sure you're using an IDE that supports go.

I'm using PyCharm with a golang plugin.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+    } else {
+        panic(fmt.Sprintf("Failed to allocate an IP address from an IPIP-enabled pool 
+for the host's IPIP tunnel device.  Pools are likely exhausted."))
+    } 
+}
+
+func findPool(ipAddr *net.IP, ipv4Pools []api.IPPool) (net.IPNet, error) {
+    for _, p in range ipv4Pools {
+        poolIP := p.IPPoolMetadata.CIDR.String()
+        if strings.HasPrefix(poolIP, ipAddr) {
```
robbrockbank: You cannot use string prefix to determine if an IP is within a CIDR.

The IPNetwork field has a Contains() method that determines if an IP is in a CIDR

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+            return p, nil
+        }
+    }
+    return nil, nil
+}
+
+func getIPv4Pools(ipPools []api.IPPool) {
+    result := []api.IPPool{}
+    for _, ipPool in ipPoolList.Items {
+        if ipPool.Metadata.CIDR.version == 4 {
```
robbrockbank: version -> Version() ?   lowercase variables are not public.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+            append(result, ipPool)
+        }
+    }
+    return result
+}
+
+func getIPIPEnabledPools(ipPools []api.IPPool) {
+    result := []api.IPPool{}
+    for _, ipPool in ipPoolList.Items {
+        if ipPool.Spec.IPIP.enabled == true {
```
robbrockbank: enabled -> Enabled

robbrockbank: Also, if Enabled is a boolean, you don't need "== true".

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+func ensureHostTunnelAddress(ipv4Pools, enabledPools []api.IPPool, client *client.Client, string nodename){
+    cnf = client.Config()
+    ipAddr, err := cnf.GetNodeIPIPTunnelAddress(nodename)
+    if err != nil {
+        panic(fmt.Sprintf("Could not retrieve IPIP tunnel address: %s", err))
+    }
+    if ipAddr != nil {
+        pool, _ := findPool(ipAddr, ipv4Pools)
+        if pool != nil {
+            if pool.Spec.IPPoolSpec.IPIP == false {
```
robbrockbank: pool.Spec.IPIP != nil && pool.Spec.IPIP.Enabled {}

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+    } else {
+        removeHostTunnelAddr(client, nodename)
+    }
+}
+
+func ensureHostTunnelAddress(ipv4Pools, enabledPools []api.IPPool, client *client.Client, string nodename){
+    cnf = client.Config()
+    ipAddr, err := cnf.GetNodeIPIPTunnelAddress(nodename)
+    if err != nil {
+        panic(fmt.Sprintf("Could not retrieve IPIP tunnel address: %s", err))
```
robbrockbank: Panics output stack trace.  I don't think we want to be doing that.  Better to just output a sensible error message and exit with a non-zero RC.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+    return nil, nil
+}
+
+func getIPv4Pools(ipPools []api.IPPool) {
+    result := []api.IPPool{}
+    for _, ipPool in ipPoolList.Items {
+        if ipPool.Metadata.CIDR.version == 4 {
+            append(result, ipPool)
```
robbrockbank: result = append(result, ipPool)

append may return a new slice if it needs to increase the size of the reserved memory chunk for the slice.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+        if strings.HasPrefix(poolIP, ipAddr) {
+            return p, nil
+        }
+    }
+    return nil, nil
+}
+
+func getIPv4Pools(ipPools []api.IPPool) {
+    result := []api.IPPool{}
+    for _, ipPool in ipPoolList.Items {
```
robbrockbank: for x in y is Python

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+func assignHostTunnelAddr(ipipPools []api.IPPool, *client.Client, nodename string) {
+    args := AutoAssignArgs{Num4: 1, Num6:0, HandleID: nil, Attrs: nil, Hostname: nodename, IPv4Pools: ipipPools}
+    ipam := client.IPAM()
+    ipv4Addrs, _, err := ipam.AutoAssign(args)
+    if err != nil {
+        panic(fmt.Sprintf("Could not autoassign host tunnel address: %s", err))
+    }
+    if len(ipv4Addrs) > 0 {
+        config := client.Config()
+        config.SetFelixConfig("IpInIpTunnelAddr", nodename, ipv4Addrs[0].String())
```
robbrockbank: Use `SetNodeIPIPTunnelAddress`

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+        panic(fmt.Sprintf("Could not retrieve IPIP tunnel address for cleanup: %s", err))
+    }
+    if ipAddr != nil {
+        _, err := client.IPAM().ReleaseIPs([]*net.IP{ipAddr})
+        if err != nil {
+            panic(fmt.Sprintf("Error releasing address: %s", err))
+        }
+    }
+    ipam := client.IPAM()
+    err := ipam.UnsetFelixConfig("IpInIpTunnelAddr", nodename)
```
robbrockbank: Use SetNodeIPIPTunnelAddress() which is part of the Config() interface.  Set it to nil.

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,112 @@
...
+
+func allocateIPIPAddr(client *client.Client, string nodename) {
+	meta := api.IPPoolMetadata{}
+	ipPoolList, err := client.IPPools().List(meta)
+	if err != nil {
+		log.Warnf("error retrieving pools: %s", err)
+	}
+    ipv4Pools := getIPv4Pools(ipPoolList.Items)
+    enabledPools := getIPIPEnabledPools(ipv4Pools)
+    if len(enabledPools) > 0 {
```
robbrockbank: I don't think we need both IPv4 and IPIP pools - it should be sufficient to just use the list of IPIP pools - if the tunl address is not in an IPIP pool then remove it and assign an address from an IPIP pool.

Makefile
```diff
@@ -45,7 +45,7 @@ NODE_CONTAINER_NAME?=calico/node
 NODE_CONTAINER_FILES=$(shell find $(NODE_CONTAINER_DIR)/filesystem -type f)
 NODE_CONTAINER_CREATED=$(NODE_CONTAINER_DIR)/.calico_node.created
 NODE_CONTAINER_BIN_DIR=$(NODE_CONTAINER_DIR)/filesystem/bin
-NODE_CONTAINER_BINARIES=startup startup-go allocate-ipip-addr calico-felix bird calico-bgp-daemon confd libnetwork-plugin
+NODE_CONTAINER_BINARIES=startup startup-go allocate-ipip-addr allocateIPIPAddr calico-felix bird calico-bgp-daemon confd libnetwork-plugin
```
robbrockbank: We already have the allocate-ipip-addr make target/binary.  I think we should just use that rather than adding another allocateIPIPAddr

calico_node/allocate_ipip_addr/allocate_ipip_addr.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	nodename := os.Getenv("NODENAME")
+
+	meta := api.IPPoolMetadata{}
+	ipPoolList, err := c.IPPools().List(meta)
+	if err != nil {
+		log.Warnf("error retrieving pools: %s", err)
+	}
+	fmt.Println(ipPoolList)
+	enabledPools := getIPIPEnabledPools(ipPoolList.Items)
```
robbrockbank: `getIPIPEnabledPools()` is really more of a "filter" than a "get".  I would move lines 27-32 into the getIPIPEnabledPools() method so that all the pools enumeration is encapsulated together,

robbrockbank:
Modified by #1447

PR #1443, opened 2017-01-06
Add --docker-host flag to pass DOCKER_HOST env var to calico node

### LINKED ISSUE #1404: Pass DOCKER_HOST to calico/node
See https://github.com/projectcalico/libnetwork-plugin/issues/90#issuecomment-266078037 for detail

gunjan5:
Fixes #1404

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -105,6 +105,9 @@ Options:
                            and there are no pre-existing Calico IP pools.
      --disable-docker-networking
                            Disable Docker networking.
+     --docker-host=<DOCKER_HOST>
+                           Docker host address where Docker daemon is running.
+                           [default: unix:///var/run/docker.sock]
      --docker-networking-ifprefix=<IFPREFIX>
                            Interface prefix to use for the network interface
                            within the Docker containers that have been networked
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/run.go
```diff
@@ -213,9 +217,10 @@ Description:
...
 	}
 
-	// Set CALICO_LIBNETWORK_IFPREFIX env variable if Docker network is enabled.
+	// Set DOCKER_HOST and CALICO_LIBNETWORK_IFPREFIX env variables if Docker network is enabled.
 	if !disableDockerNw {
 		envs["CALICO_LIBNETWORK_IFPREFIX"] = ifprefix
+		envs["DOCKER_HOST"] = dockerhost
 	}
 
 	// Add in optional environments.
```
robbrockbank: See line 247 -  you also need to change your volume mounts

gunjan5:
@robbrockbank not sure what I would have to mount if the docker host isn't the default local unix socket. It would be something like this `DOCKER_HOST=tcp://172.17.8.101:2376` what would I mount in this case?

robbrockbank:
I assume for tcp://  format we don't need to volume mount in the socket.  But for unix:// we need to volume mount in the socket to the expected location.

Probably worthy of a separate issue, we should consider adding the ability to specify certificates used for TLS authentication on the Docker TCP connection (see DOCKER_CERT_PATH).

robbrockbank:
Sorry pressed the wrong button and accidentally closed.  Reopened now :-)

tomdee:
@gunjan5 @robbrockbank What's happening with this PR?

gunjan5:
closing this for now.

PR #1447, opened 2017-01-09
All go startup and ipip allocation

robbrockbank:


calico_node/allocateipip/allocate_ipip_addr.go
```diff
@@ -0,0 +1,139 @@
...
+// with some space. Stores the result in the host's config as its tunnel
+// address.
+func assignHostTunnelAddr(c *client.Client, nodename string, ipipCidrs []net.IPNet) {
+	args := client.AutoAssignArgs{
+		Num4:      1,
+		Num6:      0,
+		HandleID:  nil,
+		Attrs:     nil,
+		Hostname:  nodename,
+		IPv4Pools: ipipCidrs,
```
robbrockbank: yeah i was thinking that too as I was playing with this.

calico_node/startup/autodetection/filtered.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+
+			// If the address has an IP and a Network then return
+			// it.  Otherwise, if it just as an IP and we haven't
+			// found one yet store this and return it if we don't
+			// find one with a network.
+			if a.IPNetwork != nil {
+				iface = &i
+				addr = &a
+				break outer
```
robbrockbank: I was kinda thinking the same too.  I almost removed it and then thought it'd be good just on the off chance -- but that said, it's untested code so perhaps we should just get rid of it.

calico_node/startup/autodetection/interfaces.go
```diff
@@ -0,0 +1,117 @@
...
+	return fmt.Sprintf("Addr(IP=%s,Net=%s)", a.IPAddress, a.IPNetwork)
+}
+
+// GetInterfaces returns a list of all interfaces, skipping any interfaces whose
+// name matches any of the exclusion list regexes, and including those on the
+// inclusion list.
+func GetInterfaces(includeRegexes []string, excludeRegexes []string, version int) ([]Interface, error) {
+	netIfaces, err := net.Interfaces()
+	if err != nil {
+		log.Warnf("Interfaces unavailable: %v", err)
```
robbrockbank: Indeed and it should use WithError()

calico_node/startup/startup.go
```diff
@@ -0,0 +1,626 @@
...
+	// Create the Calico API client.
+	cfg, client := createClient()
+
+	// If this is a Kubernetes backed datastore then just make sure the
+	// datastore is initialized and then exit.  We don't need to explicitly
+	// initialize the datastore for non-Kubernetes because the node resource
+	// management will do that for us.
+	if cfg.Spec.DatastoreType == api.Kubernetes {
+		message("Calico is using a Kubernetes datastore")
+		err = client.EnsureInitialized()
```
robbrockbank: See above comment.

calico_node/startup/startup.go
```diff
@@ -0,0 +1,600 @@
...
+	}
+
+	if cidr == nil {
+		message("Unable to auto-detect an IPv%d address", version)
+		return nil
+	}
+
+	message("Using autodetected IPv%d address on interface %s: %s", version, iface.Name, cidr.String())
+
+	return &net.IP{cidr.IP}
```
robbrockbank: I realize this is just returning the IP address at the moment -- but once the libcalico-go changes are in this will be updated to just return the CIDR as per the interface updates to the node resource.

calico_node/startup/startup.go
```diff
@@ -140,36 +148,14 @@ func determineNodeName() string {
...
-	}
-
-	// An explicit value of false is required to skip waiting for the
-	// datastore.
-	if os.Getenv("WAIT_FOR_DATASTORE") != "true" {
-		message("Skipping datastore connection test")
-		return cfg, client
-	}
-
+// waitForConnection() waits for the datastore to become accessible.
```
gunjan5: nit: the `()` part in comment will look weird in GoDocs

calico_node/startup/startup.go
```diff
@@ -140,36 +148,14 @@ func determineNodeName() string {
...
+func waitForConnection(c *client.Client) {
 	message("Checking datastore connection")
 	for {
 		// Query some arbitrary configuration to see if the connection
 		// is working.  Getting a specific node is a good option, even
 		// if the node does not exist.
-		_, err = client.Nodes().Get(api.NodeMetadata{Name: "foo"})
+		_, err := c.Nodes().Get(api.NodeMetadata{Name: "foo"})
```
gunjan5: why can't we use `EnsureInitialized()` here?

calico_node/startup/startup.go
```diff
@@ -602,21 +579,18 @@ func ensureGlobalBGPConfig(c *client.Client, key, def string) error {
 // not required in the format string.
 func message(format string, args ...interface{}) {
 	fmt.Printf(format+"\n", args...)
-	log.Infof(format, args...)
 }
 
 // warning() prints a warning to screen and to log.  A newline terminator is
```
gunjan5: `()` can go

calico_node/startup/startup.go
```diff
@@ -602,21 +579,18 @@ func ensureGlobalBGPConfig(c *client.Client, key, def string) error {
...
 // not required in the format string.
 func warning(format string, args ...interface{}) {
 	fmt.Printf("WARNING: "+format+"\n", args...)
-	log.Warnf(format, args...)
 }
 
 // fatal() prints a fatal message to screen and to log.  A newline terminator is
```
gunjan5: `()` can go

calico_node/startup/startup.go
```diff
@@ -602,21 +579,18 @@ func ensureGlobalBGPConfig(c *client.Client, key, def string) error {
 // not required in the format string.
 func message(format string, args ...interface{}) {
 	fmt.Printf(format+"\n", args...)
-	log.Infof(format, args...)
 }
 
 // warning() prints a warning to screen and to log.  A newline terminator is
 // not required in the format string.
 func warning(format string, args ...interface{}) {
 	fmt.Printf("WARNING: "+format+"\n", args...)
```
gunjan5: whouldn't this start with a `\n` if it's Printf? I'm thinking if previous print line didn't end with a `\n` this would end up in the same line

gunjan5:
just reviewed the last commit, and just a few tiny comments. Also, using variable name `interface` could throw people off since it's also a Golang keyword. But not a big deal, I can live with that variable name :)

robbrockbank:
actually using interface as a variable name is a no-no ... thought i'd fixed all those places, I guess not.  I'll fix that!

calico_node/startup/startup.go
robbrockbank: Why would the previous line not end in a \n ?
What if the next line doesn't being with a \n ?

calico_node/startup/startup.go
robbrockbank: We could - it's fewer queries to do this I think and EnsureInitialized will be called as part of the node resource management so yeah we could, but it's no biggy either way.

PR #1453, opened 2017-01-13
Add cross-subnet IPIP option to IP Pool

robbrockbank:
This needs https://github.com/projectcalico/libcalico-go/pull/308 to be merged and:
-  glide.lock and glide.yaml updated (remove the rob brockbank repo lines)

### DIFF
--- a/calico_node/filesystem/etc/calico/confd/templates/bird6.cfg.mesh.template
+++ b/calico_node/filesystem/etc/calico/confd/templates/bird6.cfg.mesh.template
@@ -2,11 +2,13 @@
 include "bird6_aggr.cfg";
 include "custom_filters6.cfg";
 include "bird6_ipam.cfg";
+{{$node_ip_key := printf "/host/%s/ip_addr_v4" (getenv "NODENAME")}}{{$node_ip := getv $node_ip_key}}
+{{$node_ip6_key := printf "/host/%s/ip_addr_v6" (getenv "NODENAME")}}{{$node_ip6 := getv $node_ip6_key}}
 
-router id {{getenv "IP"}};  # Use IPv4 address since router id is 4 octets, even in MP-BGP
+router id {{$node_ip}};  # Use IPv4 address since router id is 4 octets, even in MP-BGP
 
 {{define "LOGGING"}}
-{{$node_logging_key := printf "/host/%s/loglevel" (getenv "HOSTNAME")}}{{if exists $node_logging_key}}{{$logging := getv $node_logging_key}}
+{{$node_logging_key := printf "/host/%s/loglevel" (getenv "NODENAME")}}{{if exists $node_logging_key}}{{$logging := getv $node_logging_key}}
 {{if eq $logging "debug"}}  debug all;{{else if ne $logging "none"}}  debug { states };{{end}}
 {{else if exists "/global/loglevel"}}{{$logging := getv "/global/loglevel"}}
 {{if eq $logging "debug"}}  debug all;{{else if ne $logging "none"}}  debug { states };{{end}}
@@ -39,8 +41,8 @@ protocol direct {
   interface -"cali*", "*"; # Exclude cali* but include everything else.
 }
 
-{{if eq "" (getenv "IP6")}}# IPv6 disabled on this node.
-{{else}}{{$node_as_key := printf "/host/%s/as_num" (getenv "HOSTNAME")}}
+{{if eq "" ($node_ip6)}}# IPv6 disabled on this node.
+{{else}}{{$node_as_key := printf "/host/%s/as_num" (getenv "NODENAME")}}
 # Template for all BGP clients
 template bgp bgp_template {
   {{template "LOGGING"}}
@@ -53,7 +55,7 @@ template bgp bgp_template {
   export filter calico_pools;  # Only want to export routes for workloads.
   next hop self;     # Disable next hop processing and always advertise our
                      # local address as nexthop
-  source address {{getenv "IP6"}};  # The local address we use for the TCP connection
+  source address {{$node_ip6}};  # The local address we use for the TCP connection
   add paths on;
   graceful restart;  # See comment in kernel section about graceful restart.
 }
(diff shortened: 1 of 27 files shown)


calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -2,20 +2,29 @@
...
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
   if ( net ~ {{$data.cidr}} ) then {
     accept;
   }
 {{end}}
   reject;
 }
 
+{{$network_key := printf "/bgp/v1/host/%s/network_v4" (getenv "NODENAME")}}{{$network := getv $network_key}}
 filter calico_ipip {
```
caseydavenport: I'd like to see some comments in this template to explain a bit about what each branch is meant to do.

I think I understand the flow, but it takes a bit of reading to figure it out without any comments.

calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -2,20 +2,29 @@
...
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
   if ( net ~ {{$data.cidr}} ) then {
+{{if eq $data.ipip_mode "cross-subnet"}}
+    if ( from ~ {{$network}} ) then
+      krt_tunnel = "";
+    else
+      krt_tunnel = "{{$data.ipip}}";
+    accept;
+  } {{else}}
     krt_tunnel = "{{$data.ipip}}";
```
caseydavenport: IIUC, this line will get executed even if `ipip` is not enabled, but when not enabled `$data.ipip` will be `""` and so traffic won't go over the tunnel?  Just trying to make sure I understand what I'm reviewing :)

calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
```diff
@@ -2,20 +2,29 @@
...
   reject;
 }
 
+{{$network_key := printf "/bgp/v1/host/%s/network_v4" (getenv "NODENAME")}}{{$network := getv $network_key}}
 filter calico_ipip {
-{{range ls "/pool"}}{{$data := json (getv (printf "/pool/%s" .))}}{{if $data.ipip}}
+{{range ls "/v1/ipam/v4/pool"}}{{$data := json (getv (printf "/v1/ipam/v4/pool/%s" .))}}
   if ( net ~ {{$data.cidr}} ) then {
+{{if eq $data.ipip_mode "cross-subnet"}}
+    if ( from ~ {{$network}} ) then
```
caseydavenport: `from` in this case is the IP of the BGP peer who sent us the route, right?

calico_node/filesystem/etc/calico/confd/templates/bird6.cfg.mesh.template
```diff
@@ -65,7 +67,7 @@ template bgp bgp_template {
 {{$onode_ip_key := printf "/host/%s/ip_addr_v6" .}}{{$onode_ip := getv $onode_ip_key}}
 {{$nums := split $onode_ip ":"}}{{$id := join $nums "_"}}
 # For peer {{$onode_ip_key}}
-{{if eq $onode_ip (getenv "IP6") }}# Skipping ourselves ({{getenv "IP6"}})
+{{if eq $onode_ip ($node_ip6) }}# Skipping ourselves ({{$node_ip6}})
```
robbrockbank: That is correct - although it is possible to change the IP address by updating the node resource... which means if you restart the node with the IP flag it'll update it back again.

I think that behavior is fine because of the change we made to the default behavior of autodetecting only if a value has not yet been assigned.

calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
robbrockbank: yup

calico_node/filesystem/etc/calico/confd/templates/bird_ipam.cfg.template
robbrockbank: that is correct.

calico_node/startup/startup.go
```diff
@@ -268,9 +272,12 @@ func configureIPsAndSubnets(node *api.Node) {
 // fetchAndValidateIPAndNetwork fetches and validates the IP configuration from
```
robbrockbank: I think you could be correct.

calico_node/startup/startup.go
```diff
@@ -282,12 +289,15 @@ func parseIPEnvironment(envName, envValue string, version int) *net.IP {
 
 // validateIP checks that the IP address is actually on one of the host
 // interfaces and warns if not.
-func validateIP(ip *net.IP) {
+func validateIP(ipn *net.IPNet) {
```
robbrockbank: Could argue both ways.  The field in the API is just called IPAddress rather than CIDR *and* we are actually only validating the IP address (we aren't validating the network part).

Given the latter point, I think it's ok as is.

calico_node/startup/startup.go
```diff
@@ -282,12 +289,15 @@ func parseIPEnvironment(envName, envValue string, version int) *net.IP {
...
 	// No validation required if no IP address is specified.
-	if ip == nil {
+	if ipn == nil {
 		return
 	}
 
+	// Pull out the IP as a net.IP (it has useful string and version methods).
+	ip := net.IP{ipn.IP}
```
robbrockbank: Alas no.  `ipn` is one of our wrapped IPNet objects which still only contains the golang.net IP and Mask components.

Another reason the wrapping approach is so stinky.

caseydavenport: bah, right.

caseydavenport:
@robbrockbank got some conflicts. 

Are you planning on doing STs in a separate PR?

PR #1458, opened 2017-01-23
Makefile: Remove reference to python build container

tomdee:


caseydavenport:
LGTM.

Any ideas what's up with the tests? They don't seem related.

PR #1461, opened 2017-01-25
Makefiles: Split calicoctl and calico-node into their own Makefiles

tomdee:


caseydavenport:
What do you think about moving the `Makefile.calico-node` file into the `calico_node` directory?

caseydavenport:
I think you will also need README.md updates as well to indicate that the make targets are now in different files, right?

PR #1465, opened 2017-01-26
Makefile: Use the golang-dp-driver branch of felix

tomdee:


tomdee:
We almost have a 2.1.0 RC of felix merged so closing

PR #1473, opened 2017-01-30
Fix rc.local in calico-node due to docker overlayfs bug when using Ce…

alexey-medvedchikov:
…ntOS 7 kernels

Due to bug in docker (https://github.com/docker/docker/issues/15314) we cant rmdir
any dirs on underlying layers. Fix it by copying service files first and then rm'ming
log dirs.


marvin-tigera:
Can one of the admins verify this patch?

tomdee:
@alexey-medvedchikov Thanks for this contribution. Could you put a comment in the rc.local file briefly explaining the problem to remind developers not to break it in future.

PR #1449, opened 2017-01-10
Smoother release process

gunjan5:
Fixes #1401 

- [x] `make release` "what to do next" help doesn't mention mac/windows calicoctl binaries.
- [x] `make release` "what to do next" help doesn't mention the list of sub-component versions to embed in the release note or how to collect that.
- [x] RELEASING.md implies you can release from a non-master branch, but that messes up the automatic version numbers assigned by subsequent releases (if the branch isn't merged back to master before the next release).
- [x] What are the pre-reqs for being able to push to quay and dockerhub?
- [ ] `make clean` deletes the `calico/felix` docker image; this caught me out when I was trying to do an end-to-end build on my system without pushing anything.  Feels wrong to delete an artefact that you didn't `make`.
- [x] If build fails have to delete the tag again before rebuild.
- [x] Final check of version number detected a dirty build and failed the build.  Please can we add a dirty check at the beginning?

PR #1474, opened 2017-01-30
Smoother release process take 2

### LINKED ISSUE #1401: Release process hiccups
First time I've tried to do a calico-containers release today; hit a few snags as a set of fresh eyes:

- [x] `make release` "what to do next" help doesn't mention mac/windows calicoctl binaries.
- [x] `make release` "what to do next" help doesn't mention the list of sub-component versions to embed in the release note or how to collect that.
- [x] RELEASING.md implies you can release from a non-master branch, but that messes up the automatic version numbers assigned by subsequent releases (if the branch isn't merged back to master before the next release).
- [x] What are the pre-reqs for being able to push to quay and dockerhub?
- [ ] `make clean` deletes the `calico/felix` docker image; this caught me out when I was trying to do an end-to-end build on my system without pushing anything.  Feels wrong to delete an artefact that you didn't `make`.
- [x] If build fails have to delete the tag again before rebuild.
- [x] Final check of version number detected a dirty build and failed the build.  Please can we add a dirty check at the beginning?

gunjan5:
Fixes #1401

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ ifndef VERSION
 endif
 	git tag $(VERSION)
 
+	# Check to make sure the tag isn't "-dirty".
+	if git describe --tags --dirty | grep dirty; \
+	then echo current git working tree is "dirty". Make sure you do not have any uncommitted changes ;false; fi
+
 	# Build the calicoctl binaries, as well as the calico/ctl and calico/node images.
 	CALICOCTL_NODE_VERSION=$(VERSION) $(MAKE) dist/calicoctl dist/calicoctl-darwin-amd64 dist/calicoctl-windows-amd64.exe 
 	CALICOCTL_NODE_VERSION=$(VERSION) $(MAKE) calico/ctl calico/node
@@ -68,7 +72,12 @@ endif
 	# Check that the version output includes the version specified.
 	# Tests that the "git tag" makes it into the binaries. Main point is to catch "-dirty" builds
 	# Release is currently supported on darwin / linux only.
-	if ! docker run $(CTL_CONTAINER_NAME) version | grep 'Version:\s*$(VERSION)$$'; then echo "Reported version:" `docker run $(CTL_CONTAINER_NAME) version` "\nExpected version: $(VERSION)"; false; else echo "Version check passed\n"; fi
+	if ! docker run $(CTL_CONTAINER_NAME) version | grep 'Version:\s*$(VERSION)$$'; then \
+	  echo "Reported version:" `docker run $(CTL_CONTAINER_NAME) version` "\nExpected version: $(VERSION)"; \
+	  false; \
+	else \
+	  echo "Version check passed\n"; \
+	fi
 
 	# Retag images with corect version and quay
 	docker tag $(NODE_CONTAINER_NAME) $(NODE_CONTAINER_NAME):$(VERSION)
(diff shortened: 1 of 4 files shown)


gunjan5:
Moved over the changes from PR #1449

Makefile.calicoctl
```diff
@@ -7,6 +7,12 @@
 # Determine which OS / ARCH.
 OS := $(shell uname -s | tr A-Z a-z)
 ARCH := amd64
+###############################################################################
+# Subcomponent versions:
+GO_BUILD_VER:=latest
+# Set libcalico-go version in glide.yaml first then do a `glide up -v` to update glide.lock.
+LIBCALICOGO_VER := v1.0.0
```
tmjd: Is LIBCALICOGO_VER suppose to match up with the version glide.yaml is updated to? If so, the comment should probably state that explicitly.  I also wanted to check that LIBCALICOGO_VER is not suppose to be used anywhere in the Makefiles besides just to be printed out.

tomdee:
FWIW this looks fine to me. I don't think it's making any user visible changes, it's just doing a better job of documenting the process for ourselves, right?

Makefile
```diff
@@ -61,14 +61,20 @@ ifndef VERSION
...
+
 	# Build the calicoctl binaries, as well as the calico/ctl and calico/node images.
 	CALICOCTL_NODE_VERSION=$(VERSION) $(MAKE) dist/calicoctl dist/calicoctl-darwin-amd64 dist/calicoctl-windows-amd64.exe 
 	CALICOCTL_NODE_VERSION=$(VERSION) $(MAKE) calico/ctl calico/node
 
 	# Check that the version output includes the version specified.
 	# Tests that the "git tag" makes it into the binaries. Main point is to catch "-dirty" builds
 	# Release is currently supported on darwin / linux only.
-	if ! docker run $(CTL_CONTAINER_NAME) version | grep 'Version:\s*$(VERSION)$$'; then echo "Reported version:" `docker run $(CTL_CONTAINER_NAME) version` "\nExpected version: $(VERSION)"; false; else echo "Version check passed\n"; fi
+	if ! docker run $(CTL_CONTAINER_NAME) version | grep 'Version:\s*$(VERSION)$$'; \
```
fasaxc: Nit, I'd find it easier to read this if it was wrapped like this:

```
if ! docker run $(CTL_CONTAINER_NAME) version | grep 'Version:\s*$(VERSION)$$'; then \
  echo "Reported version:" `docker run $(CTL_CONTAINER_NAME) version` "\nExpected version: $(VERSION)"; \
  false; \
else \
  echo "Version check passed\n"; \
fi
```

Makefile
```diff
@@ -88,7 +94,19 @@ endif
 	docker run $(NODE_CONTAINER_NAME) calico-felix --version
 	docker run $(NODE_CONTAINER_NAME) libnetwork-plugin -v
 
-	@echo "\nNow push the tag and images. Then create a release on Github and attach the dist/calicoctl binary"
+	@echo "\nNow push the tag and images. Then create a release on Github and \
```
fasaxc: I think it'd be cleaner to just use another echo here rather than a line-continuation followed by a `\n`, especially when that's what we do for the lines below:
```
@echo foo
@echo bar
```
Instead of 
```
@echo foo\
\nbar
```

Makefile.calico-node
```diff
@@ -1,22 +1,30 @@
...
-SYSTEMTEST_CONTAINER?=calico/test
+SYSTEMTEST_CONTAINER?=calico/test:$(SYSTEMTEST_CONTAINER_VER)
 
 # calicoctl and calico/node current share a single version - this is it.
 CALICOCONTAINERS_VERSION?=$(shell git describe --tags --dirty --always)
+###############################################################################
+# Subcomponent versions:
+BIRD_VER := v0.2.0
+GOBGPD_VER := v0.1.1
+FELIX_VER := 2.0.1-rc1
```
fasaxc: I think Casey just shipped a 2.0.2 so this should be updated.

gunjan5: ah good point!

Makefile.calicoctl
gunjan5: This is only for printing the value. `RELEASING.md` mentions what needs to be done before updating the Makefile. But, I see your point. I have updated the comment to be more explicit.

gunjan5:
@fasaxc I've made those changes. PTAL

PR #1484, opened 2017-02-01
Add some ipam tests

lwr20:


tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+                    }
+        self.hosts[0].writefile("pools.yaml", yaml.dump(new_pool))
+        self.hosts[0].calicoctl("create -f pools.yaml")
+
+        for host in self.hosts:
+            workload = host.create_workload("wlda2-%s" % host.name,
+                                            image="workload",
+                                            network=self.network)
+            assert netaddr.IPAddress(workload.ip) in ipv4_subnet, \
+                "Workload IP in wrong pool. IP: %s, Pool: %s" % (workload.ip, ipv4_subnet.ipv4())
```
fasaxc: As discussed, would be great to have a check that, after removing a pool and all pods in the pool, no stale routes are present.

tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Tidy up workloads
+        for workload, host in workloads:
+            self.delete_workload(host, workload)
+        # Delete new pool
+        self.hosts[0].calicoctl("delete -f newpool.yaml")
+
+    @parameterized.expand([
+        # Can't use boolean values here :(
+        "False",
```
fasaxc: Do bools work in a 1-tuple: `(True,)`  (note the trailing comma)

tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+        "True",
+    ])
+    def test_pool_wrap(self, make_static_workload_str):
+        """
+        Repeatedly create and delete workloads until the system re-assigns an IP.
+        """
+        unittest.skip("remove me")
```
fasaxc: skip left in?

tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+                                                              "still in use!"
+
+            if new_workload.ip == original_ip:
+                # We assign pools to hosts in /26's - so 64 addresses.
+                poolsize = 64
+                # But if we're using one for a static workload, there will be one less
+                if make_static_workload:
+                    poolsize -= 1
+                assert i >= poolsize, "Original IP was re-assigned before entire host pool " \
+                                      "was cycled through.  Hit after %s times" % i
```
fasaxc: Does the IPAM assignment algorithm guarantee that it waits as long as possible before re-use?  I think it might be best-effort, might be best to check with Rob, say.

tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+                break
+            if i > (len(ipv4_subnet) * 2):
+                assert False, "Cycled twice through pool - original IP still not assigned."
+        # Clear out any lingering workloads
+        self.delete_workload(host, new_workload)
```
fasaxc: This clean-up won't be executed if the asserts fail, you should do cleanup in tearDown

tests/st/ipam/test_ipam.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Clear up static workload
+        if make_static_workload:
+            self.delete_workload(host, static_workload)
+
+    @staticmethod
+    def create_network(host, net_name):
```
fasaxc: Feels like a good utility method to move to a base class

tests/st/ipam/test_ipam.py
lwr20: @robbrockbank ^?

lwr20: I'm tempted to say we should just check this in because it passes (reliably AFAICT) at the moment.  If it fails later, we can have the usual chat about whether its a 'real' bug or not.

tests/st/ipam/test_ipam.py
lwr20: I've added a check for the blackhole route.  Is that sufficient?

fasaxc: How about asserting that only the expected /32s are present?

lwr20: Added check for workload routes.  Howzat?

PR #1485, opened 2017-02-01

robbrockbank:
glibc.

Change-Id: I290c93b5b8dedf6a621e9ee846216ac2ae8d34f1


fasaxc:
@robbrockbank is this ready for review?

robbrockbank:
Discuess with Tom, we think there is a better approach to getting libc in to the container.  I'll make a new PR.

PR #1490, opened 2017-02-03
Tag quay latest after build

robbrockbank:
Looks as though the default node image was updated to use quay, but we weren't tagging our calico/node image as quay, nor tarring up the quay image for use with the STs.

Upshot of this is that the STs will be downloading quay.io/calico/node:latest each time we start calico node.  Explains why the STs were taking so damn long.  It might also explain why #1488 is working ... I think that will need to be rebased off this once it is in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,13 +79,11 @@ endif
 	  echo "Version check passed\n"; \
 	fi
 
-	# Retag images with corect version and quay
+	# Retag images with correct version and quay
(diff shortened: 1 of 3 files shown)


tomdee:
Hmmm. Lemme think about this. I'm not sure we should be using quay here

robbrockbank:
@tomdee - yup no probs.  I guess another approach might be to change the default node image for the "latest" build

tomdee:
Discussed with @robbrockbank and LGTM

PR #1499, opened 2017-02-06
Remove quay.io tagging

artem-panchenko:
After #1490 was merged you are always forced to tag calico/node and calico/ctl images with quay.io, even if you just build stuff locally and not going to upload it somewhere. That change also may confuse if you   upload calico images to your private registry instead of quay.io or docker hub.

I propose to revert changes from #1490 and add an ability of image name specifying for the system tests.

### DIFF
--- a/Makefile.calico-node
+++ b/Makefile.calico-node
@@ -57,7 +57,7 @@ calico_test.created: $(TEST_CONTAINER_FILES)
 calico/node: $(NODE_CONTAINER_CREATED)    ## Create the calico/node image
 
 calico-node.tar: $(NODE_CONTAINER_CREATED)
-	docker save --output $@ quay.io/$(NODE_CONTAINER_NAME):latest
+	docker save --output $@ $(NODE_CONTAINER_NAME)
 
 # Build ACI (the APPC image file format) of calico/node.
 # Requires docker2aci installed on host: https://github.com/appc/docker2aci
@@ -67,7 +67,6 @@ calico-node-latest.aci: calico-node.tar
 # Build calico/node docker image - explicitly depend on the container binaries.
 $(NODE_CONTAINER_CREATED): $(NODE_CONTAINER_DIR)/Dockerfile $(NODE_CONTAINER_FILES) $(addprefix $(NODE_CONTAINER_BIN_DIR)/,$(NODE_CONTAINER_BINARIES))
 	docker build -t $(NODE_CONTAINER_NAME) $(NODE_CONTAINER_DIR)
-	docker tag $(NODE_CONTAINER_NAME) quay.io/$(NODE_CONTAINER_NAME):latest
 	touch $@
 
 # Get felix binaries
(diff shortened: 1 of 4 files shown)


robbrockbank:
Hi @artem-panchenko 

We were in two minds as to whether to do it this way or with the re-tagging.  I was concerned it might be easy to not set the node image on certain tests (if, say the test was explicitly running calicoctl node run without using the helper command), but it's more of a worry - I don't have an explicit situation at the moment where I have done that.

That said, I'm happy to merge - I have one comment that I think should be addressed before we merge though.

Rob

Makefile.calico-node
```diff
@@ -256,6 +255,7 @@ st: dist/calicoctl busybox.tar routereflector.tar calico-node.tar workload.tar r
 	           -e HOST_CHECKOUT_DIR=$(HOST_CHECKOUT_DIR) \
 	           -e DEBUG_FAILURES=$(DEBUG_FAILURES) \
 	           -e MY_IP=$(LOCAL_IP_ENV) \
+                   -e NODE_CONTAINER_NAME=$(NODE_CONTAINER_NAME) \
```
robbrockbank: The line spacing seems wonky compared to the other lines

Makefile.calico-node
```diff
@@ -283,6 +283,7 @@ st-ssl: run-etcd-ssl dist/calicoctl busybox.tar calico-node.tar routereflector.t
...
```
robbrockbank: Ditto

tests/st/utils/docker_host.py
```diff
@@ -216,7 +217,7 @@ def start_calico_node(self, options=""):
         Start calico in a container inside a host by calling through to the
         calicoctl node command.
         """
-        args = ['node', 'run']
+        args = ['node', 'run', '--node-image=%s' % NODE_CONTAINER_NAME]
```
robbrockbank: I think this option should only be added if it isn't supplied in the options string.   Maybe just do a `if "--node-image" not in options`.

robbrockbank: Also, there is a method `start_calico_node_with_docker` that should also be updated to use this container image.  We don't actually use it at the moment, but we should at least keep it updated so that we don't potentially run into problems if we do start using it.

robbrockbank:
@artem-panchenko Actually, I'm going to merge as is.  I'll make the minor tweaks myself as I'm making changes in this area too

robbrockbank:
@artem-panchenko ... After chatting about this, I've just remembered my rationale behind the original change :-/

Both calicoctl, and a number of our k8s manifests etc. reference the quay.io calico/node and calico/ctl images.  The upshot of this is if you are doing local builds and local testing (not just our STs), it is necessary to remember to either specify the node image on calicoctl or re-write your manifests.  This makes the dev/test cycle prone to errors.  It seems significantly safer to just ensure the calico/node and quay.io/calico/node images are kept in-sync at the source (i.e. when the image is created).

I'm currently thinking that I might actually re-instate this change - although I'm happy to entertain other suggestions for improving our builds.

artem-panchenko:
@robbrockbank Thank you for the explanation! I thought the problem is that it's not possible to use proper images from tests, so tried to fix only that issue. I also understand your point about development process specific stuff, but if you ask my opinion, I'd rather to force users (or devs) to always manually specify essential settings than implicitly set them to nonintuitive values. That's how we do when we deploy k8s by kargo :)

Anyway,  please don't waste your time on this if you think quay.io tagging after each build is important for you and team. Let's just re-revert your change then!

PR #1500, opened 2017-02-06
Add IP autodetection methods:  can-reach and interface

### LINKED ISSUE #873: Improve calico/node IP address detection

robbrockbank:
Fixes #873

Documentation covered by https://github.com/projectcalico/calico/issues/494

### DIFF
--- a/calico_node/filesystem/etc/rc.local
+++ b/calico_node/filesystem/etc/rc.local
@@ -74,7 +74,7 @@ esac
 # If running libnetwork plugin in a separate container, CALICO_LIBNETWORK_ENABLED would be false.
 # CALICO_LIBNETWORK_ENABLED is "false" by default. It can be set by passing `--libnetwork` flag while starting the calico/node via calicoctl
 if [ "$CALICO_LIBNETWORK_ENABLED" == "true" ]; then
-	echo "CALICO_LIBNETWORK_ENABLED is true - start libnetwork service"
+	echo "Starting libnetwork service"
 	cp -a /etc/service/available/libnetwork  /etc/service/enabled/
 fi
 
--- a/calico_node/startup/autodetection/reachaddr.go
+++ b/calico_node/startup/autodetection/reachaddr.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package autodetection
+
+import (
+	"fmt"
+	gonet "net"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/projectcalico/libcalico-go/lib/net"
+)
+
+// ReachDestination auto-detects the interface Network by setting up a UDP
+// connection to a "reach" destination.
+func ReachDestination(dest string, version int) (*net.IPNet, error) {
+	log.Debugf("Auto-detecting IPv%d CIDR by reaching destination %s", version, dest)
+
+	// Open a UDP connection to determine which external IP address is
+	// used to access the supplied destination.
+	protocol := fmt.Sprintf("udp%d", version)
+	address := fmt.Sprintf("[%s]:80", dest)
+	conn, err := gonet.Dial(protocol, address)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	// Get the local address as a golang IP and use that to find the matching
+	// interface CIDR.
+	addr := conn.LocalAddr()
+	if addr == nil {
+		return nil, fmt.Errorf("no address detected by connecting to %s", dest)
+	}
+	udpAddr := addr.(*gonet.UDPAddr)
+	log.WithFields(log.Fields{"IP": udpAddr.IP, "Destination": dest}).Info("Auto-detected address by connecting to remote")
+
+	// Get a full list of interface and IPs and find the CIDR matching the
+	// found IP.
+	ifaces, err := GetInterfaces(nil, nil, version)
+	if err != nil {
+		return nil, err
+	}
+	for _, iface := range ifaces {
+		log.WithField("Name", iface.Name).Info("Checking interface CIDRs")
+		for _, cidr := range iface.Cidrs {
+			log.WithField("CIDR", cidr.String()).Info("Checking CIDR")
+			if cidr.IP.Equal(udpAddr.IP) {
+				log.WithField("CIDR", cidr.String()).Info("Found matching interface CIDR")
+				return &cidr, nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("autodetected IPv%d address does not match any addresses found on local interfaces: %s", version, udpAddr.IP.String())
+}
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -32,8 +32,11 @@ import (
 )
 
 const (
-	DEFAULT_IPV4_POOL_CIDR = "192.168.0.0/16"
-	DEFAULT_IPV6_POOL_CIDR = "fd80:24e2:f998:72d6::/64"
+	DEFAULT_IPV4_POOL_CIDR      = "192.168.0.0/16"
+	DEFAULT_IPV6_POOL_CIDR      = "fd80:24e2:f998:72d6::/64"
+	AUTODETECT_METHOD_FIRST     = "first-found"
+	AUTODETECT_METHOD_CAN_REACH = "can-reach="
+	AUTODETECT_METHOD_INTERFACE = "interface="
 )
 
 // For testing purposes we define an exit function that we can override.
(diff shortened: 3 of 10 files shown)


calico_node/filesystem/etc/rc.local
```diff
@@ -74,8 +74,10 @@ esac
 # If running libnetwork plugin in a separate container, CALICO_LIBNETWORK_ENABLED would be false.
 # CALICO_LIBNETWORK_ENABLED is "false" by default. It can be set by passing `--libnetwork` flag while starting the calico/node via calicoctl
 if [ "$CALICO_LIBNETWORK_ENABLED" == "true" ]; then
-	echo "CALICO_LIBNETWORK_ENABLED is true - start libnetwork service"
-	cp -a /etc/service/available/libnetwork  /etc/service/enabled/
+	echo "Starting libnetwork service"
+	mv /etc/service/available/libnetwork  /etc/service/enabled/
+else
+	echo "Libnetwork service is not required"
```
gunjan5: I don't understand why do we need to print this message to the user. (Also, I don't understand what it means by "libnetwork service not required")

calico_node/startup/autodetection/reachaddr.go
```diff
@@ -0,0 +1,50 @@
+package autodetection
+
+import (
+	"fmt"
+	gonet "net"
```
gunjan5: nit: not a fan of renaming Go's native libraries. We implemented our "net" differently so our `/lib/net` should be imported as `cnet` or something. Most people reading the code will think of Go's native `net` when they see `net`

calico_node/startup/autodetection/reachaddr.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	// Get the local address as a golang IP and use that to find the matching
+	// interface CIDR.
+	addr := conn.LocalAddr()
+	if addr == nil {
+		return nil, fmt.Errorf("no address detected by connecting to %s", dest)
```
gunjan5: I think more friendly error message could be something like `"no addresses found matching %s interface CIDR."`. WDYT?

calico_node/startup/autodetection/reachaddr.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	udpAddr := addr.(*gonet.UDPAddr)
+	log.WithFields(log.Fields{"IP": udpAddr.IP, "Destination": dest}).Info("Auto-detected address by connecting to remote")
+
+	// Get a full list of interface and IPs and find the CIDR matching the
+	// found IP.
+	ifaces, err := GetInterfaces(nil, nil, version)
```
gunjan5: `err` var is declared but not used.

calico_node/startup/startup.go
```diff
@@ -34,6 +34,9 @@ import (
 const (
 	DEFAULT_IPV4_POOL_CIDR = "192.168.0.0/16"
 	DEFAULT_IPV6_POOL_CIDR = "fd80:24e2:f998:72d6::/64"
+	AUTOD_METHOD_FIRST     = "first-found"
```
gunjan5: hmm `AUTOD_` doesn't sound very intuitive. Can it be something like `AUTODETECT_` or `IFACE_DETECT_` or something similar?

calico_node/startup/startup.go
```diff
@@ -312,35 +341,75 @@ func validateIP(ip *net.IP) {
...
+}
+
+// autoDetectCIDRFirstFound auto-detects the first valid Network it finds across
+// all interfaces (excluding common known internal interface names).
+func autoDetectCIDRFirstFound(version int) *net.IP {
+
 	incl := []string{}
-	excl := []string{"^docker.*", "^cbr.*", "dummy.*",
+	excl := []string{
+		"docker.*", "cbr.*", "dummy.*",
```
gunjan5: curious why you removed `^` from the regex. Wondering someone with some modification to docker0 bridge can have an interface like `...-docker...` and if this will change the behavior we use to have.

calico_node/startup/startup.go
```diff
@@ -312,35 +341,75 @@ func validateIP(ip *net.IP) {
...
-		message("Unable to auto-detect an IPv%d address", version)
+	message("Using autodetected IPv%d address on interface %s: %s", version, iface.Name, cidr.String())
+
+	return &net.IP{cidr.IP}
+}
+
+// autoDetectCIDRByInterface auto-detects the first valid Network on the interfaces
+// matching the supplied interface regex.
+func autoDetectCIDRByInterface(ifaceRegex string, version int) *net.IP {
+	iface, cidr, err := autodetection.FilteredEnumeration([]string{ifaceRegex}, nil, version)
```
gunjan5: I wonder if passing in `nil` here is the right thing to do. It will work in this case since you're checking `len()` of it and it will return `0` but someone else might try to do something with it in future and they might get a nil pointer exception. How about passing  and empty slice of string `[]string{}`?

calico_node/startup/startup.go
```diff
@@ -312,35 +341,75 @@ func validateIP(ip *net.IP) {
...
 	return &net.IP{cidr.IP}
 }
 
+// autoDetectCIDRByReach auto-detects the IP and Network by setting up a UDP
+// connection to a "reach" address.
+func autoDetectCIDRByReach(dest string, version int) *net.IP {
+	if cidr, err := autodetection.ReachDestination(dest, version); err != nil {
+		warning("Unable to auto-detect IPv%d address by connecting to %s: %s", version, dest, err)
+		return nil
+	} else {
```
gunjan5: don't need the `else` here if we're returning in `if`

calicoctl/commands/node/run.go
```diff
@@ -85,6 +92,29 @@ Options:
                            node resource.  If there is no configured value
                            and the --ip6 option is omitted, the node will not
                            route IPv6.
+     --ip-autodetection-method=<IP_AUTODETECTION_METHOD>
+                           Specify the autodetection method for detecting the
+                           local IPv6 routing address for this node.  The valid
+                           options are:
+                           > first-found
+                             Use the first valid IP address on the first
+                             enumerated interface (filtering out common known
```
gunjan5: `filtering out` -> `filters out`?

gunjan5:
Some comments + need a docs PR to go with this :)

calico_node/startup/autodetection/reachaddr.go
robbrockbank: that error message would not be correct.

calico_node/filesystem/etc/rc.local
robbrockbank: not required in that it has been explicitly disabled.   I can remove... i was adding it just to be nice... but i guess omission of a message is just as good.

calico_node/startup/autodetection/reachaddr.go
robbrockbank: I don't buy the latter point - there is always a risk that if a library is named the same thing as golang native library people will assume it's the native library.  Say I didn't need go-net in here, I wouldn't need to rename the cnet one so it'd just be net...

Personally, I like the idea of always prefixing the native ones with go.  It makes it clearer that there is an identically named package *and* it is an easy naming convention to stick to.

That said, I'm not too fussed either way.  Ping me if it's really irritating you.

calico_node/startup/autodetection/reachaddr.go
robbrockbank: that error message would not be correct though.

calico_node/startup/startup.go
robbrockbank: we do a regex match, so it was superfluous

calico_node/startup/startup.go
robbrockbank: I think nil should be fine.  A nil slice can be used with both len() and range....   I could just update the function def to state that nil is valid.

calico_node/startup/startup.go
robbrockbank: Not really.

Sometimes it makes the code look a little more "symmetric".  it'll compile to the same thing I would hope.

calicoctl/commands/node/run.go
robbrockbank: no.   perhaps "common known exceptions such as the docker bridge are filtered out"

calico_node/startup/startup.go
robbrockbank: Yes you do because cidr is only valid within the scope of the if clause.

robbrockbank: Of course I could restructure to define cidr outside the clause, but meh.

calicoctl/commands/node/run.go
```diff
@@ -32,12 +33,16 @@ import (
...
 	"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
 	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/net"
 )
 
 const (
 	ETCD_KEY_NODE_FILE     = "/etc/calico/certs/key.pem"
 	ETCD_CERT_NODE_FILE    = "/etc/calico/certs/cert.crt"
 	ETCD_CA_CERT_NODE_FILE = "/etc/calico/certs/ca_cert.crt"
+	AUTOD_METHOD_FIRST     = "first-found"
```
gunjan5: this is still `AUTOD_...`

calicoctl/commands/node/run.go
```diff
@@ -85,6 +92,29 @@ Options:
                            node resource.  If there is no configured value
                            and the --ip6 option is omitted, the node will not
                            route IPv6.
+     --ip-autodetection-method=<IP_AUTODETECTION_METHOD>
+                           Specify the autodetection method for detecting the
+                           local IPv6 routing address for this node.  The valid
```
djosborne: should be `IPv4`

calicoctl/commands/node/run.go
```diff
@@ -85,6 +92,29 @@ Options:
...
+                           options are:
+                           > first-found
+                             Use the first valid IP address on the first
+                             enumerated interface (common known exceptions are
+                             filtered out, e.g. the docker bridge).  It is not
+                             receommended to use this if you have multiple
+                             external interfaces on your host.
+                           > can-reach=<IP OR DNS>
```
djosborne: So based on this description, I would assume the syntax for can-reach is `calicoctl node run --ip-autodetection-method=can-reach=10.0.0.1`. Am I correct?

calicoctl/commands/node/run.go
```diff
@@ -85,6 +92,29 @@ Options:
...
+                             Use the interface determined by your host routing
+                             tables that will be used to reach the supplied
+                             destination IP or DNS.
+                           > interface=<IFACE NAME REGEX>
```
djosborne: I'm concerned how ba/sh would handl regex strings here. Would `$` and `*` for example need to be escaped?

calicoctl/commands/node/run.go
djosborne: If I am correct, It's good that I was able to deduce the correct syntax by reading the docstring. I worry that it is a bit un-bashlike, but can't see any issues beyond that. Might be useful to put my example stringin the docstring itself to help users.

calicoctl/commands/node/run.go
robbrockbank: yes you'll almost certainly have to escape some stuff for BASH - but I don't think that should matter here.

calicoctl/commands/node/run.go
robbrockbank: It's more bash like if you don't use a "=" after the parameter, e.g.:

`--ip-autodetection-method can-reach=10.0.0.1`

It's similar to the --output method where you can specify parameters on some options.   I believe k8s follows the same approach.

Personally, I think this is OK.

robbrockbank: Oh, and yes, you are correct with the syntax

djosborne: Ah yes that does look like the good ole `-e ETCD_ENDPOINTS=10.0.0.1` for docker.

calicoctl/commands/node/run.go
```diff
@@ -85,6 +92,29 @@ Options:
...
+                           > first-found
+                             Use the first valid IP address on the first
+                             enumerated interface (common known exceptions are
+                             filtered out, e.g. the docker bridge).  It is not
+                             receommended to use this if you have multiple
+                             external interfaces on your host.
+                           > can-reach=<IP OR DNS>
+                             Use the interface determined by your host routing
+                             tables that will be used to reach the supplied
+                             destination IP or DNS.
```
caseydavenport: DNS _record_ ?

calicoctl/commands/node/run.go
caseydavenport: I think this is fine.  I'm a bit concerned that this doesn't solve the problem with the k8s self-hosted manifests w.r.t always choosing the right interface without the user making any changes, since there's no single IP that will work for all deployments.

It's definitely a lot better than it is today - at least users will have a knob to turn.

calicoctl/commands/node/run.go
djosborne: I'd prefer just "hostname"

PR #1503, opened 2017-02-07

artem-panchenko:
glibc.

Change-Id: I290c93b5b8dedf6a621e9ee846216ac2ae8d34f1


tomdee:
LGTM

PR #1506, opened 2017-02-08
Fix WAIT_FOR_DATASTORE so it breaks when connected.

### LINKED ISSUE #1505: WAIT_FOR_DATASTORE broken
Looks like the WAIT_FOR_DATASTORE loop has no exit condition. 

This code: https://github.com/projectcalico/calicoctl/blob/23939adccbd0340787098b558641c615210dd510/calico_node/startup/startup.go#L152

should probably have a break in it.

caseydavenport:
Fixes #1505


robbrockbank:
LGTM once the tests pass.

PR #1511, opened 2017-02-09
Adding env passthrough

### LINKED ISSUE #1501: Calicoctl should pass through some environment variables automatically when starting node
IIUC, `calicoctl node run` already passes through environment variables prefixed with `FELIX_` when running the `calico/node` image. Libnetwork-plugin uses environment variables prefixed with `CALICO_` so `calicoctl` should pass those through too.

### LINKED ISSUE #1597: Don't populate IP / AS in policy-only mode
When using policy-only mode (CALICO_NETWORKING_BACKEND=none) we don't need to populate the IP / AS on the Node resource.

heschlie:
Adding the ability to pass env vars through `calicoctl node run` that are prefixed with `CALICO_*` or FELIX_*`.

Fixes #1501
Fixes #1597

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -111,12 +111,10 @@ Options:
                            > interface=<IFACE NAME REGEX>
                              Use the first valid IP address found on interfaces
                              named as per the supplied interface name regex.
-                           [default: first-found]
      --ip6-autodetection-method=<IP6_AUTODETECTION_METHOD>
                            Specify the autodetection method for detecting the
                            local IPv6 routing address for this node.  See
                            ip-autodetection-method flag for valid options.
-                           [default: first-found]
      --log-dir=<LOG_DIR>   The directory containing Calico logs.
                            [default: /var/log/calico]
      --node-image=<DOCKER_IMAGE_NAME>
@@ -216,6 +214,20 @@ Description:
 			ifprefix = env
 		} else {
 			ifprefix = "cali"
+		}
+	}
+	if ipv4ADMethod == "" {
+		if env := os.Getenv("CALICO_IP_AUTODETECTION_METHOD"); env != "" {
+			ipv4ADMethod = env
+		} else {
+			ipv4ADMethod = "first-found"
+		}
+	}
+	if ipv6ADMethod == "" {
+		if env := os.Getenv("CALICO_IP6_AUTODETECTION_METHOD"); env != "" {
+			ipv6ADMethod = env
+		} else {
+			ipv6ADMethod = "first-found"
 		}
 	}
 
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -287,14 +287,16 @@ Description:
 	}
 
 	// Add in optional environments.
-	if asNumber != "" {
-		envs["AS"] = asNumber
-	}
-	if ipv4 != "" {
-		envs["IP"] = ipv4
-	}
-	if ipv6 != "" {
-		envs["IP6"] = ipv6
+	if strings.ToLower(backend) != "none" {
+		if asNumber != "" {
+			envs["AS"] = asNumber
+		}
+		if ipv4 != "" {
+			envs["IP"] = ipv4
+		}
+		if ipv6 != "" {
+			envs["IP6"] = ipv6
+		}
 	}
 
 	// Create a struct for volumes to mount.
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -127,7 +127,6 @@ Options:
                            to "none", Calico node runs in policy only mode.
                            The option to run with gobgp is currently
                            experimental.
-                           [default: bird]
      --dryrun              Output the appropriate command, without starting the
                            container.
      --init-system         Run the appropriate command to use with an init
@@ -141,7 +140,6 @@ Options:
                            Interface prefix to use for the network interface
                            within the Docker containers that have been networked
                            by the Calico driver.
-                           [default: cali]
      --use-docker-networking-container-labels
                            Extract the Calico-namespaced Docker container labels
                            (org.projectcalico.label.*) and apply them to the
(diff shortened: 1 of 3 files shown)


heschlie:
@lwr20 Do you think we need more interesting tests for this?  The one I made sets some dummy env vars and verifies they make it into the container but I'm having trouble thinking of a clever way to stress this more if we need it.

tmjd:
I think a good test would be to test that the flags take precedence over env vars.

tmjd:
If you moved the loop for transferring the env vars to be the first thing put into the envs map then you wouldn't need the check if they already exist.

tests/st/calicoctl/test_node_run.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        os.environ['CALICO_TESTVAL'] = 'dummyVal'
+        os.environ['FELIX_TESTVAL'] = 'felixDummyVal'
+
+        with DockerHost('host', dind=False) as host:
+            calico_out = host.execute('printenv CALICO_TESTVAL')
+            felix_out = host.execute('printenv FELIX_TESTVAL')
+
+            self.assertEquals(calico_out, 'dummyVal')
+            self.assertEquals(felix_out, 'felixDummyVal')
```
heschlie: You are right I do want to be checking the calico-node, I had thought that the DockerHost was the Node container.  Thanks for pointing that out.

lwr20: You can tell the Dockerhost to start calico-node inside it by simply passing it the argument `start_calico=True` when you create it.

djosborne:
Fixes #1501

heschlie:
Looking over this in more detail now, the `CALICO_*` will never make it through if we want to prioritize CLI flags (which is definitely what we want) as we default the flags to some value when they are not set.  For example `calicoctl node run --backend=` defaults to `bird` so it always has a value that later gets assigned in the envs map that is passed to the docker command.

Could we instead, not have those set to a default, check for a blank string, then check for an env var, then setting a default if neither are set?  WDPT?  @gunjan5 @lwr20 @tmjd

tmjd:
That will work fine for the args that set a string, the boolean ones you'll have to figure something else out because the argutils.ArgBoolOrFalse will always be set in its current implimentation.  Unless there is a way to check if the arg was defaulted or not.

So that is :+1: to what you are proposing but I think there is a little more to switching things around.

heschlie:
Sorting out what's causing the tests to fail on Semaphore but not locally...

tomdee:
@heschlie What's happening with this PR?

heschlie:
@tomdee Sorry for the late reply, finally getting back to this in the last couple days, sorting out why the tests I wrote pass in `make test-containerized` and `make node-test-containerized` but not in `make semaphore`

heschlie:
@lwr20 Finally got the tests passing, could I get a review and LGTM if you think it's ready?  I'll squash after that.

tests/st/calicoctl/test_node_run.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class TestNodeRun(TestBase):
+    
+    def test_node_run_passes_env_vars(self):
+        """
+        Set some CALICO_ and FELIX_ env vars and see that they make it through to the
+        Node container
+        """
+        os.environ['CALICO_TESTVAL'] = 'dummyVal'
```
robbrockbank: Rather than setting and unsetting envs, it might be nicer to change the signature of DockerHost calicoctl() method to accept a list of environments ["X=Y", "A=B"] and then prepend the command with exported envs (such as we already do for the ETCD parms)

neiljerram:
Just commenting here so that I will be notified when this is merged.  Looking forward to it!

tests/st/calicoctl/test_node_run.py
robbrockbank: I think the approach of pre-pending the command with inline envs also means that we don't risk having dangling environments if the test bails out halfway through.

robbrockbank:
@heschlie  LGTM, but looks like there is a merge conflict - can you resolve please and I think we should be able to get this checked in.

heschlie:
@robbrockbank @neiljerram I've added arguments to the chain of commands down to executing `calicoctl` take a peek and let know WDYT.  

I've fixed the conflicts locally just running the tests again before I push it up.

heschlie:
Conflicts removed, tests passed locally, and sqaushed.  Should be ready for merge once Semaphore is green.

robbrockbank:
Hmmm, actually I do have a follow up question....

There are a bunch of defaulted values [default: blah] - which all basically map through to environment variables.  Why have we only changed a couple of them to change the inheritance order?

calicoctl/commands/node/run.go
```diff
@@ -262,6 +262,17 @@ Description:
 		vols[etcdcfg.EtcdCertFile] = ETCD_CERT_NODE_FILE
 	}
 
+	// Parse the environment variables passing through CALICO_* and FELIX_*
+	for _, env := range os.Environ() {
+		if strings.HasPrefix(env, "CALICO_") || strings.HasPrefix(env, "FELIX_") {
+			split_env := strings.Split(env, "=")
+			// Let's avoid clobbering anything already set
+			if _, ok := envs[split_env[0]]; !ok {
```
tomdee: @heschlie Did this get resolved? It looks like maybe it did?

calicoctl/commands/node/run.go
```diff
@@ -205,6 +203,21 @@ Description:
 		// the AS number, so convert.
 		asNumber = argutils.ValidateASNumber(asNumber).String()
 	}
+	// CLI flags will take precedence over env vars
```
tomdee: I don't understand this change, what is it for?

robbrockbank: It is to re-instate the defaults that were removed from the doc-string.

I think this is because docopt does not distinguish between a value that was explicitly set on the CLI and one that was set through the [default: XXX] option.

robbrockbank: I think the auto-detection methods should also be there now that I think about it.

calicoctl/commands/node/run.go
heschlie: It did in the else clause

calicoctl/commands/node/run.go
heschlie: @robbrockbank is correct I had to skip setting a default in the docstring so we could...

1. see if the user set a CLI flag
2. Check if there is an env var if no flag
3. Set a default if neither are present

heschlie: Rob, I can get those added in, I'm going to run through the reference list again as well to see if I missed anything.

heschlie:
@robbrockbank @tomdee This is ready for another review, I went ahead and pulled in the upstream changes to my PR and have updated based on your feedback.

neiljerram:
@heschlie @robbrockbank @tomdee Is this PR still planned to land sometime soon?  I'm interested so that I can finish off #1587 , which Rob has suggested should be based on this one.  Thanks!

heschlie:
@tomdee @robbrockbank I'm going to rebase this again today to get it caught up, could we get this reviewed and merged if possible?

caseydavenport:
I'm going to close this since it's pretty stale and I don't think it's a priority right now.

We can re-open/visit this in the future if the requirement arises.

PR #1512, opened 2017-02-10
Use FelixConfig in WAIT_FOR_DATASTORE

caseydavenport:
Changing the query we use for WAIT_FOR_DATASTORE so that it's compatible with the k8s datastore driver.

Didn't work with a Profile because the k8s datastore driver doesn't support ProfileTagsKey, just ProfileKey, and the `lib/backend/compat` code translates requests for `ProfileKey` into `ProfileTagsKey`


robbrockbank:
LGTM

PR #1513, opened 2017-02-10
Add Docker Networking Labeling flag to node run

tmjd:


calicoctl/commands/argutils/args.go
```diff
@@ -31,3 +31,12 @@ func ArgBoolOrFalse(args map[string]interface{}, argName string) bool {
 	}
 	return false
 }
+
+// ArgBoolOrTrue returns the requested argument as a boolean, or as true
+// if the argument is not present.
+func ArgBoolOrTrue(args map[string]interface{}, argName string) bool {
```
robbrockbank: Not sure we need this - a missing boolean flag implies False.

calicoctl/commands/node/run.go
```diff
@@ -141,6 +142,10 @@ Options:
                            within the Docker containers that have been networked
                            by the Calico driver.
                            [default: cali]
+     --enable-docker-networking-labeling
```
robbrockbank: I think the doc string here needs a little tweak.  This doesn't enable adding labels to Docker containers, it enables support for pulling Docker labels from the container and adding those labels to the Calico endpoints for use with Calico policy.

I think we could also say that only "calico.blah" (I can't remember the syntax) namespaced labels are transferred.

calicoctl/commands/node/run.go
```diff
@@ -141,6 +142,10 @@ Options:
...
+                           Enables adding labels to Docker containers that can
+                           be used in policies. Setting this flag also disables
+                           the default network profiles.
```
robbrockbank: Not sure what a "default network profile" is - I assume that this also disables adding a network-specific Calico profile

robbrockbank: Actually, given that this flag enables one feature and disables another, I wonder if it's named correctly.

I'm mulling over whether something like    `--use-docker-networking-label-policy` would be more descriptive.  Not sure.  WDYT?

calicoctl/commands/node/run.go
```diff
@@ -141,6 +142,13 @@ Options:
                            within the Docker containers that have been networked
                            by the Calico driver.
                            [default: cali]
+     --use-docker-networking-meta-data
```
robbrockbank: meta-data/meta data   ->  metadata

Actually, I think we should be using the term labels, not metadata.   Metadata is a more generic term, but it is labels that we are actually using.

Also, we aren't pulling selectors - the selectors are part of the policy resources used to select which endpoints the policy applies to.

I'd probably do something along the lines of:

```
--use-docker-networking-container-labels

Extract the Calico-namespaced Docker container labels
(org.projectcalico.label.*) and apply them to the container endpoints
for use with Calico policy (as opposed to using Calico profiles).  This option is only valid when using Calico
Docker networking, and when enabled will automatically disable the use of
Calico profiles to provide network isolation.
```

robbrockbank: I'd recommend getting a second opinion on that though.

tmjd: I went with meta data because that is what the docker run command help calls that data, I did have a note in one of my revisions like `(as set with --label)` but removed it.
I'm good with going back to calling them labels but just wanted to put some reasoning behind what I did.
The pulling selectors was referring to them being pulled from the Docker container metadata, I like your phrasing of them being 'Extracted'.

matthewdupre: I'm sorry to throw more spanners in, but "disable the use of Calico profiles" is unclear: I'm not sure what that actually means.

I think the best way to go is to say something about the effect - will traffic be allowed at all?  If not, maybe say something like "when enabled traffic must be explicitly allowed by configuring Calico policies.".  People might not know how the libnetwork networks turn into profiles, but they need to know what happens to traffic.

matthewdupre: To your question, I think I prefer labels too.  https://docs.docker.com/engine/userguide/labels-custom-metadata/ suggests that labels are an implementation of metadata, and it's labels we're going to extract, rather than generic (possibly not-yet-implemented) metadata.

robbrockbank: My original objection to the term "pulling selectors" wasn't the term pulling - it was the term selectors.  Calico uses both the terms "labels" and "selectors" in its policy model.  What we are pulling from the container and applying to the endpoints are the "labels".  The "selectors" are defined on the policy resources.

I'm happy to follow Matts suggestion about talking about traffic and explicit allowing it through calico policy.  If it can be done succinctly I think mentioning that profiles will not be used to enforce network isolation that would also be good - but at the very least we should at least be going into that level of detail in our documentation.

tests/st/libnetwork/test_labeling.py
```diff
@@ -81,9 +81,7 @@ def setUpClass(cls):
         cls.hosts.append(cls.host2)
 
         for host in cls.hosts:
-            start_calico_node(host,
-                              ["CALICO_LIBNETWORK_LABEL_ENDPOINTS=true",
-                               "CALICO_LIBNETWORK_CREATE_PROFILES=false"])
+            host.start_calico_node(options='--use-docker-networking-container-labels')
```
robbrockbank: Should be able to get rid of the start_calico_node method now.

tmjd: oops yeah, thanks.

calicoctl/commands/node/run.go
tmjd: This is what I have right now
```
Extract the Calico-namespaced Docker container labels
(org.projectcalico.label.*) and apply them to the
container endpoints for use with Calico policy.
When this option is enabled Calico Docker networking
must be enabled, traffic must be explicitly allowed
by configuring Calico policies, and Calico profiles
are not used to enforce network isolation.
```
I'm thinking about dropping the`Calico Docker networking must be enabled' because it is a little obvious and the code has a check that will insure that it is enabled.  How's that sound?

tmjd: Actually here is what I'm liking right now
```
Extract the Calico-namespaced Docker container labels
(org.projectcalico.label.*) and apply them to the
container endpoints for use with Calico policy.
When this option is enabled traffic must be
explicitly allowed by configuring Calico policies
and Calico profiles are disabled.
```
Probably not necessary to say what profiles are used for, just that they are disabled.

tmjd: Oh thanks for pointing that out about the selectors vs labels.  I was kind of thinking of them as interchangeable but I think I get the differentiation.  Labels can be thought of as a key:value on a workload and a selector is matching one or more key:value for the purposes of applying policy.
Please correct me if I'm wrong.

tests/st/libnetwork/test_labeling.py
tmjd: I wanted to make sure I addressed the correct thing.  Are you referring to removing the method definition at the bottom of the file?  I believe the place you commented still requires the `host.start_calico_node` call, if I have misunderstood please let me know.  Thanks.

robbrockbank:
LGTM

Issue #1518, opened 2017-02-14
calico/node fails if there are active iptables rules that need plugins


fasaxc:
Felix runs `iptables-save` to dump the contents of iptables for inspection.  If another process, outside the container, has added a rule that requires a plugin that we don't have inside our container then `iptables-save` fails.

One example is the shaman monitor daemon used in Virtuozzo linux.

jaxxstorm:
Any suggestions on how to rectify this? I've tried:
Mounting `/usr/lib64/xtables` into `/lib/xtables` on the container.
Made sure kernel modules are mounted, and can see the custom module with `lsmod`

It seems iptables isn't picking up the module, as when I run:

```
cat /proc/net/ip_tables_matches
```

I can't see the module loaded.

jaxxstorm:
One option here is to have BIRD and conf running inside a container, then an env var to disable the calico-felix binary.

I opened an issue here to detail instructions on how to get this working as well: https://github.com/projectcalico/felix/issues/1340

fasaxc:
Looks like @jaxxstorm was able to get his particular case working by running Felix outside the container: https://github.com/projectcalico/felix/issues/1340

fasaxc:
Even if we can't solve the root cause, we could have made this easier to diagnose by doing a sanity check of iptables in statup.go.  For example, just run `iptables-save` and bail out with an error if it fails?

fasaxc:
Assigning to @heschlie for triage.

caseydavenport:
We should at least document this behavior - that might be enough to close the issue given nobody else has really hit this?

caseydavenport:
I'm going to close this for now. If we start seeing this pop up more often I'd be happy to reopen.

knfoo:
I have run it to this as well.
I am using the geoip module.
I would like to avoid to run calico-felix binary on the host.

fasaxc:
Not really sure what we an do  about custom modules.  Best option might be for you to build your own calico/node image with the extra module added.

fasaxc:
I guess if there are just one or two modules and they're not large, we could just add them to the main image; is the geoip module packaged for RH?

knfoo:
I installed the xtables-common-addons in ubuntu to get it working.
I "hacked" the calico-node deployment to do

```
        - name: calico-node
          image: calico/node:v3.15.0
          command: ["/usr/bin/bash"]
[...]

PR #1522, opened 2017-02-15
Allow disabling felix

jaxxstorm:
helps with #1518

### DIFF
--- a/calico_node/filesystem/etc/rc.local
+++ b/calico_node/filesystem/etc/rc.local
@@ -27,7 +27,8 @@ mkdir /etc/service/enabled
 # dirs contained in Docker image. This is due to bug in Docker with graphdriver
 # overlay on CentOS 7.X kernels (https://github.com/docker/docker/issues/15314)
 
-# In certain unique situation we want to disable felix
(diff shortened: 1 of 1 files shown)


calico_node/filesystem/etc/rc.local
```diff
@@ -27,8 +27,10 @@ mkdir /etc/service/enabled
 # dirs contained in Docker image. This is due to bug in Docker with graphdriver
 # overlay on CentOS 7.X kernels (https://github.com/docker/docker/issues/15314)
 
-# Felix is always enabled
-cp -a /etc/service/available/felix /etc/service/enabled/
+# In certain unique situation we want to disable felix
```
fasaxc: How about adding an example:  `# Allow Felix to be disabled, for example, if the user is running Felix outside the container.`

jaxxstorm: Good idea, fixed!

fasaxc:
@heschlie Can this go in, or are frozen for the release?

jaxxstorm:
ping? @heschlie

Issue #1537, opened 2017-02-21
CALICO_IPV4POOL_CIDR will accept pools which calicoctl will not


lwr20:
I run this, expecting it to fail because I've asked for a pool which is too small (10.0.0.0/32):
`$ docker run --net=host --privileged --name=calico-node -d --restart=always -e IP_AUTODETECT_METHOD=first-found -e CALICO_LIBNETWORK_CREATE_PROFILES=true -e CALICO_LIBNETWORK_LABEL_ENDPOINTS=false -e CALICO_LIBNETWORK_IFPREFIX=cali -e NODENAME=drogon -e CALICO_NETWORKING_BACKEND=bird -e CALICO_IPV4POOL_CIDR=10.0.0.0/32 -e CALICO_NAT_ENABLED=false -e CALICO_LIBNETWORK_ENABLED=true -e IP6_AUTODETECT_METHOD=first-found -e ETCD_AUTHORITY=127.0.0.1:2379 -e ETCD_SCHEME=http -e ETCD_ENDPOINTS= -v /var/log/calico:/var/log/calico -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock quay.io/calico/node:v1.1.0-rc5
`

But it succeeds:
```
$ ./calicoctl get ippool -o yaml
- apiVersion: v1
  kind: ipPool
  metadata:
    cidr: 10.0.0.0/32
  spec:
    nat-outgoing: true
- apiVersion: v1
  kind: ipPool
  metadata:
    cidr: fd80:24e2:f998:72d6::/64
  spec:
    nat-outgoing: true
```

If I try to create that pool via calicoctl, it is rejected (as expected):
```
$ ./calicoctl get ippool -o yaml > pool
$ ./calicoctl apply -f pool
Failed to execute command: error with field CIDR = '10.0.0.0/32' (IP pool is too small)
```

lwr20:
Similarly for `CALICO_IPV6POOL_CIDR` accepting `fd00::/123`

tmjd:
Ready for retest with #1542 merged

lwr20:
Fix looks good in calico/node:v1.1.0-rc5-14-g1f5381f3

Issue #1534, opened 2017-02-17
CALICO_IPIP_ENABLED should only enable on IPv4 pools


tmjd:
IPIP is not valid on IPv6 pools. So the environment variable should only set it on default IPv4 pools.

tmjd:
Did a test using Vagrant k8s and verified IPIP was enabled on just IPv4 pool and not IPv6.  Good to go.

PR #1538, opened 2017-02-21
Add tests of default pool settings

lwr20:
Will not pass CI until https://github.com/projectcalico/calicoctl/pull/1542 is merged (or else fixes for #1539, #1537, #1534 go in somehow)

### DIFF
--- a/tests/st/calicoctl/test_default_pools.py
+++ b/tests/st/calicoctl/test_default_pools.py
@@ -1,0 +1,161 @@
+# Copyright (c) 2015-2016 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import functools
+import logging
+
+import netaddr
+import yaml
+from nose_parameterized import parameterized
+
+from tests.st.test_base import TestBase
+from tests.st.utils.docker_host import DockerHost, CLUSTER_STORE_DOCKER_OPTIONS, NODE_CONTAINER_NAME
+from tests.st.utils.exceptions import CommandExecError
+from tests.st.utils.utils import get_ip, wipe_etcd, retry_until_success
+
+_log = logging.getLogger(__name__)
+_log.setLevel(logging.DEBUG)
+
+
+class TestDefaultPools(TestBase):
+    @classmethod
+    def setUpClass(cls):
+        # First, create a (fake) host to run things in
+        cls.host = DockerHost("host",
+                              additional_docker_options=CLUSTER_STORE_DOCKER_OPTIONS,
+                              start_calico=False,
+                              dind=False)
+
+    def setUp(self):
+        try:
+            self.host.execute("docker rm -f calico-node")
+        except CommandExecError:
+            # Presumably calico-node wasn't running
+            pass
+        wipe_etcd(get_ip())
+
+    @classmethod
+    def tearDownClass(cls):
+            cls.host.cleanup()
+
+    @parameterized.expand([
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/27", 0, None, "Too small"),
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/32", 0, None, "Too small, but legal CIDR"),
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/33", 0, None, "Impossible CIDR"),
+        (False, "CALICO_IPV4POOL_CIDR", "256.0.0.0/24", 0, None, "Invalid IP"),
+        (True, "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", 2, None, "Typical non-default pool"),
+        (True, "CALICO_IPV4POOL_CIDR", "10.0.0.0/26", 2, None, "Smallest legal pool"),
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", 2, None, "Smallest legal pool"),
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/123", 0, None, "Too small"),
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/128", 0, None, "Too small, but legal CIDR"),
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/129", 0, None, "Impossible CIDR"),
+        (True, "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", 2, "cross-subnet", "Typ. non-def pool, IPIP"),
+        (True, "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", 2, "always", "Typ. non-default pool, IPIP"),
+        (True, "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", 2, "off", "Typical pool, explicitly no IPIP"),
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", 2, "always", "IPv6 - IPIP not permitted"),
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", 2, "cross-subnet", "IPv6 - IPIP not allowed"),
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", 2, "off", "IPv6, IPIP explicitly off"),
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/122", 0, "junk", "Invalid IPIP value"),
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", 0, "reboot", "Invalid IPIP value"),
+        (True, "CALICO_IPV4POOL_CIDR", "0.0.0.0/0", 2, None, "Valid, but not sensible"),
+        (True, "CALICO_IPV6POOL_CIDR", "::/0", 2, None, "Valid, but not sensible"),
+        (True, "CALICO_IPV6POOL_CIDR", "fd80::0:0/120", 2, None, "Valid, but non-canonical form"),
+        (False, "CALICO_IPV6POOL_CIDR", "1.2.3.4/24", 0, None, "Wrong type"),
+        (False, "CALICO_IPV4POOL_CIDR", "fd00::/24", 0, None, "Wrong type"),
+        (True, "CALICO_IPV6POOL_CIDR", "::0:a:b:c:d:e:0/120", 2, None, "Valid, non-canonical form"),
+        (False, "CALICO_IPV4POOL_CIDR", "1.2/16", 0, None, "Valid, unusual form"),
+    ])
+    def test_default_pools(self, success_expected, param, value, exp_num_pools, ipip, description):
+        """
+        Test that the various options for default pools work correctly
+        """
+        _log.debug("Test description: %s", description)
+        # Get command line for starting docker
+        output = self.host.calicoctl("node run --dryrun --node-image=%s" % NODE_CONTAINER_NAME)
+        base_command = output.split('\n')[-4].rstrip()
+
+        # Modify command line to add the options we want to test
+        env_inserts = "-e %s=%s " % (param, value)
+        if ipip is not None:
+            env_inserts += "-e CALICO_IPV4POOL_IPIP=%s " % ipip
+        prefix, _, suffix = base_command.partition("-e")
+        command = prefix + env_inserts + "-e" + suffix
+
+        # Start calico-docker
+        self.host.execute(command)
+
+        if not success_expected:
+            # check for "Calico node failed to start"
+            self.wait_for_node_log("Calico node failed to start")
+            return
+
+        # Check we started OK
+        self.wait_for_node_log("Calico node started successfully")
+        # check the expected pool is present
+        pools_output = self.host.calicoctl("get ippool -o yaml")
+        pools_dict = yaml.safe_load(pools_output)
+        cidrs = [pool['metadata']['cidr'] for pool in pools_dict]
+        # Convert to canonical form
+        value = str(netaddr.IPNetwork(value))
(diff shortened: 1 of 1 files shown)


tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,150 @@
...
+        # Get command line for starting docker
+        output = self.host.calicoctl("node run --dryrun")
+        base_command = output.split('\n')[-4].rstrip()
+
+        # Modify command line to add the options we want to test
+        env_index = base_command.find("-e")
+        command = base_command[:env_index] + "-e %s=%s " % (param, value) + base_command[env_index:]
+        if ipip is not None:
+            command = command[:env_index] + \
+                      "-e CALICO_IPIP_ENABLED=%s " % ipip + \
```
tmjd: The changes that this PR is waiting on also change the name of this variable to CALICO_IPV4POOL_IPIP with possible values of <empty string>, "off", "always", and "cross-subnet".  Just pointing that out but you probably already know changes were coming.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,167 @@
...
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", "off"),  # Typical pool, explicitly no IPIP
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", "always"),  # IPv6 - IPIP not permitted
+        (True, "CALICO_IPV6POOL_CIDR", "fd00::/122", "cross-subnet"),  # IPv6 - IPIP not permitted
+        (True,  "CALICO_IPV4POOL_CIDR", "0.0.0.0/0",    None),    # Valid, but not sensible
+        (True,  "CALICO_IPV6POOL_CIDR", "::/0",         None),    # Valid, but not sensible
+        (True,  "CALICO_IPV6POOL_CIDR", "fd80::0:0/120", None),   # Valid, but non-canonical form
+        (False, "CALICO_IPV6POOL_CIDR", "1.2.3.4/24",    None),   # Wrong type
+        (False, "CALICO_IPV4POOL_CIDR", "fd00::/24",     None),   # Wrong type
+        (True,  "CALICO_IPV6POOL_CIDR", "::0:a:b:c:d:e:0/120", None),  # Valid, non-canonical form
+        (False, "CALICO_IPV4POOL_CIDR", "1.2/16",       None),    # Valid, unusual form
```
tmjd: If that is a valid format this should probably work.  I believe the IP validation we depend on comes from the go library so it might just be a version that it does not support.
Either way I don't think it is a big deal.  We could create an issue around this but I feel like it could be marked 'won't fix' .... or just not worry about it all.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,167 @@
...
+        (False, "CALICO_IPV4POOL_CIDR", "256.0.0.0/24", None),    # Invalid IP
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24",  None),    # Typical non-default pool
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/26",  None),    # Smallest legal pool
+        (True,  "CALICO_IPV6POOL_CIDR", "fd00::/122",   None),    # Smallest legal pool
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/123",   None),    # Too small
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/128",   None),    # Too small, but legal
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/129",   None),    # Impossible CIDR
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", "cross-subnet"),  # Typical non-default pool, IPIP
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", "always"),  # Typical non-default pool, IPIP
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24", "off"),  # Typical pool, explicitly no IPIP
```
tmjd: Could have another test case with an inalid string for the IPIP mode?  Not too necessary to add because I know there is one at the unit test level but just wanted to mention for completeness.

tests/st/calicoctl/test_default_pools.py
lwr20: It is a valid format, but it is an unusual one.  I guess we'll fix this if/when someone raises an issue and provides a user-story.

tests/st/calicoctl/test_default_pools.py
lwr20: Good call.  I'll add that.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+            pass
+        wipe_etcd(get_ip())
+
+    @classmethod
+    def tearDownClass(cls):
+        for host in cls.hosts:
+            host.cleanup()
+
+    @parameterized.expand([
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/27",  None),    # Too small
```
fasaxc: Please move the informative descriptions from comment into a parameter.  That way, you can log them and they'll probably get spat out by `parametrised.expand` too.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+        Test that the various options for default pools work correctly
+        """
+        # Get command line for starting docker
+        output = self.host.calicoctl("node run --dryrun")
+        base_command = output.split('\n')[-4].rstrip()
+
+        # Modify command line to add the options we want to test
+        # Command line will attempt to use quay.io/calico/node:latest - we want calico/node:latest
+        base_command = base_command.replace("quay.io/", "")
+        env_index = base_command.find("-e")
```
fasaxc: This might be a bit cleaner since it avoids the indexing, which is always fiddly:
```
env_inserts = "-e %s=%s " % (param, value)
if ipip is not None:
    env_inserts += "-e CALICO_IPV4POOL_IPIP=%s " % ipip
prefix, _, suffix := base_command.partition("-e")
command = prefix + env_inserts + "-e" + suffix
```

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+        cidrs = [pool['metadata']['cidr'] for pool in pools_dict]
+        # Convert to canonical form
+        value = str(netaddr.IPNetwork(value))
+        assert value in cidrs, "Didn't find %s in %s" % (value, cidrs)
+
+        # Dump pools and attempt to load them with calicoctl (to confirm consistency)
+        self.host.calicoctl("get ippool -o yaml > pool")
+        self.host.calicoctl("apply -f pool")
+
+        assert len(pools_dict) <= 2, "Found too many pools in %s" % pools_dict
```
fasaxc: `<=` feels a bit fuzzy, should the number of expected pools be passed in?

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+        if ipip is not None:
+            command = command[:env_index] + \
+                      "-e CALICO_IPV4POOL_IPIP=%s " % ipip + \
+                      command[env_index:]
+
+        # Start calico-docker
+        self.host.execute(command)
+
+        def check():
+            assert expected_string in self.host.execute("docker logs calico-node"), \
```
fasaxc: Not keen on this using a "global" my preference would be to move this to its own method:
```
def assert_calico_node_log_contains(expected_msg):
    def check():
        assert ...
    
    retry_until_success(check, ...)
```
Then it can be shared with the other test too.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+    @parameterized.expand([
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/27",  None),    # Too small
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/32",  None),    # Too small, but legal
+        (False, "CALICO_IPV4POOL_CIDR", "10.0.0.0/33",  None),    # Impossible CIDR
+        (False, "CALICO_IPV4POOL_CIDR", "256.0.0.0/24", None),    # Invalid IP
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/24",  None),    # Typical non-default pool
+        (True,  "CALICO_IPV4POOL_CIDR", "10.0.0.0/26",  None),    # Smallest legal pool
+        (True,  "CALICO_IPV6POOL_CIDR", "fd00::/122",   None),    # Smallest legal pool
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/123",   None),    # Too small
+        (False, "CALICO_IPV6POOL_CIDR", "fd00::/128",   None),    # Too small, but legal
```
fasaxc: _Why_ is this legal?

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,180 @@
...
+
+from tests.st.test_base import TestBase
+from tests.st.utils.docker_host import DockerHost
+from tests.st.utils.exceptions import CommandExecError
+from tests.st.utils.utils import ETCD_CA, ETCD_CERT, \
+    ETCD_KEY, ETCD_HOSTNAME_SSL, ETCD_SCHEME, get_ip, wipe_etcd, retry_until_success
+
+# Load the calico-node we've just built
+POST_DOCKER_COMMANDS = ["docker load -i /code/calico-node.tar"]
+if ETCD_SCHEME == "https":
```
robbrockbank: These ADDITIONAL_DOCKER_OPTIONS are now defined in docker_host.py - better to use those.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Test that the various options for default pools work correctly
+        """
+        _log.debug("Test description: %s", description)
+        # Get command line for starting docker
+        output = self.host.calicoctl("node run --dryrun")
+        base_command = output.split('\n')[-4].rstrip()
+
+        # Modify command line to add the options we want to test
+        # Command line will attempt to use quay.io/calico/node:latest - we want calico/node:latest
```
robbrockbank: or you could pass in the --node-image option with the image that you want - seems better than a search/replace.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,168 @@
...
+        base_command = base_command.replace("quay.io/", "")
+
+        env_inserts = "-e %s=%s " % (param, value)
+        if ipip is not None:
+            env_inserts += "-e CALICO_IPV4POOL_IPIP=%s " % ipip
```
robbrockbank: I assume if you do something like:

`CALICO_IPV4POOL_IPIP=always calicoctl node run --dryrun` the command will include the IPIP variable since it should do the CALICO_* and FELIX_* environment pass-thru.

robbrockbank: Oh scrap that.... just realized that PR has not yet been merged.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,168 @@
...
+
+_log = logging.getLogger(__name__)
+_log.setLevel(logging.DEBUG)
+
+
+class TestDefaultPools(TestBase):
+    @classmethod
+    def setUpClass(cls):
+        # First, create a host to run things in
+        cls.hosts = []
```
robbrockbank: Do we need hosts here - in some places you assume only one host and others (cleanup) that there are multiple hosts.  Maybe just get rid of it and assume this is a one-host suite.

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Check NAT setting
+        assert pool['spec']['nat-outgoing'] is True, "Didn't find nat enabled in pool %s" % pool
+
+    def test_no_default_pools(self):
+        """
+        Test that NO_DEFAULT_POOLS works correctly
+        """
+        # Start calico-docker
+        self.host.calicoctl("node run --no-default-ippools")
```
robbrockbank: docker_host has a start_calico_node() method that allows you to pass in extra options.  It also takes care of using the correct container name.

tests/st/calicoctl/test_default_pools.py
robbrockbank: Actually....

docker_host has a start_calico_node() method that allows you to pass in extra options.  It also takes care of using the correct container name.

robbrockbank: NODE_CONTAINER_NAME is defined in docker_host

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,168 @@
...
+        # check the expected pool is present
+        pools_output = self.host.calicoctl("get ippool -o yaml")
+        pools_dict = yaml.safe_load(pools_output)
+        cidrs = [pool['metadata']['cidr'] for pool in pools_dict]
+        # Convert to canonical form
+        value = str(netaddr.IPNetwork(value))
+        assert value in cidrs, "Didn't find %s in %s" % (value, cidrs)
+
+        # Dump pools and attempt to load them with calicoctl (to confirm consistency)
+        self.host.calicoctl("get ippool -o yaml > pool")
```
robbrockbank: I assume this is going to write some files out to the local file system - might be good to use a file name that we already have in our .gitignore

lwr20: I think this will write out to the Dockerhost's file system and get destroyed with the host.  I'm certainly not seeing this file in my local file system?

tests/st/calicoctl/test_default_pools.py
```diff
@@ -0,0 +1,162 @@
...
+            env_inserts += "-e CALICO_IPV4POOL_IPIP=%s " % ipip
+        prefix, _, suffix = base_command.partition("-e")
+        command = prefix + env_inserts + "-e" + suffix
+
+        # Start calico-docker
+        self.host.execute(command)
+
+        if not success_expected:
+            # check for "Calico node failed to start"
+            expected_string = "Calico node failed to start"
```
fasaxc: I'd probably wrap this up as `wait_for_node_log("Callico node failed to start")`, for example

tests/st/calicoctl/test_default_pools.py
robbrockbank: Ah you are using a Docker-in-Docker host - I think because you just had a single host I assumed you weren't.

robbrockbank:
LGTM once the tests pass.

Issue #1539, opened 2017-02-21
CALICO_IPV4POOL_CIDR accepts IPv6 addresses


lwr20:
I run calico-node with:
```
$ docker exec -it host sh -c 'docker run --net=host --privileged --name=calico-node -d --restart=always -e CALICO_IPV4POOL_CIDR=fd00::/24 -e ETCD_AUTHORITY=10.240.0.12:2379 -e ETCD_SCHEME=http -e IP6_AUTODETECT_METHOD=first-found -e CALICO_LIBNETWORK_CREATE_PROFILES=true -e CALICO_LIBNETWORK_LABEL_ENDPOINTS=false -e NODENAME=2b6048a03e65 -e CALICO_NETWORKING_BACKEND=bird -e NO_DEFAULT_POOLS= -e CALICO_LIBNETWORK_ENABLED=true -e IP_AUTODETECT_METHOD=first-found -e CALICO_LIBNETWORK_IFPREFIX=cali -e ETCD_ENDPOINTS= -v /var/log/calico:/var/log/calico -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock quay.io/calico/node:v1.1.0-rc5'
```
expecting it to fail, and it does not:
```
$ docker exec -it host sh -c 'docker logs calico-node'
  # time="2017-02-21T19:53:16Z" level=info msg="Loading config from environment"
  # Skipping datastore connection test
  # time="2017-02-21T19:53:16Z" level=info msg="Building new node resource" Name=2b6048a03e65
  # time="2017-02-21T19:53:16Z" level=info msg="Initialise BGP data"
  # Using autodetected IPv4 address on interface eth0: 172.17.0.4/16
  # No AS number configured on node resource, using global value
  # time="2017-02-21T19:53:16Z" level=info msg="Ensure default IPv6 pool is created. IPIP enabled: false"
  # Created default IPv6 pool (fd00::/24) with NAT outgoing enabled. IPIP enabled: false
  # time="2017-02-21T19:53:16Z" level=info msg="IPv6 supported on this platform: true"
  # time="2017-02-21T19:53:16Z" level=info msg="Ensure default IPv6 pool is created. IPIP enabled: false"
  # Created default IPv6 pool (fd80:24e2:f998:72d6::/64) with NAT outgoing enabled. IPIP enabled: false
[...]

PR #1542, opened 2017-02-22
Fix IPIP env var and add several checks

tmjd:
- IPIP env var only applies to IPv4, renamed as appropriate
- Check IPIP env var for validity
- Check IP Pool size
- Check Ipv4 var has v4 Address and Ipv6 var has v6 Address
- Added tests for the above

I believe these changes address #1539 #1537 #1534

### DIFF
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -27,6 +27,7 @@ import (
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/ipip"
 	"github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/numorstring"
 )
@@ -459,7 +460,6 @@ func configureIPPools(client *client.Client) {
 	// Read in environment variables for use here and later.
 	ipv4Pool := os.Getenv("CALICO_IPV4POOL_CIDR")
 	ipv6Pool := os.Getenv("CALICO_IPV6POOL_CIDR")
-	ipipEnabled := strings.ToLower(os.Getenv("CALICO_IPIP_ENABLED")) == "true"
 
 	if strings.ToLower(os.Getenv("NO_DEFAULT_POOLS")) == "true" {
 		if len(ipv4Pool) > 0 || len(ipv6Pool) > 0 {
@@ -471,11 +471,14 @@ func configureIPPools(client *client.Client) {
 		return
 	}
 
+	ipv4IpipModeEnvVar := strings.ToLower(os.Getenv("CALICO_IPV4POOL_IPIP"))
+
 	// Get a list of all IP Pools
 	poolList, err := client.IPPools().List(api.IPPoolMetadata{})
 	if err != nil {
 		fatal("Unable to fetch IP pool list: %s", err)
 		terminate()
+		return // not really needed but allows testing to function
 	}
 
 	// Check for IPv4 and IPv6 pools.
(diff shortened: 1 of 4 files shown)


robbrockbank:
At the moment this doesn't fix #1537 (as far as I can tell).

I think you will need to get your libcalico-go fix in and then add an extra UT for that.

calico_node/startup/startup.go
```diff
@@ -542,20 +560,44 @@ func ipv6Supported() bool {
...
 	}
 
+	// Use off when logging for disabled instead of blank
+	ipipModeString := "off"
+	if ipipMode != ipip.Undefined {
+		ipipModeString = string(ipipMode)
+	}
+
+	// Validate the pool Metadata and Spec
+	err := validator.Validate(pool.Metadata)
```
robbrockbank: I think the real fix is to make sure we validate the data in the API.  So I think we might have to bite the bullet and fix up that issue as it's a caring omission.

calico_node/startup/startup_test.go
```diff
@@ -202,44 +216,35 @@ var _ = Describe("FV tests against a real etcd", func() {
...
-		configureIPPools(c)
+			Expect(my_ec).To(Equal(1))
+		},
 
-		It("should have terminated", func() {
-			Expect(exitCode).To(Equal(1))
-		})
-	})
+		Entry("Bad IPv4 Pool CIDR", []EnvItem{{"CALICO_IPV4POOL_CIDR", "172.16.0.0a/24"}}),
+		Entry("Too small IPv4 Pool CIDR", []EnvItem{{"CALICO_IPV4POOL_CIDR", "172.16.0.0/27"}}),
```
tmjd: I think this test is checking that #1537 is fixed

tmjd: My changes to libcalico-go aren't necessary for that Validation, that validation was already being done by libcalico-go.  Or I misunderstood the validation that it does.

tmjd:
I think we are waiting on https://github.com/projectcalico/libcalico-go/pull/347 so this code can use the real fix in the API?

calico_node/startup/startup_test.go
robbrockbank: Ah yes you are quite right the CIDR check was already in place.  I think now, though, since we have moved the validation check into the Create() method in the API - we need to update glide to pick up the libcalico-go changes I think.  Is that not the case?

tmjd: Yes it is, I included that in my most recent update.

robbrockbank: oh crikey - i'm having a brain fart .... sorry!  yup didn't see that :-/

calico_node/startup/startup.go
```diff
@@ -542,20 +547,35 @@ func ipv6Supported() bool {
 
 // createIPPool creates an IP pool using the specified CIDR.  This
 // method is a no-op if the pool already exists.
-func createIPPool(client *client.Client, cidr *net.IPNet, ipip bool) {
+func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
 	version := cidr.Version()
+	ipipMode := ipip.Mode(ipipModeName)
+
+	// off is not an actual valid value so switch it to an empty string
+	if "off" == ipipModeName {
```
robbrockbank: A nit, but it's more standard to have the variable on the LHS .

robbrockbank:
one minor comment then LGTM

calico_node/startup/startup.go
tmjd: NP, not sure why I did it that way, probably a holdover from it doing that way in C++ so the compiler would complain if it was an assignment to a constant instead of the equality operator.

calico_node/startup/startup_test.go
shapeofarchitect: But why is this check included , I used /16 and it errored out saying too strict check , while I use /18 it pass and create ippoll !

PR #1557, opened 2017-03-08
calico/node: Use final 2.1 felix

tomdee:


tomdee:
Tests won't pass till felix is released

PR #1564, opened 2017-03-10
Edit command

robbrockbank:
Allows for `calicoctl edit X`, which opens resource X in an editor and applies changes on-save to the datastore.

### DIFF
--- a/calicoctl/commands/edit.go
+++ b/calicoctl/commands/edit.go
@@ -1,0 +1,257 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO
+// Better error handling
+// Distinguish between deleted updates and other failures
+// Copyright stuff related to k8s
+// Windows/Linus newlines etc.
+// Need to verify that user is not adding new entry or changing resource identifiers
+// CreateResourcesFromBytes needs to be converted to slice ... this indicates a bug
+//   in the CreateResourcesFromBytes which must be returning structs rather than pointers.
+
+package commands
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"os"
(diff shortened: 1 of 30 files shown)


robbrockbank:
This was a prototype - that I don't think needs a whole lot to get it production ready.  I'm leaving here for now so that if someone wants to take it on they can.  There is a TODO list in the edit.go file.

calicoctl/commands/resources.go
```diff
@@ -102,6 +102,18 @@ func convertToSliceOfResources(loaded interface{}) []unversioned.Resource {
 	return r
 }
 
+func convertToSliceOfResourceObjects(resources []unversioned.Resource) ([]unversioned.ResourceObject, error) {
```
robbrockbank: I think the method convertToSliceOfResources should actually be convertToSliceOfResourceObjects ... I think that was it's intention - so shouldn't need this separate function.

calicoctl/commands/edit.go
```diff
@@ -255,3 +263,21 @@ func addComment(b []byte, err error) []byte {
...
+// by comparing the ID fields.
+func ValidateIDFieldsSame(original, changed []unversioned.ResourceObject) error {
+	if len(original) != len(changed) {
+		return errors.New("Number of resources after editing does not match the number of resources provided. Cannot add or remove resources during edit.")
+	}
+
+	// Assume that that the resources are kept in the same order
+	for i, _ := range original {
+		if original[i].String() != changed[i].String() {
+			return fmt.Errorf("Cannot change metadata fields: %s should match %s", changed[i].String(), original[i].String())
```
robbrockbank: not quite true - you are allowed to change the labels which are a metadata field.

calicoctl/commands/edit.go
```diff
@@ -174,9 +169,20 @@ Description:
 
 		// Convert the file into a set of resources.  If this succeeds
 		// then it passes validation, so exit this loop.
+		var editedObjs []unversioned.ResourceObject
+		var prevObjs []unversioned.ResourceObject
 		resources, err = resourcemgr.CreateResourcesFromBytes(b)
 		if err == nil {
```
robbrockbank: I think it would be easier to invert this logic to reduce the nesting.

Better I think now to check `if err != nil` to log a specific log and then `continue` the loop, eg.

```
if resources, err = resourcemgr.CreateResourcesFromBytes(b); err != nil {
    log
    continue
} else if editedObjs, err = convertToSliceOfResourceObjects(resources); err != nil {
    log
    continue
} else ...
```

mgleung:
TODO:
[] Documentation in https://github.com/projectcalico/calico

caseydavenport:
This PR has gotten pretty out of date, and is quite old.

I'd love to see this feature added, but for now I'm going to close this. Shout if you'd like to pick it back up!

PR #1565, opened 2017-03-10
Turn off confd discovery

djosborne:
This one's a doozy.

There's a bug in etcd (3.1.0, 3.1.1, and 3.1.2) where [advertise-client-urls is always overwritten to be localhost:2379, even when it is manually set to something else](https://github.com/coreos/etcd/issues/7348). 

If I have a client programmed to connect to a remote etcd, and my client has "discovery" turned on, the etcd server will tell the client to reconnect via `localhost:2379`. If my client isn't on the same node as etcd, this will of course fail.

Currently, all other calico components besides confd (i.e. Felix, startup.go, calicoctl, and k8s-policy) do not connect with "discovery" on. They ignore the bad advertise url and have no issue reading and writing from etcd.

Confd, however, defaults to discovery mode. It will reconnect to localhost and fail repeatedly printing the following message:

`2017-03-10T23:29:26Z my-host-30j9 confd[57]: ERROR 501: All the given peers are not reachable (failed to propose on members [https://localhost:2379] twice [last error: Get https://localhost:2379/v2/keys/calico/v1/ipam/v6?quorum=false&recursive=true&sorted=false: dial tcp 127.0.0.1:2379: connection refused]) [0]`

This PR turns off discovery mode in confd, preventing this from happening.

### DIFF
--- a/calico_node/filesystem/etc/rc.local
+++ b/calico_node/filesystem/etc/rc.local
@@ -63,10 +63,10 @@ case "$CALICO_NETWORKING_BACKEND" in
 	# confd needs a "-node" arguments for each etcd endpoint.
 	ETCD_ENDPOINTS_CONFD=`echo "-node=$ETCD_NODE" | sed -e 's/,/ -node=/g'`
 
-	confd -confdir=/etc/calico/confd -onetime ${ETCD_ENDPOINTS_CONFD} \
+	confd -confdir=/etc/calico/confd -onetime ${ETCD_ENDPOINTS_CONFD} -no-discover \
 	      -client-key=${ETCD_KEY_FILE} -client-cert=${ETCD_CERT_FILE} \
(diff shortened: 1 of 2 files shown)


caseydavenport:
Given all the other components don't do this we should definitely make the behavior consistent.

This LGTM.

What about all the other places we call confd?

e.g https://github.com/projectcalico/calicoctl/blob/master/calico_node/filesystem/etc/rc.local#L66

djosborne:
@caseydavenport Those other calls to confd are `-oneshot` which AFAICT don't make use of the discovery address. So I don't think they _need_ to be changed. But I went ahead and changed them anyways :smile:

caseydavenport:
Interesting.. so the discovery stuff behaves differently for oneshot vs not?

djosborne:
@caseydavenport AFAICT, yes

tomdee:
I support merging this 😄

PR #1570, opened 2017-03-15
Update templates to handle missing node BGP IP

robbrockbank:
Check that the IPv4 address key is present before trying to get the value.

It's a very "mainline" change in the template so it's getting good coverage in the STs, but there is no ST to test the actual issue that led to us wanting this fix (since that would require jimmying around in etcd).

### DIFF
--- a/calico_node/filesystem/etc/calico/confd/templates/bird.cfg.mesh.template
+++ b/calico_node/filesystem/etc/calico/confd/templates/bird.cfg.mesh.template
@@ -62,13 +62,13 @@ template bgp bgp_template {
 {{if (json (getv "/global/node_mesh")).enabled}}
 {{range $host := lsdir "/host"}}
 {{$onode_as_key := printf "/host/%s/as_num" .}}
-{{$onode_ip_key := printf "/host/%s/ip_addr_v4" .}}{{$onode_ip := getv $onode_ip_key}}
+{{$onode_ip_key := printf "/host/%s/ip_addr_v4" .}}{{if exists $onode_ip_key}}{{$onode_ip := getv $onode_ip_key}}
 {{$nums := split $onode_ip "."}}{{$id := join $nums "_"}}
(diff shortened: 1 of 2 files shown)


insequent:
Hey @robbrockbank, I'm seeing issues similar to this, when trying to use calico/routereflector. My issue may be different than your intent with this PR, but the timing of your change was impeccable. I'm curious if maybe you saw something similar to what I was seeing?

I noticed the routereflector's bird.cfg.template failing in my confd log:
```
2017-03-16T00:56:22Z mgmt-01 /confd[15]: DEBUG Compiling source template /templates/bird.cfg.template
2017-03-16T00:56:22Z mgmt-01 /confd[15]: ERROR template: bird.cfg.template:66:71: executing "bird.cfg.template" at <getv $cnode_ip_key>: errorcalling getv: key does not exist
```
When I started checking the keys, I noticed a discrepancy in the hostnames:
```
"key": "/calico/bgp/v1/host/control-02.conant01.com",
  "key": "/calico/bgp/v1/host/control-02.conant01.com/peer_v4",
    "key": "/calico/bgp/v1/host/control-02.conant01.com/peer_v4/10.232.255.4",
    "value": "{\"ip\":\"10.232.255.4\",\"as_num\":\"64512\"}"
  "key": "/calico/bgp/v1/host/control-02",
    "key": "/calico/bgp/v1/host/control-02/ip_addr_v4",
    "value": "10.232.255.6"
    "key": "/calico/bgp/v1/host/control-02/ip_addr_v6",
    "value": ""
```
So in other words, the template is failing because while calico was generating these etcd entries, something switched it between using short and long hostnames. I'm trying to narrow down where this is happening, and I'm hopeful you might have seen this somewhere on your side. If not, please ignore this rudely long PR comment!

insequent:
Found the issue. It was a misconfigured ansible task. Deleting these comments to remove the spam.

PR #1576, opened 2017-03-23
Node status state

frnkdny:


tests/st/bgp/test_node_status_resilience.py
```diff
@@ -0,0 +1,123 @@
...
+                host.calicoctl("ipam release --ip=%s" % host_workload.ip)
+                host.execute("docker rm -f %s" % host_workload.name)
+                host.workloads.remove(host_workload)
+
+            for iteration in range(1, 4):
+                _log.debug("Iteration %s", iteration)
+                _log.debug("identify and pkill bird pid")
+                host3.execute("docker exec -it calico-node pgrep bird")
+                host3.execute("docker exec -it calico-node pkill bird")
+
```
lwr20: I'd kinda like to see a negative test in here, but I'm not totally sure if its possible.  What if the `pkill` fails to actually kill BIRD?  That might give us a false positive.  If we can check that the connection goes down momentarily, we could be fairly sure that BIRD has died.

lwr20: Also - we're only killing BIRD here - can we put this loop in a loop which cycles round and kills all the BGP clients in turn?  That way we'd test all the BGP backends.  

Or maybe we can pass a parameter into the test and run the test for each of BIRD/GoBGP, and have it report them as separate cases?

tests/st/bgp/test_node_status_resilience.py
```diff
@@ -0,0 +1,123 @@
...
+            self.assert_ip_connectivity(workload_list=[workload_host1,
+                                                       workload_host2,
+                                                       workload_host3],
+                                        ip_pass_list=[workload_host1.ip,
+                                                      workload_host2.ip,
+                                                      workload_host3.ip])
+
+            # Check the BGP status on the BIRD/GoBGP host.
+            _log.debug("==== docker exec -it calico-node ps -a  ====")
+            host3.execute("docker exec -it calico-node ps -a")
```
frnkdny: Added _log.debug().

tests/st/bgp/test_node_status_resilience.py
frnkdny: To address the possible false positive. Added an assert to validate that started process and new process have different pids after a pkill. 

I’ve parameterize some of the aspects that will make this script easy to test all the BGP backends. Although, I’d like to add this as an enhancement after merging with just the BIRD functionality.

frnkdny:
lwr20, I've made the requested updates please take another look at the latest push.

lwr20:
LGTM

PR #1578, opened 2017-03-24
Use BIRD v0.2.2-alpha2

neiljerram:


tomdee:
This might be better in a different branch (not master) until you have a non-alpha release of bird

neiljerram:
It's only here for testing purposes, really - note the lack of requested reviewers!  And I hope it won't be too much longer before we've done the needed testing and can release the BIRD code properly.

PR #1583, opened 2017-03-28
Restore default endpointtohostaction

### LINKED ISSUE #1575: calico/node:v1.1 changes FELIX_DEFAULTENDPOINTTOHOSTACTION default from RETURN to DENY
I've found that in both an openshift (k8s 1.4) a kubeadm cluster (k8s 1.5),  FELIX_DEFAULTENDPOINTTOHOSTACTION=ACCEPT must be explicitly set. Otherwise, pods started on the master are unable to speak to the Kubernetes API. 

In both environments, this flag did not need to be specified when using calico/node:v1.0.0. It was only upon upgrading to calico/node:v1.1.0 that the value needed to be explicitly set.

I think the connectivity issue that this causes will be tricky for newer users to diagnose as it took me awhile to narrow down that it was k8s-master only, and furthermore is commonly diagnosed by reading iptables chain packet counts, which can be a bit tricky for newer users. Furthermore, users performing a simple calico/node version bump upgrade could see their connectivity impacted unless they add this variable.

cc: @fasaxc

djosborne:
Fixes #1575

### DIFF
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -672,6 +672,16 @@ func ensureDefaultConfig(c *client.Client, node *api.Node) error {
 		return err
 	}
 
+	// Configure Felix to allow traffic from the containers to the host (if
+	// not otherwise firewalled by the host administrator or profiles).
+	// This is important for container deployments, where it is common
+	// for containers to speak to services running on the host (e.g. k8s
+	// pods speaking to k8s api-server, and mesos tasks registering with agent
+	// on startup).
+	if err := ensureFelixConfig(c, node.Metadata.Name, "DefaultEndpointToHostAction", "RETURN"); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


calico_node/startup/startup.go
```diff
@@ -672,6 +672,14 @@ func ensureDefaultConfig(c *client.Client, node *api.Node) error {
 		return err
 	}
 
+	// Configure Felix to allow traffic from the containers to the host (if
```
robbrockbank: It's not just Mesos deployments is it, it's also k8s no?  More generally though is this basically the default we want for containerized deployments?  If  so probably worth saying that here and that's why we don't do this processing generically in libcalico-go.

robbrockbank:
LGTM - I think one of the comments needs a little tweak because this isn't just for Mesos IIRC its going to be the default for all containerized deployments.

robbrockbank:
LGTM.

PR #1584, opened 2017-03-28
Add test to simulate GCE instance setup

neiljerram:
This change adds a test that sets up hosts' addressing and routing as it
is on a GCE instance, and checks that we still get connectivity between
workloads on different hosts.


tests/st/policy/test_profile.py
```diff
@@ -30,15 +30,16 @@
 class MultiHostMainline(TestBase):
     @parameterized.expand([
         #"tags",
-        "rules.tags",
+        ("rules.tags", False),
+        ("rules.tags", True),
```
robbrockbank: I don't think this should be part of the policy test (unless you think we need to be testing this with different sets of policy.  

We have a `test_ipip.py` in the `calicoctl` tests which I think could be moved over to the bgp tests (since it's more testing Bird) - and then maybe we could include this test in test_ipip.py.

robbrockbank: Once GoBGP supports IPIP fully I assume we should be running these tests on GoBGP?

No action for you here.  Just wondering whether GoBGP might be affected by the same issue as Bird.

neiljerram:
By the way, I also checked that the new testing:
- passes with BIRD_VER v0.2.0
- fails with BIRD_VER v0.2.2-alpha1
- passes with BIRD_VER v0.2.2-alpha2

Will look at moving to another part of the suite tomorrow.

neiljerram:
@robbrockbank I think this is good for you to take another look at now.

tests/st/bgp/test_ipip.py
```diff
@@ -287,3 +287,51 @@ def get_tunl_tx(self, host):
...
+
+            host1.start_calico_node()
+            host2.start_calico_node()
+
+            # We are simulating GCE instance routing, where there is a router
+            # between the instances, and each instance has a /32 address that
+            # appears not to be directly connected to any subnet.  Hence we
+            # also need to enable IP-in-IP to get from one host to the other.
+            host1.enable_ipip()
+            host2.enable_ipip()
```
robbrockbank: IP pools are globally configured and not host specific, so I think you can drop one these as the second call won't actually make any more updates.

robbrockbank:
LGTM - one minor comment, but I'm happy for this to go in.

tests/st/bgp/test_ipip.py
```diff
@@ -287,3 +287,51 @@ def get_tunl_tx(self, host):
...
+                        simulate_gce_routing=True,
+                        start_calico=False) as host1, \
+             DockerHost('host2',
+                        additional_docker_options=CLUSTER_STORE_DOCKER_OPTIONS,
+                        simulate_gce_routing=True,
+                        start_calico=False) as host2:
+
+            host1.start_calico_node()
+            host2.start_calico_node()
+
```
robbrockbank: It occurs to me that we should also be testing a failure case when IPIP is not enabled - otherwise we don't have any visibility that the emulated GCE network is really doing its thing.  WDYT?

neiljerram:
@robbrockbank @tmjd Done!

neiljerram:
@robbrockbank markups here are done now

tests/st/bgp/test_ipip.py
```diff
@@ -287,3 +288,62 @@ def get_tunl_tx(self, host):
...
+            # We are simulating GCE instance routing, where there is a router
+            # between the instances, and each instance has a /32 address that
+            # appears not to be directly connected to any subnet.  Hence we
+            # also need to enable IP-in-IP to get from one host to the other.
+            # (In the case where we don't do this, we will expect workload
+            # connectivity to fail.)
+            if with_ipip:
```
robbrockbank: Sorry for the back and forth....

A big chunk of the time it takes to run the STs is performing the setup of the docker hosts.  I think it would be better to reduce this to a single test rather than running twice with the following flow:

-  create workloads
-  check connectivity fails
-  turn on IPIP
-  check connectivity passes

that way we only have to set up the hosts once for this test *and* we get the added bonus of testing that you can enable IPIP after creating the workloads and it should kick into life.

robbrockbank:
One more comment.  Sorry!  Back with you @neiljerram

tests/st/bgp/test_ipip.py
robbrockbank: unless of course there is some reason why that wouldn't work :-)

neiljerram:
Oh dear, this is becoming rather demoralising now.  Please let's discuss.

neiljerram:
@robbrockbank Ready for your eyes again...

robbrockbank:
LGTM

PR #1602, opened 2017-04-11
Remove ETCD AUTHORITY and SCHEME

### LINKED ISSUE #1422: Don't print ETCD_AUTHORITY in `node start` output if possible
To further help phase out the deprecated `$ETCD_AUTHORITY`, we should avoid printing it in the `calicoctl node run` output when possible, i.e. when user has specified `$ETCD_ENDPOINTS` instead.

For example, I ran 

`[vagrant@calico-mesos-03 ~]$ sudo ETCD_ENDPOINTS=http://172.24.197.101:2379 calicoctl node run --dryrun --init-system`

and I get:

`docker run --net=host --privileged --name=calico-node --rm -e CALICO_NETWORKING_BACKEND=bird -e NO_DEFAULT_POOLS= -e CALICO_LIBNETWORK_ENABLED=true -e CALICO_LIBNETWORK_IFPREFIX=cali -e ETCD_ENDPOINTS=http://172.24.197.101:2379 -e ETCD_AUTHORITY= -e ETCD_SCHEME= -e NODENAME=calico-mesos-03 -v /var/run/calico:/var/run/calico -v /lib/modules:/lib/modules -v /var/log/calico:/var/log/calico -v /run/docker/plugins:/run/docker/plugins -v /var/run/docker.sock:/var/run/docker.sock calico/node:v1.0.0`

No need to point ETCD_AUTHORITY here.

tmjd:
Fixes #1422

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -304,13 +304,9 @@ Description:
 	}
 
 	if etcdcfg.EtcdEndpoints == "" {
-		envs["ETCD_AUTHORITY"] = etcdcfg.EtcdAuthority
-		envs["ETCD_SCHEME"] = etcdcfg.EtcdScheme
-		envs["ETCD_ENDPOINTS"] = ""
+		envs["ETCD_ENDPOINTS"] = etcdcfg.EtcdScheme + "://" + etcdcfg.EtcdAuthority
 	} else {
 		envs["ETCD_ENDPOINTS"] = etcdcfg.EtcdEndpoints
-		envs["ETCD_AUTHORITY"] = ""
-		envs["ETCD_SCHEME"] = ""
 	}
 	if etcdcfg.EtcdCACertFile != "" {
 		envs["ETCD_CA_CERT_FILE"] = ETCD_CA_CERT_NODE_FILE
--- a/tests/st/calicoctl/test_node_run.py
+++ b/tests/st/calicoctl/test_node_run.py
@@ -28,3 +28,4 @@ class TestNodeRun(TestBase):
             output = host.calicoctl("node run --dryrun")
             assert "ETCD_AUTHORITY" not in output
             assert "ETCD_SCHEME" not in output
+            assert "ETCD_ENDPOINTS" in output
--- a/tests/st/calicoctl/test_node_run.py
+++ b/tests/st/calicoctl/test_node_run.py
@@ -1,0 +1,30 @@
+# Copyright (c) 2015-2016 Tigera, Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from tests.st.test_base import TestBase
+from tests.st.utils.docker_host import DockerHost
+
+"""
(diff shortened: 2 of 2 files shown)


calico_node/filesystem/etc/rc.local
```diff
@@ -61,8 +61,7 @@ case "$CALICO_NETWORKING_BACKEND" in
 		confd -onetime -backend=k8s -confdir=/etc/calico/confd -log-level=debug -keep-stage-file >/felix-startup-1.log 2>&1 || true
 		confd -onetime -backend=k8s -confdir=/etc/calico/confd -log-level=debug -keep-stage-file >/felix-startup-2.log 2>&1 || true
 	else
-		# Use ETCD_ENDPOINTS in preferences to ETCD_AUTHORITY
```
robbrockbank: This change isn't necessary - I don't think we are fully retiring ETCD_SCHEME and ETCD_AUTHORITY at this point - just that we don't want to advertise it in the CLI output.

calico_node/filesystem/etc/service/available/confd/run
```diff
@@ -5,7 +5,7 @@ if [ "$DATASTORE_TYPE" = "kubernetes" ]
 then
     exec confd -confdir=/etc/calico/confd -interval=5 -log-level=debug -backend=k8s
 else
-    ETCD_NODE=${ETCD_ENDPOINTS:=${ETCD_SCHEME:=http}://${ETCD_AUTHORITY}}
```
robbrockbank: Same here.

calico_node/filesystem/etc/service/available/felix/run
```diff
@@ -5,9 +5,7 @@ exec 2>&1
 if [ ! -z $NODENAME ]; then
     export FELIX_FELIXHOSTNAME=$NODENAME
 fi
-export FELIX_ETCDADDR=$ETCD_AUTHORITY
```
robbrockbank: Here too.

calicoctl/commands/node/run.go
```diff
@@ -303,15 +303,7 @@ Description:
 			vol{hostPath: "/var/run/docker.sock", containerPath: "/var/run/docker.sock"})
 	}
 
-	if etcdcfg.EtcdEndpoints == "" {
```
robbrockbank: For the CLI output I think we should convert the authority and scheme to an endpoint.  This means the output from the CLI will *never* show the ETCD_AUTHORITY environment.  If a user has already constructed their `docker run` incantation using ETCD_AUTHORITY then that will still work - but we shouldn't advertise that when creating a new command using `calicoctl node run`

tmjd:
Sorry @robbrockbank I misunderstood 'drop' in the original issue.  I'll fix this.  Thanks for taking a look.

tests/st/calicoctl/test_node_run.py
```diff
@@ -0,0 +1,30 @@
...
+
+class TestNodeRun(TestBase):
+    def test_node_run_dryrun(self):
+        """
+        Test that dryrun does not output ETCD_AUTHORITY or ETCD_SCHEME.
+        """
+        with DockerHost('host', dind=False, start_calico=False) as host:
+            output = host.calicoctl("node run --dryrun")
+            assert "ETCD_AUTHORITY" not in output
+            assert "ETCD_SCHEME" not in output
```
robbrockbank: I think we should check that ETCD_ENDPOINTS is in the output too - just to validate out assertion technique.

robbrockbank:
One more comment and then LGTM

tests/st/calicoctl/test_node_run.py
tmjd: I actually did put `assert "ETCD_ENDPOINTS" not in output` to verify that it worked as a test, which caused a failure.  I'll put in a positive test for it though.

tmjd:
@robbrockbank There you go sir :grinning:

robbrockbank:
LGTM

PR #1604, opened 2017-04-12
Ensure run dir exists and log if not

### LINKED ISSUE #1123: Bird fails startup if /var/run/calico not volume mounted
I'm using `docker run calico/node` to start node (instead of using `calicoctl`). I accidentally did not volume mount in `/var/run/calico` from the host, which means that `/var/run/calico` doesn't exist in the container. As a result, bird repeatedly fails with `2016-08-19_21:06:16.72566 bird: Cannot create control socket /var/run/calico/bird.ctl: No such file or directory`.

This was not a very intuitive problem to diagnose. We should make this more user friendly. I'll suggest two possible fixes:
1. create an empty `/var/run/calico` during the Docker build no matter what, so bird won't fail startup.
2. Add checks during startup.py to better handle if `/var/run/calico` does not exist.

tmjd:
Fixes #1123

### DIFF
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -79,6 +79,9 @@ func main() {
 	// Check for conflicting node configuration
 	checkConflictingNodes(client, node)
 
+	// Check expected filesystem
+	ensureFilesystemAsExpected()
+
 	// Apply the updated node resource.
 	if _, err := client.Nodes().Apply(node); err != nil {
 		fatal("Unable to set node resource configuration: %s", err)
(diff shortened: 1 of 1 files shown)


tmjd:
I tested this manually by doing the following:

1.  Taking output of `calicoctl node run --dryrun`
2. Removing the `/var/log/calico` mount
3. Running with the current calico/node (v1.1.1)
4. Checking that the error was being logged by bird and the path did not exist in the docker container
5. Built calico/node with my changes
6. Changed the `docker run` command to use my newly built container and ran it.
6. Checked that bird was not logging the error, checked that the path existed in the container, and checked that in `docker logs calico-node` the expected warning message was present.

calico_node/startup/startup.go
```diff
@@ -633,6 +636,20 @@ func checkConflictingNodes(client *client.Client, node *api.Node) {
...
+func ensureFilesystemAsExpected() {
+	runDir := "/var/run/calico"
+	// Check if directory already exists
+	if _, err := os.Stat(runDir); err != nil {
+		// Create the runDir
+		if err = os.MkdirAll(runDir, os.ModeDir); err != nil {
+			fatal("Unable to create '%s'", runDir)
+			terminate()
+		}
+		warning("%s was not mounted, created directory and continuing", runDir)
```
robbrockbank: Rather than saying created directory and continuing - perhaps we should state the impact, i.e. it will not be possible to use `calicoctl node status` from the host.

tmjd:
@robbrockbank back to you. Also added checking `/var/log/calico`.

robbrockbank:
LGTM

PR #1606, opened 2017-04-14
In calico/ctl put binary in PATH

### LINKED ISSUE #1525: PATH problems with calicoctl binary
In the image `calico/ctl`, the binary is placed at the top level directory(`/`). But this directory is not in the PATH variable.

```
/ # echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

So when running a container with this image, I would have to prepend `./` when running the command.

```
/ # ./calicoctl get nodes
NAME
calico1
calico2
calico3
```

Problem:
```
/ # calicoctl get nodes
sh: calicoctl: not found
/ #
```

Expected:
```
/ # calicoctl get nodes
NAME
calico1
calico2
calico3
```

Either the binary can be copied into `/usr/bin/calicoctl` or when building the docker image, the top level directory needs to be added to the path.

tmjd:
Fixes #1525

### DIFF
--- a/calicoctl/Dockerfile.calicoctl
+++ b/calicoctl/Dockerfile.calicoctl
@@ -4,6 +4,7 @@ MAINTAINER Tom Denham <tom@projectcalico.org>
 ADD dist/calicoctl ./calicoctl
 
 ENV CALICO_CTL_CONTAINER=TRUE
+ENV PATH=$PATH:/
 
 # glibc and libltdl are needed by docker command line tool
 ENV GLIBC_VERSION 2.23-r3


robbrockbank:
One concern here is that people may have already set up their scripts to access calicoctl in the current location in the container.

Is it possible to achieve the same effect by updating the PATH instead?  That way it won't break any one that assumes the binary is where it currently lives?

robbrockbank:
LGTM.  Just waiting on the green light.

PR #1607, opened 2017-04-18
Bumping versions

heschlie:
- Bumped libcalico-go
- Bumped confd
- Bumped felix

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -21,7 +21,7 @@ import:
   - json
 - package: github.com/projectcalico/go-yaml-wrapper
 - package: github.com/projectcalico/libcalico-go
-  version: 1.2.0
+  version: v1.2.1-rc1
   subpackages:
   - lib/api
   - lib/api/unversioned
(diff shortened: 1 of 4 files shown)


glide.yaml
```diff
@@ -21,7 +21,7 @@ import:
   - json
 - package: github.com/projectcalico/go-yaml-wrapper
 - package: github.com/projectcalico/libcalico-go
-  version: 1.2.0
+  version: 1.2.1-rc1
```
gunjan5: this should be `v1.2.1-rc1`, missing the `v` (glide might be smart enough to figure it out I guess, but we should make is explicit)

heschlie: gah so close

robbrockbank: Yeah I'm pretty sure glide does figure this out - I don't think we usually put the v's in but i' might be mistaken.

gunjan5:
LGTM, squash and merge away!

PR #1622, opened 2017-05-04
Use alternative IP pool if overlaps with etcd (for IPIP tests)

### LINKED ISSUE #1630: Intermittent failures in the IPIP STs.
Current suspicion is that etcd is in the same subnet as the IPIP IP pool and therefore routing to etcd will be out of whack (attempt to go over tunnel)

robbrockbank:
Looks like we have some intermittent ipip failures which I believe are caused by the etcd IP address being on the same CIDR as the IP pool that has IPIP enabled on it.

If I understand correctly this sometimes causes transient etcd issues as it's possible that rules are in place on the host to route etcd traffic over the IPIP tunnel (which will never work).

I regard the fix in this PR as a temporary workaround - a better fix would be to allow the default IP Pool CIDR in our tests to be configurable.  However, that is pending another calicoctl PR to allow the CIDR to be passed into the node container through `calicoctl node run`

Fixes #1630


tests/st/bgp/test_ipip.py
```diff
@@ -369,3 +376,15 @@ def test_gce(self, with_ipip, backend='bird'):
                 # Flip the IP-in-IP state for the next iteration.
                 with_ipip = not with_ipip
                 host1.set_ipip_enabled(with_ipip)
+
+    def get_pool_cidr(self):
```
tmjd: Should this function be relocated to utils.utils?

PR #1623, opened 2017-05-04
Log to stdout in calico/node startup [WIP]

robbrockbank:


calico_node/allocateipip/allocate_ipip_addr.go
```diff
@@ -17,6 +17,15 @@ import (
 // that is configured.
 
 func main() {
+	// Set logging to info and to stdout.
+	log.SetLevel(log.InfoLevel)
```
caseydavenport: We should make this configurable

calico_node/allocateipip/allocate_ipip_addr.go
```diff
@@ -17,6 +17,15 @@ import (
...
+	log.SetOutput(os.Stdout)
+	log.SetFormatter(&log.JSONFormatter{
```
caseydavenport: Do we have a standard format we can use? Grab the one from Felix?

calico_node/startup/startup.go
```diff
@@ -43,6 +43,9 @@ const (
 // For testing purposes we define an exit function that we can override.
 var exitFunction = os.Exit
 
+// Prefix all logs that we output to in `calicotl node run` with a specific string.
+var calicoctlLogPrefix = "progress: "
```
caseydavenport: Hm.  Can we decide which logs to send to screen using log level perhaps? INFO and above get sent to screen via calicoctl?  Would be easy to then say, send DEBUG logs to screen for a `--verbose` option or something.

I'd also like to remove references to `calicoctl` within the `calico/node` image to avoid tying the two together unnecessarily.  At a minimum can we rename this variable to something else like `progressLogPrefix`?

calico_node/startup/startup.go
```diff
@@ -749,19 +761,19 @@ func ensureGlobalBGPConfig(c *client.Client, key, def string) error {
...
 func message(format string, args ...interface{}) {
-	fmt.Printf(format+"\n", args...)
+	log.Infof(calicoctlLogPrefix+format, args...)
 }
 
 // warning prints a warning to screen and to log.  A newline terminator is
 // not required in the format string.
 func warning(format string, args ...interface{}) {
-	fmt.Printf("WARNING: "+format+"\n", args...)
+	log.Warningf(calicoctlLogPrefix+"WARNING: "+format, args...)
```
caseydavenport: Do we still want the `WARNING` bit produced here? I would have thought calicoctl could add that if it wants to, but it's not necessary here now that we're using `log.Warningf`

calico_node/filesystem/etc/rc.local
```diff
@@ -1,3 +1,8 @@
+# Log an info level log to stdout (JSON format)
+log() {
+	echo '{"msg":"progress: '$1'","severity":"info","time":"'$(date -Iseconds -u)'Z"}'
```
caseydavenport: Is json the right thing to emit? Shouldn't this mirror the output format we use for `log.Infof`?

Issue #1580, opened 2017-03-24
Support disabling NAT in default ip pool


caseydavenport:
NATOutgoing is hard-coded to be enabled in the default pool: https://github.com/projectcalico/calicoctl/blob/master/calico_node/startup/startup.go#L564

This is the right default, but some users may want to disable it.

I think we should have a `CALICO_NAT_OUTGOING` variable or similar to control this.

@tmjd WDYT? Was there a reason we didn't add this in v2.1?

tmjd:
No specific reason, I was just keeping the changes to the minimum I needed to do at the time.  I totally agree we should have it.

tmjd:
There is a docs issue for this too https://github.com/projectcalico/calico/issues/516

VincentS:
Should it be possible to be able to differentiate between ipv4 and ipv6 pool? Meaning  that we have an CALICO_NAT_OUTGOING_IPV6 and CALICO_NAT_OUTGOING_IPV4 variable or is one CALICO_NAT_OUTGOING for both pools enough?

tmjd:
Yes I would imagine NAT_OUTGOING should have an IPV4 and IPV6 variable like the other CALICO_IPV*POOL_* variables for configuring default pools.  @caseydavenport is there any reason you would expect otherwise?

VincentS:
My current pull request only introduces the parameter CALICO_NAT_OUTGOING which sets `nat-outgoing`for both IPv4 and IPv6 default IPPools.

VincentS:
This issue should be solved by https://github.com/projectcalico/calico/pull/861

PR #1471, opened 2017-01-27
Configure default pools through Env Var

tmjd:
This PR is addressing Issue #1445 
Should I be adding a flag to calicoctl to set the default pools too?

PR #1640, opened 2017-05-10
Added parameter CALICO_NAT_OUTGOING for default IPPool Creation

VincentS:
This patch relates to the issue: #1580

It adds the parameter to CALICO_NAT_OUTGOING for the default IPPool creation.
This enables to disable nat-outgoing in the default IPv4 and IPv6 IPPools created.

### DIFF
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -342,9 +342,10 @@ func validateIP(ipn *net.IPNet) {
 }
 
 // evaluateENVBool evaluates a passed environment variable
-// Returns an boolean indicating if the parameter was defined
-// and an boolean if the argument was true or false
-// If an log entry will always be written
+// Returns True if the envVar is defined and set to true.
+// Returns False if the envVar is defined and set to false.
+// Returns def in the envVar is not defined.
+// An log entry will always be written
 func evaluateENVBool(envVar string, defaultValue bool) bool {
 	envValue, isEmpty := os.LookupEnv(envVar)
 
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -345,15 +345,21 @@ func validateIP(ipn *net.IPNet) {
 // Returns an boolean indicating if the parameter was defined
 // and an boolean if the argument was true or false
 // If an log entry will always be written
-func evaluateENVBool(envVar string) (bool, bool) {
+func evaluateENVBool(envVar string, defaultValue bool) bool {
 	envValue, isEmpty := os.LookupEnv(envVar)
-	switch strings.ToLower(envValue) {
+
 	log.Info("%s is %s through environment variable", envVar, envValue)
-	case "false", "0", "no", "n", "f":
-		return isEmpty, false
+	if !isEmpty {
+		switch strings.ToLower(envValue) {
+
+		case "false", "0", "no", "n", "f":
+			return false
+		}
+
+		return true
 	}
 
-	return isEmpty, true
+	return defaultValue
 }
 
 // autoDetectCIDR auto-detects the IP and Network using the requested
@@ -512,19 +518,14 @@ func configureIPPools(client *client.Client) {
 	// Ensure there are pools created for each IP version.
 	if !ipv4Present {
 		log.Debug("Create default IPv4 IP pool")
-		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV4POOL_NAT_OUTGOING")
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, outgoingNATisDefined, outgoingNATEnabled)
+		outgoingNATEnabled := evaluateENVBool("CALICO_IPV4POOL_NAT_OUTGOING", true)
+		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, outgoingNATEnabled)
 	}
 	if !ipv6Present && ipv6Supported() {
 		log.Debug("Create default IPv6 IP pool")
-		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV6POOL_NAT_OUTGOING")
+		outgoingNATEnabled := evaluateENVBool("CALICO_IPV6POOL_NAT_OUTGOING", false)
 
-		// Workaround for NatOutgoing where IPv6 default is false
-		if outgoingNATisDefined == false {
-			outgoingNATEnabled = false
-		}
-
-		createIPPool(client, ipv6Cidr, string(ipip.Undefined), outgoingNATisDefined, outgoingNATEnabled)
+		createIPPool(client, ipv6Cidr, string(ipip.Undefined), outgoingNATEnabled)
 	}
 }
 
@@ -534,7 +535,7 @@ func configureIPPools(client *client.Client) {
 // compiled in will not have this entry).
 func ipv6Supported() bool {
 	// First check if Felix param is false
-	_, outgoingNATEnabled := evaluateENVBool("FELIX_IPV6SUPPORT")
+	outgoingNATEnabled := evaluateENVBool("FELIX_IPV6SUPPORT", false)
 	if !outgoingNATEnabled {
 		return false
 	}
--- a/calico_node/startup/startup.go
+++ b/calico_node/startup/startup.go
@@ -341,6 +341,21 @@ func validateIP(ipn *net.IPNet) {
 	warning("Unable to confirm IPv%d address %s is assigned to this host", ip.Version(), ip)
 }
 
+// evaluateENVBool evaluates a passed environment variable
+// Returns an boolean indicating if the parameter was defined
+// and an boolean if the argument was true or false
+// If an log entry will always be written
+func evaluateENVBool(envVar string) (bool, bool) {
+	envValue, isEmpty := os.LookupEnv(envVar)
+	switch strings.ToLower(envValue) {
+	log.Info("%s is %s through environment variable", envVar, envValue)
+	case "false", "0", "no", "n", "f":
+		return isEmpty, false
+	}
+
+	return isEmpty, true
+}
+
 // autoDetectCIDR auto-detects the IP and Network using the requested
 // detection method.
 func autoDetectCIDR(method string, version int) *net.IPNet {
@@ -497,11 +512,19 @@ func configureIPPools(client *client.Client) {
 	// Ensure there are pools created for each IP version.
 	if !ipv4Present {
 		log.Debug("Create default IPv4 IP pool")
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar)
+		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV4POOL_NAT_OUTGOING")
+		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, outgoingNATisDefined, outgoingNATEnabled)
 	}
 	if !ipv6Present && ipv6Supported() {
 		log.Debug("Create default IPv6 IP pool")
-		createIPPool(client, ipv6Cidr, string(ipip.Undefined))
+		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV6POOL_NAT_OUTGOING")
+
+		// Workaround for NatOutgoing where IPv6 default is false
+		if outgoingNATisDefined == false {
+			outgoingNATEnabled = false
+		}
+
+		createIPPool(client, ipv6Cidr, string(ipip.Undefined), outgoingNATisDefined, outgoingNATEnabled)
 	}
 }
 
@@ -510,10 +533,9 @@ func configureIPPools(client *client.Client) {
 // simplistic check of /proc/sys/net/ipv6 (since platforms that do not have IPv6
 // compiled in will not have this entry).
 func ipv6Supported() bool {
-	// First check the Felix parm.
-	switch strings.ToLower(os.Getenv("FELIX_IPV6SUPPORT")) {
-	case "false", "0", "no", "n", "f":
-		log.Info("IPv6 support disabled through environment")
+	// First check if Felix param is false
+	_, outgoingNATEnabled := evaluateENVBool("FELIX_IPV6SUPPORT")
+	if !outgoingNATEnabled {
 		return false
 	}
 
@@ -526,7 +548,7 @@ func ipv6Supported() bool {
 
 // createIPPool creates an IP pool using the specified CIDR.  This
 // method is a no-op if the pool already exists.
-func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
+func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string, isNATOutgoingEnabled bool) {
 	version := cidr.Version()
 	ipipMode := ipip.Mode(ipipModeName)
 
@@ -540,7 +562,7 @@ func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
 			CIDR: *cidr,
 		},
 		Spec: api.IPPoolSpec{
-			NATOutgoing: true,
+			NATOutgoing: isNATOutgoingEnabled,
 			IPIP: &api.IPIPConfiguration{
 				Enabled: ipipMode != ipip.Undefined,
 				Mode:    ipipMode,
--- a/calico_node/startup/startup_test.go
+++ b/calico_node/startup/startup_test.go
@@ -69,7 +69,7 @@ type EnvItem struct {
 }
 
 var _ = Describe("FV tests against a real etcd", func() {
-	changedEnvVars := []string{"CALICO_IPV4POOL_CIDR", "CALICO_IPV6POOL_CIDR", "NO_DEFAULT_POOLS", "CALICO_IPV4POOL_IPIP"}
+	changedEnvVars := []string{"CALICO_IPV4POOL_CIDR", "CALICO_IPV6POOL_CIDR", "NO_DEFAULT_POOLS", "CALICO_IPV4POOL_IPIP", "CALICO_IPV6POOL_NAT_OUTGOING", "CALICO_IPV4POOL_NAT_OUTGOING"}
 
 	BeforeEach(func() {
 		for _, envName := range changedEnvVars {
@@ -83,7 +83,7 @@ var _ = Describe("FV tests against a real etcd", func() {
 	})
 
 	DescribeTable("Test IP pool env variables",
-		func(envList []EnvItem, expectedIPv4 string, expectedIPv6 string, expectIpv4IpipMode string) {
+		func(envList []EnvItem, expectedIPv4 string, expectedIPv6 string, expectIpv4IpipMode string, expectedIPV4NATOutgoing bool, expectedIPV6NATOutgoing bool) {
 			// Create a new client.
 			cfg, _ := client.LoadClientConfigFromEnvironment()
 			c := testutils.CreateCleanClient(*cfg)
@@ -106,6 +106,7 @@ var _ = Describe("FV tests against a real etcd", func() {
 			// Look through the pool for the expected data.
 			foundv4Expected := false
 			foundv6Expected := false
+
 			for _, pool := range poolList.Items {
 				if pool.Metadata.CIDR.String() == expectedIPv4 {
 					foundv4Expected = true
@@ -118,6 +119,13 @@ var _ = Describe("FV tests against a real etcd", func() {
 					if pool.Spec.IPIP != nil {
 						Expect(pool.Spec.IPIP.Enabled).To(BeFalse())
 					}
+
+					if expectedIPV6NATOutgoing {
+						Expect(pool.Spec.NATOutgoing).To(BeTrue())
+					} else {
+						Expect(pool.Spec.NATOutgoing).To(BeFalse())
+					}
+
 				} else {
 					// off is not a real mode value but use it instead of empty string
 					if expectIpv4IpipMode == "off" {
@@ -128,6 +136,13 @@ var _ = Describe("FV tests against a real etcd", func() {
 						Expect(pool.Spec.IPIP.Enabled).To(BeTrue())
 						Expect(pool.Spec.IPIP.Mode).To(Equal(ipip.Mode(expectIpv4IpipMode)))
 					}
+
+					if expectedIPV4NATOutgoing == false {
+						Expect(pool.Spec.NATOutgoing).To(BeFalse())
+					} else {
+						Expect(pool.Spec.NATOutgoing).To(BeTrue())
+					}
+
 				}
 			}
 			Expect(foundv4Expected).To(BeTrue(),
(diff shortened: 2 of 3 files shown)


tmjd:
This could use some tests too.  Take a look at #1471 for the tests that were added for the other default pool configuration variables.

calico_node/startup/startup.go
```diff
@@ -524,11 +526,26 @@ func ipv6Supported() bool {
 	return supported
 }
 
+// defineNATOutgoing checks the CALICO_NAT_OUTGOING parameter used for
+// default IPPool Creation. It returns a boolean.
+
+func defineNATOutgoing () bool {
+	switch strings.ToLower(os.Getenv("CALICO_NAT_OUTGOING")) {
```
tmjd: I think this name should be `CALICO_IPPOOL_NAT_OUTGOING` if there is only one variable for both IPv4 and IPv6.  If, as I suspect, that we should have 2 then it should match the format of the other default pool configuration variables.

calico_node/startup/startup.go
```diff
@@ -524,11 +526,26 @@ func ipv6Supported() bool {
 	return supported
 }
 
+// defineNATOutgoing checks the CALICO_NAT_OUTGOING parameter used for
+// default IPPool Creation. It returns a boolean.
+
+func defineNATOutgoing () bool {
```
tmjd: I think this should be named something like `isNATOutgoingEnabled`

VincentS:
Refactored the code based on the feedback. I introduced a new helper function called `evaluateENVBool` to prevent code duplication (2 - 3 identical almost switch statements). 
With this refactoring it would be quite simple to separate NAT_OUTGOING configuration for IPv4 and IPv6.

calico_node/startup/startup.go
```diff
@@ -341,6 +341,21 @@ func validateIP(ipn *net.IPNet) {
 	warning("Unable to confirm IPv%d address %s is assigned to this host", ip.Version(), ip)
 }
 
+
+// evaluateENVBool evaluates a passed environment variable and writes an
+// log entry if false otherwise it returns true by default
+
```
caseydavenport: nit: could you remove the extra newline here?

calico_node/startup/startup.go
```diff
@@ -450,6 +465,7 @@ func configureIPPools(client *client.Client) {
 		return
 	}
 
+
```
caseydavenport: nit: and this one too :)

calico_node/startup/startup.go
```diff
@@ -341,6 +341,21 @@ func validateIP(ipn *net.IPNet) {
...
+func evaluateENVBool(envvar string, logentry string) bool {
+	switch strings.ToLower(os.Getenv(envvar)) {
+	case "false", "0", "no", "n", "f":
+		log.Info(logentry)
+		return false
+	}
+
+	return true
+
```
caseydavenport: nit: extra newline

calico_node/startup/startup.go
```diff
@@ -497,23 +513,22 @@ func configureIPPools(client *client.Client) {
 	// Ensure there are pools created for each IP version.
 	if !ipv4Present {
 		log.Debug("Create default IPv4 IP pool")
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar)
+		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, evaluateENVBool("CALICO_IPPOOL_NAT_OUTGOING","NAT_Outgoing disabled for Default IPv4 Pool."))
 	}
 	if !ipv6Present && ipv6Supported() {
 		log.Debug("Create default IPv6 IP pool")
-		createIPPool(client, ipv6Cidr, string(ipip.Undefined))
+		createIPPool(client, ipv6Cidr, string(ipip.Undefined), evaluateENVBool("CALICO_IPPOOL_NAT_OUTGOING","NAT_Outgoing disabled for Default IPv6 Pool."))
```
caseydavenport: @fasaxc does Felix actually support outgoing NAT on ipv6 pools?

calico_node/startup/startup.go
```diff
@@ -497,23 +513,22 @@ func configureIPPools(client *client.Client) {
...
 // a check on the appropriate Felix parameter and if supported also performs a
 // simplistic check of /proc/sys/net/ipv6 (since platforms that do not have IPv6
 // compiled in will not have this entry).
 func ipv6Supported() bool {
-	// First check the Felix parm.
-	switch strings.ToLower(os.Getenv("FELIX_IPV6SUPPORT")) {
-	case "false", "0", "no", "n", "f":
-		log.Info("IPv6 support disabled through environment")
+	// First check if Felix param is false
+	if !evaluateENVBool("FELIX_IPV6SUPPORT","IPv6 support disabled through environment") {
```
caseydavenport: nit: could you add spaces after each argument?

e.g.

```
("FELIX_IPV6SUPPORT", "IPv6 support disabled through environment")
```

calico_node/startup/startup.go
VincentS: This is still TBD

VincentS:
I made the changes based on the feedback by @caseydavenport. Test will be added soon.

calico_node/startup/startup.go
```diff
@@ -497,23 +509,22 @@ func configureIPPools(client *client.Client) {
 	// Ensure there are pools created for each IP version.
 	if !ipv4Present {
 		log.Debug("Create default IPv4 IP pool")
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar)
+		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, evaluateENVBool("CALICO_IPPOOL_NAT_OUTGOING", "NAT_Outgoing disabled for Default IPv4 Pool."))
```
tmjd: Please change the environment variable names to be IP version specific.
ie. CALICO_IPV4POOL_NAT_OUTGOING

tmjd:
@caseydavenport Can you confirm that we do want a separate IPv4 and IPv6 env var for IPIP?

caseydavenport:
@tmjd @VincentS sorry, yes I think we do (although NAT outgoing on IPv6 pools is probably not generally something people _should_ be doing, so that flag should default to `False`)

tmjd:
@VincentS It has been a few days and I wanted to check in with you, is there anything I can help with on this PR or an idea of when you may have more updates?

VincentS:
Hey I havent been in the office last week.

VincentS:
I adjusted and rebased/squashed the commits of the PR. Currently Tests are still WIP since some of the new are failing at the moment.

calico_node/startup/startup.go
```diff
@@ -526,10 +544,16 @@ func ipv6Supported() bool {
...
 // createIPPool creates an IP pool using the specified CIDR.  This
 // method is a no-op if the pool already exists.
-func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
+func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string, isNATOutgoingDefined bool, isNATOutgoingEnabled bool) {
 	version := cidr.Version()
 	ipipMode := ipip.Mode(ipipModeName)
 
+
+	//Workaround for NatOutgoing where IPv6 default is false
+	if(version == 6 && isNATOutgoingDefined == false) {
```
tmjd: I think this should be placed in configureIPPools, there is already an IPv6 section there and then it won't be necessary to pass isNATOutgoingDefined into createIPPool.

calico_node/startup/startup.go
```diff
@@ -341,6 +341,22 @@ func validateIP(ipn *net.IPNet) {
 	warning("Unable to confirm IPv%d address %s is assigned to this host", ip.Version(), ip)
 }
 
+// evaluateENVBool evaluates a passed environment variable
+// Returns an boolean indicating if the parameter was defined
+// and an boolean if the argument was true or false
+// If it was false an additional log entry will be written
+
+func evaluateENVBool(envvar string, logentry string) (isDefined bool, isValue bool ) {
```
tmjd: I think there could be a better name for the 'isValue' variable.  Maybe envValue or possibly just value, or the extra verbose option envBoolValue.

calico_node/startup/startup.go
```diff
@@ -497,23 +513,25 @@ func configureIPPools(client *client.Client) {
 	// Ensure there are pools created for each IP version.
 	if !ipv4Present {
 		log.Debug("Create default IPv4 IP pool")
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar)
+		OutgoingNATisDefined, OutgoingNATisValue := evaluateENVBool("CALICO_IPV4POOL_NAT_OUTGOING", "NAT_Outgoing disabled for Default IPv4 Pool.")
```
tmjd: With other suggested changes I think OutgoingNATisValue should just be outgoingNATEnabled.

calico_node/startup/startup.go
```diff
@@ -497,23 +513,25 @@ func configureIPPools(client *client.Client) {
...
 // a check on the appropriate Felix parameter and if supported also performs a
 // simplistic check of /proc/sys/net/ipv6 (since platforms that do not have IPv6
 // compiled in will not have this entry).
 func ipv6Supported() bool {
-	// First check the Felix parm.
-	switch strings.ToLower(os.Getenv("FELIX_IPV6SUPPORT")) {
-	case "false", "0", "no", "n", "f":
-		log.Info("IPv6 support disabled through environment")
+	// First check if Felix param is false
+	_, isValue := evaluateENVBool("FELIX_IPV6SUPPORT", "IPv6 support disabled through environment");
```
tmjd: I think whatever you change isValue to in evaluateENVBool you can probably use that name here too.

calico_node/startup/startup_test.go
```diff
@@ -137,31 +152,49 @@ var _ = Describe("FV tests against a real etcd", func() {
...
 			[]EnvItem{
 				{"CALICO_IPV6POOL_CIDR", "fdff:ffff:ffff:ffff:ffff::/80"},
 				{"CALICO_IPV4POOL_IPIP", "always"},
 			},
-			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "always"),
+			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "always", true, false),
+		Entry("IPv6 NATOutgoing Set Enabled",
+			[]EnvItem{
+				{"CALICO_IPV6POOL_NAT_OUTGOING", "true"}},
+			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "off", true, true),
```
tmjd: You need to use the default IPv6 pool here. `fd80:24e2:f998:72d6::/64` like in the first test entry

calico_node/startup/startup_test.go
```diff
@@ -137,31 +152,49 @@ var _ = Describe("FV tests against a real etcd", func() {
...
+		Entry("IPv6 NATOutgoing Set Disabled",
+			[]EnvItem{
+				{"CALICO_IPV6POOL_NAT_OUTGOING", "false"}},
+			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "off", true, false),
```
tmjd: Same with the IPv6 default here and for the 2 test Entries below too

calico_node/startup/startup_test.go
```diff
@@ -137,31 +152,49 @@ var _ = Describe("FV tests against a real etcd", func() {
...
+		Entry("IPv4 NATOutgoing Set Disabled",
+			[]EnvItem{
+				{"CALICO_IPV4POOL_NAT_OUTGOING", "false"}},
+			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "off", false, false),
+		Entry("IPv6 NAT OUTGOING and IPV4 NAT OUTGOING SET",
+			[]EnvItem{
+				{"CALICO_IPV4POOL_NAT_OUTGOING", "false"},
+				{"CALICO_IPV6POOL_NAT_OUTGOING", "true"},
+			},
+			"192.168.0.0/16", "fdff:ffff:ffff:ffff:ffff::/80", "always", false, true),
```
tmjd: `"always"` will require having CALICO_IPV4POOL_IPIP set.

VincentS:
Fixed all remaining issues and squashed / rebase commit.

calico_node/startup/startup.go
```diff
@@ -341,6 +341,22 @@ func validateIP(ipn *net.IPNet) {
 	warning("Unable to confirm IPv%d address %s is assigned to this host", ip.Version(), ip)
 }
 
+// evaluateENVBool evaluates a passed environment variable
+// Returns an boolean indicating if the parameter was defined
+// and an boolean if the argument was true or false
+// If it was false an additional log entry will be written
+
```
caseydavenport: Could you remove this extra blank line here?

calico_node/startup/startup.go
```diff
@@ -341,6 +341,22 @@ func validateIP(ipn *net.IPNet) {
...
+func evaluateENVBool(envvar string, logentry string) (isDefined bool, outgoingNATEnabled bool) {
```
caseydavenport: I think the return values here shouldn't be named unless they're being used within the function.

Generally, we don't use named returns in Calico code, so I think it would be best to remove the names and ensure the code is commented with correct info.

calico_node/startup/startup.go
```diff
@@ -341,6 +341,22 @@ func validateIP(ipn *net.IPNet) {
...
+	envValue, isEmpty := os.LookupEnv(envvar)
+	switch strings.ToLower(envValue) {
+	case "false", "0", "no", "n", "f":
+		log.Info(logentry)
```
caseydavenport: I think the logging would be cleaner if instead of passing in the log entry, we just logged something like this:

```
log.Info("%s (%s) is disabled through environment variable", envVar, envValue)
```

Then we won't need to pass in the logs throughout, and all the logs will be guaranteed to have the same format.

calico_node/startup/startup.go
caseydavenport: nit: `envvar` should be named `envVar`

calico_node/startup/startup.go
```diff
@@ -497,11 +513,19 @@ func configureIPPools(client *client.Client) {
...
-		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar)
+		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV4POOL_NAT_OUTGOING", "NAT_Outgoing disabled for Default IPv4 Pool.")
+		createIPPool(client, ipv4Cidr, ipv4IpipModeEnvVar, outgoingNATisDefined, outgoingNATEnabled)
 	}
 	if !ipv6Present && ipv6Supported() {
 		log.Debug("Create default IPv6 IP pool")
-		createIPPool(client, ipv6Cidr, string(ipip.Undefined))
+		outgoingNATisDefined, outgoingNATEnabled := evaluateENVBool("CALICO_IPV6POOL_NAT_OUTGOING", "NAT_Outgoing disabled for Default IPv6 Pool.")
+
+		//Workaround for NatOutgoing where IPv6 default is false
```
caseydavenport: nit: needs a space after the slashes. `// Workaround`

calico_node/startup/startup.go
```diff
@@ -341,6 +341,22 @@ func validateIP(ipn *net.IPNet) {
...
+
+func evaluateENVBool(envvar string, logentry string) (isDefined bool, outgoingNATEnabled bool) {
+	envValue, isEmpty := os.LookupEnv(envvar)
+	switch strings.ToLower(envValue) {
+	case "false", "0", "no", "n", "f":
+		log.Info(logentry)
+		return isEmpty, false
+	}
+
+	return isEmpty, true
```
caseydavenport: I think we should tweak this just a little bit.  IIUC, right now the code will return `true` even if the environment variable isn't defined.  I think a nicer interface would be that a default is passed in, and if the environment variable isn't defined the default is returned.

e.g.

```
// Returns True if the envVar is defined and set to true.
// Returns False if the envVar is defined and set to false.
// Returns def in the envVar is not defined.
func evaluateENVBool(envVar string, def bool) (bool, bool)
```

WDYT?

It would remove the need for this bit of code below:

```
if outgoingNATisDefined == false {
    outgoingNATEnabled = false
}
```

calico_node/startup/startup.go
```diff
@@ -526,7 +549,7 @@ func ipv6Supported() bool {
 
 // createIPPool creates an IP pool using the specified CIDR.  This
 // method is a no-op if the pool already exists.
-func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
+func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string, isNATOutgoingDefined bool, isNATOutgoingEnabled bool) {
```
tmjd: isNATOutgoingDefined is not used in the function anymore, please remove it.

calico_node/startup/startup.go
tmjd: I think we should always log when the environment variable is set.  @caseydavenport ??

calico_node/startup/startup.go
tmjd: @caseydavenport Your example still shows 2 returned values though now we could get by with just one, do you think the isDefined return value should remain? I don't think so.

I had considered suggesting a change like this but had thought it allowed more flexibility because then the calling function could take a specific action if it was not defined.  That flexibility isn't really necessary here so I'm good with the change.  Plus with Casey's suggestion the feature to return if it is defined or not could always be added back.

calico_node/startup/startup_test.go
```diff
@@ -128,6 +136,13 @@ var _ = Describe("FV tests against a real etcd", func() {
 						Expect(pool.Spec.IPIP.Enabled).To(BeTrue())
 						Expect(pool.Spec.IPIP.Mode).To(Equal(ipip.Mode(expectIpv4IpipMode)))
 					}
+
+					if expectedIPV4NATOutgoing == false {
```
tmjd: I think you should have the same logic as the IPV6 and not check that it `== false` and switch the 2 cases below.

calico_node/startup/startup.go
caseydavenport: > do you think the isDefined return value should remain? I don't think so.

Yeah, I missed that - we don't need the isDefined return value if we take this approach.  If we need an isDefined down the road for some other bit of code, we can add that (or that code can check it explicitly).

tmjd:
I just looked at the failing tests and it looks like there are some of the python tests that were checking the defaults on IPV6 pools are failing because of the change to the default behavior.
Can you fix those please?
@caseydavenport We'll need to ensure we have a note about the change in default behavior in the release notes.  How do we do that?

tmjd:
@caseydavenport Do you think it is useful to add a new parameterized python test that uses these new env vars?  Kind of overkill but it is probably just a copy paste and a few modifications.

calico_node/startup/startup_test.go
tmjd: I think it would be better to replace this if statement and the nested checks with
`Expect(pool.Spec.NATOutgoing).To(Equal(expectedIPv6NATOutgoing))`

calico_node/startup/startup_test.go
```diff
@@ -118,6 +119,13 @@ var _ = Describe("FV tests against a real etcd", func() {
 					if pool.Spec.IPIP != nil {
 						Expect(pool.Spec.IPIP.Enabled).To(BeFalse())
 					}
+
+					if expectedIPV6NATOutgoing {
```
tmjd: I think it would be better to replace this if statement and the nested checks with
`Expect(pool.Spec.NATOutgoing).To(Equal(expectedIPv4NATOutgoing))`

caseydavenport:
@tmjd @VincentS I'd really like to get this guy merged if we can, what's left to do here?

VincentS:
Sorry, some other things came up and I didnt had time to finish it up. I begun to do some of the requested changes and hope to get the rest done by this week.

calico_node/startup/startup.go
VincentS: Not sure If I understood your suggestion correctly in my current PR.

tmjd: @VincentS and I chatted in slack to clarify the request.

VincentS:
All changes were incorporated based on the feedback.

calico_node/startup/startup.go
```diff
@@ -510,10 +535,9 @@ func configureIPPools(client *client.Client) {
 // simplistic check of /proc/sys/net/ipv6 (since platforms that do not have IPv6
 // compiled in will not have this entry).
 func ipv6Supported() bool {
-	// First check the Felix parm.
-	switch strings.ToLower(os.Getenv("FELIX_IPV6SUPPORT")) {
-	case "false", "0", "no", "n", "f":
-		log.Info("IPv6 support disabled through environment")
+	// First check if Felix param is false
+	outgoingNATEnabled := evaluateENVBool("FELIX_IPV6SUPPORT", false)
```
tmjd: This outgoingNATEnabled is not an appropriate name.
I don't think the default was false previously.  If FELIX_IPV6SUPPORT was not set then I think the value would have been the empty string which would not match one of those cases.

calico_node/startup/startup.go
```diff
@@ -341,6 +341,28 @@ func validateIP(ipn *net.IPNet) {
...
 
+// evaluateENVBool evaluates a passed environment variable
+// Returns True if the envVar is defined and set to true.
+// Returns False if the envVar is defined and set to false.
+// Returns def in the envVar is not defined.
+// An log entry will always be written
+func evaluateENVBool(envVar string, defaultValue bool) bool {
+	envValue, isEmpty := os.LookupEnv(envVar)
+
+	log.Info("%s is %s through environment variable", envVar, envValue)
```
tmjd: I think you should log the value that this function is returning, that way if the user is setting the value incorrectly it will be obvious if the variable is not being interpreted as they expect.
I think it would be nice to also include "(defaulted)" in the log message if the default value is being used.

You need to use `log.Infof` to make a formatted message, otherwise you get `%s is %s through environment variableCALICO_IPV4POOL_NAT_OUTGOING` as the output

calico_node/startup/startup_test.go
```diff
@@ -118,6 +119,9 @@ var _ = Describe("FV tests against a real etcd", func() {
 					if pool.Spec.IPIP != nil {
 						Expect(pool.Spec.IPIP.Enabled).To(BeFalse())
 					}
+
+					Expect(pool.Spec.NATOutgoing).To(Equal(expectedIPV6NATOutgoing))
```
tmjd: I think you can change that line to this and it will produce a much better error. `Expect(pool.Spec.NATOutgoing).To(Equal(expectedIPV6NATOutgoing), "Expected IPv6 to be %t but was %t", expectedIPV6NATOutgoing, pool.Spec.NATOutgoing)`

calico_node/startup/startup_test.go
```diff
@@ -128,6 +132,9 @@ var _ = Describe("FV tests against a real etcd", func() {
 						Expect(pool.Spec.IPIP.Enabled).To(BeTrue())
 						Expect(pool.Spec.IPIP.Mode).To(Equal(ipip.Mode(expectIpv4IpipMode)))
 					}
+
+					Expect(pool.Spec.NATOutgoing).To(Equal(expectedIPV4NATOutgoing))
```
tmjd: Similar change as with the IPv6 version.

calico_node/startup/startup.go
```diff
@@ -526,7 +550,7 @@ func ipv6Supported() bool {
 
 // createIPPool creates an IP pool using the specified CIDR.  This
 // method is a no-op if the pool already exists.
-func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string) {
+func createIPPool(client *client.Client, cidr *net.IPNet, ipipModeName string, isNATOutgoingEnabled bool) {
```
tmjd: Please update the `message` below that has `with NAT outgoing enabled` to use the passed in isNATOutgoingEnabled.  You can use %t to convert a bool into the string `true` or `false`.

VincentS:
Close PR because it was moved to the Calico Repository https://github.com/projectcalico/calico/pull/861

PR #1649, opened 2017-05-23
Trace out docker logs if calicoctl node run fails

robbrockbank:
Trace out additional diags if `calicoctl node run` fails.

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -17,15 +17,12 @@ package node
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
 	gonet "net"
 	"os"
 	"os/exec"
-	"strings"
-
 	"regexp"
-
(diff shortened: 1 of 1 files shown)


tmjd:
Tests show a couple exceptions like

```
Running command on host1
  - Command: head -100 /var/log/calico/felix/current
Final command: docker exec -it host1 sh -c 'head -100 /var/log/calico/felix/current'
- return: 1
Hit exception getting logs from host1 - skip logs
  File "/code/tests/st/utils/log_analyzer.py", line 219, in _parse_logs
    for line in self.host.execute_readline(cmd):
  File "/code/tests/st/utils/docker_host.py", line 217, in execute_readline
    (command, status))
Exception: Command docker exec -it host1 sh -c 'head -100 /var/log/calico/felix/current' returned non-zero exit code 1
First log has timestamp: None
```

PR #1650, opened 2017-05-24
Makefile: Free even more diskspace on semaphore

tomdee:
We really shouldn't need >15GB to run our tests, but we do...


tmjd:
LTGM

PR #1651, opened 2017-05-25
Only create directories if they are needed

caseydavenport:
## Description

Fix automatic directory creation so that it's only performed when necessary.  This removes unnecessary warning logs which occur when these features are disabled.

## Todos
- [x] Tests (manual)
- [x] Documentation not required


caseydavenport:
I do intend to!

caseydavenport:
I've run through manual tests on Kubernetes with and without the directories mounted, and with/without the environment variables set and things seem to be behaving as expected.

The existing GoBGP STs cover the NETWORKING_BACKEND code path, and the DISABLE_FILE_LOGGING seems harder to implement than I think is worth...

PR #1659, opened 2017-06-06
Add licenses to calico/node 

gunjan5:


caseydavenport:
I think given we're including the specific license files we can remove the more generic ones now.

gunjan5:
@caseydavenport we only have the licenses included for the one's that are not a 100% match with the standard license they use, most others still use the standard licenses 100% so I think we should still keep them for that reason

caseydavenport:
@gunjan5 you've got a failing test

PR #1660, opened 2017-06-07
Fix coreos/etcd pin

caseydavenport:


gunjan5:
This built fine for me.
I'm guessing STs will fail because of that bug, if we are 100% sure it's that bug and shouldn't affect us then this should be ok. 
And we should catch that in QA as well

TL;DR: LGTM

caseydavenport:
> I'm guessing STs will fail because of that bug

I think this is a pin from before that was introduce.

gunjan5:
ah ok, I thought you meant back to what we had before the last commit.

PR #1662, opened 2017-06-08
calico/node: Dockerfile, use latest alpine

tomdee:
Better security to track latest


ozdanborne:
Is there a newer alpine release that doesn't have the vulernability? Or are we planning to pin to one soon as it comes out?

tomdee:
The fix has been ported to other alpine versions, but I think it's worth us tracking the latest.

PR #1664, opened 2017-06-09
calico/node: Makefile, always pull alpine

tomdee:
To ensure that latest alpine release is used


caseydavenport:
Downside is that you can't build from a pre-pulled image called `latest`, but I think that's fine... LGTM

ozdanborne:
I thought this was possible with `docker build --pull`

PR #1670, opened 2017-06-20
Enhance calicoctl error status with cfg path when cfg file is malformed

bcreane:
## Description
Add path to config file to the error message when we can't parse the config file.
- fixes issue: https://github.com/projectcalico/calicoctl/issues/1635
- ran ut's and test-containerized successfully

## Todos
- [x] Tests
- [x] Documentation


calicoctl/commands/clientmgr/client.go
```diff
@@ -35,6 +36,7 @@ func NewClient(cf string) (*client.Client, error) {
 
 	c, err := client.New(*cfg)
 	if err != nil {
+		err = fmt.Errorf("Syntax error in %s: %s", cf, err)
```
caseydavenport: I think we should just combine these into a single line:

```
return nil, fmt.Errorf("Syntax error in %s: %s", cf, err)
```

caseydavenport:
One comment but then LGTM.

bcreane:
crap, screwed up the commit log. I did squash, but got a conflict and then introduced the extra commits. please hold off and I'll try to squash again.

calicoctl/commands/config.go
```diff
@@ -108,7 +108,7 @@ The table below details the valid config options.
 	cf := parsedArgs["--config"].(string)
 	client, err := clientmgr.NewClient(cf)
 	if err != nil {
-		fmt.Println(err)
+		fmt.Printf("Syntax error in %s: %s\n", cf, err)
```
caseydavenport: I don't think this is right - `clientmgr.NewClient` can return errors that aren't syntax errors (e.g. [this branch](https://github.com/projectcalico/calicoctl/pull/1670/files#diff-bd7ad3177a0a66cbe8ca523f2bb2f8efL31)).

I think this should not have the "Syntax error" bit - same for resources.go

calicoctl/commands/clientmgr/client.go
```diff
@@ -35,7 +36,7 @@ func NewClient(cf string) (*client.Client, error) {
 
 	c, err := client.New(*cfg)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Syntax error in %s: %s", cf, err)
```
caseydavenport: I think the general recommendation is to start error strings with a lowercase character so that they are more easily wrapped in the calling code.

This can return `syntax error in calicoctl.cfg: file is not yaml`

And then in can be wrapped in the calling code:

`failed to load config: syntax error in calicoctl.cfg: file is not yaml` or something like that.

caseydavenport:
> please hold off and I'll try to squash again.

oops, didn't see that and put some more comments on :p

calicoctl/commands/clientmgr/client.go
bcreane: sounds good, will do

calicoctl/commands/config.go
bcreane: looks like you're right, I'll rework. thanks for catching.

bcreane:
@caseydavenport  - thanks for your helpful comments. I'm going to close this PR for now and start fresh.

PR #1674, opened 2017-06-22
Remove CircleCI, all tests run through semaphore.

caseydavenport:
## Description

Removes CircleCI, which is currently just a subset of the tests that are running on semaphore.


caseydavenport:
Actually, we still use circle to build artifacts from master, so leaving this for now.

Issue #1555, opened 2017-03-06
Allow configurable interface matching for first-found auto detection.


robbrockbank:
Based on user experience.

caseydavenport:
@robbrockbank rather than continue to add more options here, why don't we support letting users pass in the interface regexes they want to skip / match?

I'm worried that we're going to keep adding these and it's going to break someone who is relying on us finding a, for example, "vtep" interface.  Also worried that it's not scalable for us to change the code every time someone else has another interface they want to ignore..

robbrockbank:
Yup I'm for that too :-)

caseydavenport:
Has this been fixed by https://github.com/projectcalico/calico/pull/885 and https://github.com/projectcalico/calicoctl/pull/1675?

mgleung:
Yup!  I'll add the release notes

PR #1675, opened 2017-06-27
add new interface option for skipping interfaces

mgleung:
## Description
Added an option to skip network interfaces specified by regexes so that users can have first-found logic without having to manually specify system/setup specific interfaces to skip by changing the code.

Is dependent upon https://github.com/projectcalico/calico/pull/878 to function properly

Addresses #1555 

## Testing
Ran this against an image from https://github.com/projectcalico/calico/pull/878 and verified that the specified interfaces were skipped if they matched the regexes given

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -34,12 +34,13 @@ import (
 )
 
 const (
-	ETCD_KEY_NODE_FILE             = "/etc/calico/certs/key.pem"
-	ETCD_CERT_NODE_FILE            = "/etc/calico/certs/cert.crt"
-	ETCD_CA_CERT_NODE_FILE         = "/etc/calico/certs/ca_cert.crt"
-	AUTODETECTION_METHOD_FIRST     = "first-found"
-	AUTODETECTION_METHOD_CAN_REACH = "can-reach="
-	AUTODETECTION_METHOD_INTERFACE = "interface="
+	ETCD_KEY_NODE_FILE                  = "/etc/calico/certs/key.pem"
+	ETCD_CERT_NODE_FILE                 = "/etc/calico/certs/cert.crt"
+	ETCD_CA_CERT_NODE_FILE              = "/etc/calico/certs/ca_cert.crt"
+	AUTODETECTION_METHOD_FIRST          = "first-found"
+	AUTODETECTION_METHOD_CAN_REACH      = "can-reach="
+	AUTODETECTION_METHOD_INTERFACE      = "interface="
+	AUTODETECTION_METHOD_SKIP_INTERFACE = "skip-interface="
 )
 
 var (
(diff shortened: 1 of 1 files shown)


bcreane:
lgtm!

calicoctl/commands/node/run.go
```diff
@@ -531,6 +538,19 @@ func validateIpAutodetectionMethod(method string, version int) {
 			os.Exit(1)
 		}
 		return
+	} else if strings.HasPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE) {
+		// Auto-detection method is "skip-interface", validate that the
+		// interface regexes used are valid golang regexes.
+		ifStr := strings.TrimPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE)
+		// regexes are provided in a string separated by ","
```
caseydavenport: nit: Should have a newline before this comment.
nit: Should capitalize the beginning of a comment.

bcreane:
Hey Matt, suggest you squash the commits, probably down to 1, to make it easy to revert if we need to.

glide.lock
```diff
@@ -1,5 +1,5 @@
 hash: 91a8d9a89a8e34c483b837ccd059cb40ebcea88394a9e6d9360425a749a0c7f2
```
bcreane: is glide.lock change intended?

mgleung: Yeah, its not intended, I was testing some other things and this got updated.  I will revert.

calicoctl/commands/node/run.go
```diff
@@ -526,9 +535,28 @@ func validateIpAutodetectionMethod(method string, version int) {
...
+	} else if strings.HasPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE) {
+		// Auto-detection method is "skip-interface", validate that the
+		// interface regexes used are valid golang regexes.
+		ifStr := strings.TrimPrefix(method, AUTODETECTION_METHOD_SKIP_INTERFACE)
+
+		// Regexes are provided in a string separated by ","
+		ifRegexes := strings.Split(ifStr, ",")
+		for _, ifRegex := range ifRegexes {
+			if _, err := regexp.Compile(ifRegex); err != nil {
+				fmt.Printf("Error executing command: invalid interface skipping regex specified for IP autodetection: %s\n", ifRegex)
```
caseydavenport: nit: can we reword to remove the word "skipping" from this?

I think `invalid interface regex` gets the point across well enough - as is it sounds like it's going to skip the invalid regex, when actually it's just exiting.

mgleung: Sounds good.  I wanted originally to have something in there that was a little more specific as to what exactly was failing, but it does create more issues in plain English...

PR #1677, opened 2017-06-29
Elide redundant calico/node environment variables.

bcreane:
Remove the following calico/node environment variables when running (or dry run of) calico/node:
- NO_DEFAULT_POOLS: only set when nopools is true
- IP[6]_AUTODETECTION_METHOD: calico/node defaults to
  first-found, so only set when the method is not first-found

See issue: https://github.com/projectcalico/calicoctl/issues/1520
Testing: manual with enhanced calico/node debug logging

## Todos
- [x] Tests
- [x] Documentation

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -234,20 +234,11 @@ Description:
 	}
 	etcdcfg := cfg.Spec.EtcdConfig
 
-	// Convert the nopools boolean to either an empty string or "true".
-	noPoolsString := ""
-	if nopools {
-		noPoolsString = "true"
-	}
-
 	// Create a mapping of environment variables to values.
 	envs := map[string]string{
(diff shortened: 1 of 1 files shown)


bcreane:
@tomdee - can you take a look? The PR isn't quite as ambitious as the original issue, for example, I left the ETCD environment variables alone. If you think its worth getting rid of those too, please let me know and I'll go back and revise.

bcreane:
@caseydavenport - can you take a look?

bcreane:
@tomdee, @caseydavenport - any comments? thanks

tomdee:
Looks fine to me. And I did a quick local test and it works. You'll need someone from @projectcalico/calicoctl-maintainers to approve and merge it though.

caseydavenport:
Looks fine to me.

PR #1679, opened 2017-07-06
Report error and exit when non-default config file is missing.

bcreane:
## Description
Print an error and exit when non-default config file is missing. Error only when the user specifies a non-default ("/etc/calico/calicoctl.cfg") config file via --config= option.

Issue: https://github.com/projectcalico/calicoctl/issues/1678

## Testing
* Manually tested all commands with the following scenarios:
  * Valid /etc/calico/calicoctl.cfg file
  * Missing /etc/calico/calicoctl.cfg file
  * Valid non-default config file (e.g. fun.yaml)
  * Missing non-default config file

### DIFF
--- a/calicoctl/commands/clientmgr/client.go
+++ b/calicoctl/commands/clientmgr/client.go
@@ -15,9 +15,11 @@
 package clientmgr
 
 import (
+	"fmt"
 	"os"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/client"
 )
@@ -45,7 +47,11 @@ func NewClient(cf string) (*client.Client, error) {
 // otherwise will load from environment variables.
 func LoadClientConfig(cf string) (*api.CalicoAPIConfig, error) {
 	if _, err := os.Stat(cf); err != nil {
-		log.Infof("Config file cannot be read - reading config from environment")
+		if (cf != constants.DefaultConfigPath) {
+			fmt.Printf("Error reading config file: %s\n", cf)
+			os.Exit(1)
+		}
+		log.Infof("Config file: %s cannot be read - reading config from environment", cf)
 		cf = ""
 	}
 
--- a/calicoctl/commands/apply.go
+++ b/calicoctl/commands/apply.go
@@ -42,7 +42,7 @@ Options:
                             "-" loads from stdin.
   -c --config=<CONFIG>      Path to the file containing connection
                             configuration in YAML or JSON format.
-                            [default: /etc/calico/calicoctl.cfg]
+                            [default: ` + constants.DefaultConfigPath + `]
 
 Description:
   The apply command is used to create or replace a set of resources by filename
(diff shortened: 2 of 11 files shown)


caseydavenport:
I think there are two cases:
- The default config location (/etc/calicoctl.cfg or something like that) is provided but not present on disk.
- A non-default config is provided but not present on disk.

For the first, we want to ignore and continue (maybe a debug log which says it's not there).
For the second, I think we should log.Fatal.

WDYT?

bcreane:
@caseydavenport - a nicely nuanced approach. I'll respin the PR. Thanks for the comments.

bcreane:
@caseydavenport - can I get your eyes on this when you have a few spare minutes? thanks!

calicoctl/commands/clientmgr/client.go
```diff
@@ -45,7 +47,11 @@ func NewClient(cf string) (*client.Client, error) {
 // otherwise will load from environment variables.
 func LoadClientConfig(cf string) (*api.CalicoAPIConfig, error) {
 	if _, err := os.Stat(cf); err != nil {
-		log.Infof("Config file cannot be read - reading config from environment")
+		errMsg := fmt.Sprintf("Config file: %s cannot be read - reading config from environment", cf)
+		if (cf != constants.DefaultConfigPath) {
+			fmt.Fprintln(os.Stderr, errMsg)
```
caseydavenport: I think we want to be more aggressive in this case, no?

Any reason we don't exit 1 when this happens?

bcreane: No reason other than being overly conservative. I'll make the change.

bcreane:
@caseydavenport - the pr is ready for your return review, thanks.

calicoctl/commands/clientmgr/client.go
caseydavenport: I think we shouldn't try to re-use the error message for the failure case and the warning case, since they're really different behaviors.

i.e., right now if if fails to read the file and exists, it still prints "reading config from environment", which is a lie :)

bcreane: good point, left overs from a previous iteration. will respin.

bcreane: @caseydavenport - there's a mix of stdout and stderr currently. what's your preference for standardizing?

bcreane:
hey @caseydavenport  - thanks for your persistence. broke apart the error and the info level log messages ...

calicoctl/commands/clientmgr/client.go
```diff
@@ -45,7 +47,11 @@ func NewClient(cf string) (*client.Client, error) {
 // otherwise will load from environment variables.
 func LoadClientConfig(cf string) (*api.CalicoAPIConfig, error) {
 	if _, err := os.Stat(cf); err != nil {
-		log.Infof("Config file cannot be read - reading config from environment")
+		if (cf != constants.DefaultConfigPath) {
```
caseydavenport: Looks like in other error scenarios we do something like this: https://github.com/bcreane/calicoctl/blob/0b8193829d8a332b1c7e7d2711146d746f600fa1/calicoctl/commands/config.go#L110

bcreane: You're right, in general, though there is the following exception:
https://github.com/bcreane/calicoctl/blob/master/calicoctl/calicoctl.go#L94

But since 99% of the error messages are going to stdout, I'll respin the PR again and do the same.

bcreane:
yeah! thanks.

PR #1680, opened 2017-07-06
Added ability for calicoctl to read multiple yaml documents in the same file

mgleung:
## Description
Added the ability for calicoctl to iterate over multiple YAML documents from the same input 

Fixes https://github.com/projectcalico/libcalico-go/issues/47

## Testing
- Ran/wrote unit tests
- Ran system tests
- Tested by creating multiple policies from the same YAML file locally

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,21 +15,18 @@
 package resourcemgr
 
 import (
+	"bytes"
 	"errors"
-
 	"fmt"
+	"io"
+	"os"
 	"reflect"
 	"strings"
 
-	"io/ioutil"
-	"os"
-
-	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
-
-	"bytes"
-
 	log "github.com/Sirupsen/logrus"
+	yamlsep "github.com/projectcalico/calicoctl/calicoctl/util/yaml"
 	"github.com/projectcalico/go-yaml-wrapper"
+	"github.com/projectcalico/libcalico-go/lib/api/unversioned"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	"github.com/projectcalico/libcalico-go/lib/validator"
 )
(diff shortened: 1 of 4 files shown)


calicoctl/util/yaml/separator.go
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
bcreane: 2017?

calicoctl/util/yaml/separator_test.go
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) 2016 Tigera, Inc. All rights reserved.
```
bcreane: 2017

tmjd:
Please make sure you do a little testing with failures in the docs (invalid fields, bad values in fields) to make sure the error reporting makes sense.  From what I've seen before I expect that it will.

mgleung:
Tested a few misconfigured yaml docs (invalid types in fields, empty fields, wrong options in fields, empty documents) and found that everything makes sense.  We could modify the error messages in go-yaml-wrapper so that they are less specific to the underlying resource object and instead reference the field name in the yaml document but those changes would be a part of a separate PR and would also change the existing error messages.  

Also forgetting the yaml `---` separator in the files only applies the last resource specified in the document which mirrors the current behavior of calicoctl.  This could be improved to throw errors, but would also require modifications to the go-yaml library and is outside of the scope of this set of changes.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -228,19 +229,41 @@ func unmarshalSliceOfResources(tml []unversioned.TypeMetadata, b []byte) ([]unve
...
+			}
+			return nil, err
+		}
+
+		r, err := createResourcesFromBytes(b)
+		if err != nil {
+			return nil, err
+		}
+
+		if resources == nil {
```
robbrockbank: IIRC, you don't need this first branch - you can append to a nil slice which will create the slice for you.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -21,14 +21,15 @@ import (
 	"reflect"
 	"strings"
 
-	"io/ioutil"
+	"io"
```
robbrockbank: nit:  I realize this was already here, but we should probably just group together all of the native golang imports rather than having them split up across four groups.

robbrockbank: Copyright statements in all your updated files should be updated to include 2017.

robbrockbank:
To reiterate Eriks comment - deffo test live with some odd scenarios - e.g. no data between the --- separators, invalid data in one of the documents but not the others etc.

Also, good to test using both file input and stdin.

calicoctl/resourcemgr/resourcemgr.go
mgleung: good to know

PR #1686, opened 2017-07-14
Pin libcalico-go to version with multiple nets in rule support.

fasaxc:
## Description
Add support for multiple CIDRs in each rule by revving libcalico-go to pick up that change.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note - handled in other repo


robbrockbank:
@fasaxc STs need fixing.  We have similar STs in the calicoctl repo that test calicoctl round-tripping data.  Should we be removing these STs and just updating the calicoctl ones?

fasaxc:
I've updated the existing STs to use nets but didn't check the "auto-convert on read" behaviour since that's tested in the calico repo.

Interestingly, these STs did catch a bug in libcalico-go (incorrect JSON name for a field) so hard to argue for their removal.

robbrockbank:
@fasaxc:  Ha!  I think I must have been looking at a couple of different PRs at the same time -I could have sworn this was a calico PR (hence my comment about duplicate tests in calicoctl).

TLDR:  Ignore my comment about removing them!

LGTM.

PR #1687, opened 2017-07-20
Report version and cluster type from datastore in version command

tmjd:
## Description
As title says.
Also fixed build date in the Makefile so it is set properly.

## Todos
- [ ] Documentation
- [x] Release note

## Release Note
```release-note
The calicoctl version command now includes the CalicoVersion and ClusterType as retrieved from the datastore.
```


caseydavenport:
We should update the reference doc: http://docs.projectcalico.org/v2.3/reference/calicoctl/commands/version

PR #1696, opened 2017-07-28
Update the import for logrus to use lowercasing

mgleung:
## Description
Updating the logrus import to use the new updated user name (which is lowercased).  This will prevent import errors with vendoring going forward.

Requires https://github.com/projectcalico/libcalico-go/pull/484

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note
```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -5,14 +5,14 @@ owners:
 - name: Rob Brockbank
   email: rob@tigera.io
 import:
-- package: github.com/Sirupsen/logrus
+- package: github.com/sirupsen/logrus
   version: ^0.10.0
 - package: github.com/docopt/docopt-go
 - package: github.com/mcuadros/go-version
 - package: github.com/mitchellh/go-ps
 - package: github.com/olekukonko/tablewriter
 - package: github.com/osrg/gobgp
-  version: 38c47609a2eeae67c2c9cc042adea60484c2f481
+  version: v1.22
   subpackages:
   - client
   - packet/bgp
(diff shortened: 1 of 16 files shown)


glide.yaml
```diff
@@ -5,14 +5,14 @@ owners:
...
-- package: github.com/Sirupsen/logrus
+- package: github.com/sirupsen/logrus
   version: ^0.10.0
 - package: github.com/docopt/docopt-go
 - package: github.com/mcuadros/go-version
 - package: github.com/mitchellh/go-ps
 - package: github.com/olekukonko/tablewriter
 - package: github.com/osrg/gobgp
-  version: 38c47609a2eeae67c2c9cc042adea60484c2f481
+  version: v1.21.0
```
gunjan5: isn't the tag just `v1.21` (not `v1.21.0`)?

gunjan5: one more thing we need to keep in mind is we're updating the GoBGP client version, but server that's included in calico/node is still the older version

mgleung:
Tested the update to go-bgp with https://github.com/projectcalico/calico-bgp-daemon/pull/26 and got the following results.  Seems everything seems to work as expected.
```
core@m ~ $ sudo calicoctl node status
Calico process is running.

IPv4 BGP status
+---------------+-------------------+-------+----------+-------------+
| PEER ADDRESS  |     PEER TYPE     | STATE |  SINCE   |    INFO     |
+---------------+-------------------+-------+----------+-------------+
| 172.18.18.102 | node-to-node mesh | up    | 00:06:08 | Established |
| 172.18.18.103 | node-to-node mesh | up    | 00:06:41 | Established |
+---------------+-------------------+-------+----------+-------------+

IPv6 BGP status
No IPv6 peers found.
```

PR #1702, opened 2017-08-16
calicoctl node diags grab logs from hosted installs

### LINKED ISSUE #1690: calicoctl node diags should attempt to collect calico/node startup logs when started by Kubernetes
`calico node diags` fails to capture the calico/node startup logs when calico/node is running under (self-hosted) Kubernetes.  These logs are really important.

If possible, we should try and grab them (although invoking kubectl might not always work).

If not, documentation should be updated to encourage people to get kubectl logs for calico/node as well.

heschlie:
## Description
Fixes #1690 
Fixes projectcalico/calico#810

With this change `calicoctl node diags` will pickup logs from hosted installs.  This will allow for easier and quicker debugging when a user sends us their debug logs with hosted installs.

There is also a small change to add some clarity when a command fails, as before it was simply saying `exit status 1` will now tell you what command failed to run. This will give the user a better idea of what failed to run in the diags.

I've tested this via the Calico vagrant demo and verified it works with and without Calico containers (obviously not writing any logs when there are no Calico containers present).

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
`calicoctl node diags` will now gather logs from hosted Calico containers.
```

### DIFF
--- a/calicoctl/commands/node/diags.go
+++ b/calicoctl/commands/node/diags.go
@@ -15,11 +15,13 @@
 package node
 
 import (
+	"bufio"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -134,6 +136,9 @@ func runDiags(logDir string) {
 		fmt.Printf("No logs found in %s; skipping log copying", logDir)
 	}
 
+	// Try to copy logs from containers for hosted installs.
+	getNodeContainerLogs(tmpLogDir)
+
 	// Get the current time and create a tar.gz file with the timestamp in the name
 	tarFile := fmt.Sprintf("diags-%s.tar.gz", time.Now().Format("20060102_150405"))
 
@@ -155,7 +160,51 @@ such as transfer.sh using curl or similar.  For example:
 
 }
 
-// writeDiags executes the dignostic commans and outputs the result in the file
+// getNodeContainerLogs will attempt to grab logs for any "calico" named containers for hosted installs.
+func getNodeContainerLogs(logDir string) {
+	os.Mkdir(logDir, os.ModeDir)
+
+	// Get a list of Calico containers running on this Node.
+	result, err := exec.Command("docker", "ps", "-a", "--filter", "\"name=calico\"", "--format", "{{.Names}}: {{.CreatedAt}}").CombinedOutput()
+	if err != nil {
+		fmt.Printf("Could not run docker command: %s\n", string(result))
+		return
+	}
+
+	// No Calico containers found.
+	if string(result) == "" {
+		log.Debug("Did not find any Calico containers")
+		return
+	}
+
+	// Remove any containers that have "k8s_POD" in them.
+	re := regexp.MustCompile("(?m)[\r\n]+^.*k8s_POD.*$")
+	containers := re.ReplaceAllString(string(result), "")
+
+	fmt.Println("Copying logs from Calico containers")
+	err = ioutil.WriteFile(logDir+"/"+"container_creation_time", []byte(containers), 0666)
+	if err != nil {
+		fmt.Printf("Could not save output of `docker ps` command to container_creation_time: %s\n", err)
+	}
+
+	// Grab the log for each container and write it as <containerName>.log.
+	scanner := bufio.NewScanner(strings.NewReader(containers))
+	for scanner.Scan() {
+		name := strings.Split(scanner.Text(), ":")[0]
+		log.Debugf("Getting logs for container %s", name)
+		cLog, err := exec.Command("docker", "logs", name).CombinedOutput()
+		if err != nil {
+			fmt.Printf("Could not pull log for container %s: %s\n", name, err)
+			continue
+		}
+		err = ioutil.WriteFile(logDir+"/"+name+".log", cLog, 0666)
+		if err != nil {
+			fmt.Printf("Failed to write log for container %s to file: %s\n", name, err)
+		}
+	}
+}
+
+// writeDiags executes the diagnostic commands and outputs the result in the file
 // with the filename and directory passed as arguments
 func writeDiags(cmds diagCmd, dir string) {
 
@@ -165,11 +214,9 @@ func writeDiags(cmds diagCmd, dir string) {
 
 	parts := strings.Fields(cmds.cmd)
 
-	command := exec.Command(parts[0], parts[1:]...)
-
-	content, err := command.Output()
+	content, err := exec.Command(parts[0], parts[1:]...).CombinedOutput()
 	if err != nil {
-		fmt.Println(err)
+		fmt.Printf("Failed to run command: %s\nError: %s\n", cmds.cmd, string(content))
 	}
 
 	// This is for the commands we want to run but don't want to save the output


calicoctl/commands/node/diags.go
```diff
@@ -155,7 +159,39 @@ such as transfer.sh using curl or similar.  For example:
...
-// writeDiags executes the dignostic commans and outputs the result in the file
+// getDockerLogs will attempt to grab logs for any "calico" named containers for hosted installs.
+func getDockerLogs(logDir string) {
+	os.Mkdir(logDir, os.ModeDir)
+
+	fmt.Println("Copying logs from Calico containers")
+	// Get a list of Calico containers running on this Node.
+	out, err := exec.Command("docker", "ps", "-a", "--filter", "\"name=calico\"", "--format", "{{.Names}}").Output()
+	if err != nil {
+		fmt.Printf("Could not run docker command: %s\n", err)
```
bcreane: does it make sense to capture these failures (170 and 175) in the diag output? it's possible the errors are too unlikely to bother trying to capture.

heschlie: It certainly isn't much effort to do, and could shed light on a possible permissions issue if the user isn't running as root.  As for the No Calico containers found it might not really be an error, as someone might not be running a hosted install.

Also, this should be short circuiting here, and it's not :(

bcreane:
lgtm!

calicoctl/commands/node/diags.go
```diff
@@ -134,6 +136,9 @@ func runDiags(logDir string) {
 		fmt.Printf("No logs found in %s; skipping log copying", logDir)
 	}
 
+	// Try to copy logs from containers for hosted installs.
+	getDockerLogs(tmpLogDir)
```
gunjan5: I think we should call this `getContainerLogs`, because in future if we want to add rkt logs or any other container runtime logs it sounds more generic

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,43 @@ such as transfer.sh using curl or similar.  For example:
...
-// writeDiags executes the dignostic commans and outputs the result in the file
+// getDockerLogs will attempt to grab logs for any "calico" named containers for hosted installs.
+func getDockerLogs(logDir string) {
+	os.Mkdir(logDir, os.ModeDir)
+
+	fmt.Println("Copying logs from Calico containers")
+	// Get a list of Calico containers running on this Node.
+	var stderr bytes.Buffer
+	cmd := exec.Command("docker", "ps", "-a", "--filter", "\"name=calico\"", "--format", "{{.Names}}")
+	cmd.Stderr = &stderr
```
gunjan5: I'm not sure if this is necessary. I think you can just do the following:
```go
out,err := cmd.Output()
if err != nil {
     fmt.Println("blah: %s\n", err.Stderr)
     return
}
```
From Golang spec:
> Any returned error will usually be of type *ExitError. If c.Stderr was nil, Output populates ExitError.Stderr.

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,43 @@ such as transfer.sh using curl or similar.  For example:
 
 }
 
-// writeDiags executes the dignostic commans and outputs the result in the file
+// getDockerLogs will attempt to grab logs for any "calico" named containers for hosted installs.
+func getDockerLogs(logDir string) {
+	os.Mkdir(logDir, os.ModeDir)
+
+	fmt.Println("Copying logs from Calico containers")
```
gunjan5: we should move this after line 182, since if we don't find any Calico containers then we print this message suggesting we're collecting logs but in fact we are not (and the debug log goes silently unnoticed by the user)

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,43 @@ such as transfer.sh using curl or similar.  For example:
...
+	cmd := exec.Command("docker", "ps", "-a", "--filter", "\"name=calico\"", "--format", "{{.Names}}")
+	cmd.Stderr = &stderr
+	out, err := cmd.Output()
+	if err != nil {
+		fmt.Printf("Could not run docker command: %s\n", stderr.String())
+		return
+	}
+
+	// No Calico containers found.
+	if string(out) == "" {
```
gunjan5: not a fan of this variable name `out`, I think it should be `containerNames`, `containers` or `cNames` or something similar. WDYT?

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,43 @@ such as transfer.sh using curl or similar.  For example:
...
+		return
+	}
+
+	// Grab the log for each container and write it as <containerName>.log.
+	scanner := bufio.NewScanner(strings.NewReader(string(out)))
+	for scanner.Scan() {
+		name := scanner.Text()
+		cLog, err := exec.Command("docker", "logs", name).Output()
+		if err != nil {
+			fmt.Printf("Could not pull log for container %s: %s\n", name, err)
```
gunjan5: shouldn't we `continue` here if we get an error from line 188? else it will try to access `cLog` which might be a `nil` in line 192

calicoctl/commands/node/diags.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/docopt/docopt-go"
 	log "github.com/sirupsen/logrus"
 	shutil "github.com/termie/go-shutil"
+	"bufio"
```
gunjan5: nit: not sure if you ran `goimports` on this, but the recommendation is to have all the Go native imports together at the top

calicoctl/commands/node/diags.go
heschlie: I think `getHostedLogs` might be more appropriate?  As it is what we are trying to get, WDYT?

gunjan5: I like that!

calicoctl/commands/node/diags.go
heschlie: This gives me a runtime error if `docker` doesn't exist, but works if docker throws an error.  Whereas the other method works in both situations.

calicoctl/commands/node/diags.go
```diff
@@ -22,10 +22,12 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+	"bufio"
+	"bytes"
 
 	"github.com/docopt/docopt-go"
 	log "github.com/sirupsen/logrus"
-	shutil "github.com/termie/go-shutil"
```
gunjan5: nit: I much rather explicitly name the import when import name is different than the package name (not a good practice) like in this case package is `shutil` (and import is `go-shutil`) and we use `shutil` in the code, so I like to make it explicit.

gunjan5:
squash it when you address the last bit of nit (if you agree with it)

calicoctl/commands/node/diags.go
heschlie: Hm, I missed that, I agree with what you're saying, my IDE must have thought different.

heschlie:
@caseydavenport could I get you to do a once over of this PR to see if we can get it merged?

calicoctl/commands/node/diags.go
```diff
@@ -169,7 +212,7 @@ func writeDiags(cmds diagCmd, dir string) {
 
 	content, err := command.Output()
 	if err != nil {
-		fmt.Println(err)
```
caseydavenport: We're losing the error in this log the way it's written now.  I think we need to keep that so we can tell what failed.

calicoctl/commands/node/diags.go
```diff
@@ -134,6 +136,9 @@ func runDiags(logDir string) {
 		fmt.Printf("No logs found in %s; skipping log copying", logDir)
 	}
 
+	// Try to copy logs from containers for hosted installs.
+	getHostedLogs(tmpLogDir)
```
caseydavenport: a nit: this isn't specific to "self-hosted" calico installs - this is just "get stdout logs from any calico containers"

So, basically a change to the comment and function name (e.g. `getNodeContainerLogs`)

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,45 @@ such as transfer.sh using curl or similar.  For example:
...
 }
 
-// writeDiags executes the dignostic commans and outputs the result in the file
+// getHostedLogs will attempt to grab logs for any "calico" named containers for hosted installs.
+func getHostedLogs(logDir string) {
+	os.Mkdir(logDir, os.ModeDir)
+
+	// Get a list of Calico containers running on this Node.
+	var stderr bytes.Buffer
+	cmd := exec.Command("docker", "ps", "-a", "--filter", "\"name=calico\"", "--format", "{{.Names}}")
```
caseydavenport: Does this `--filter name=calico` actually work when Calico is deployed as a DaemonSet? I would think the name would be something like `k8s_calico-node-asdfasfdksdflahsdfhsdf`

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,45 @@ such as transfer.sh using curl or similar.  For example:
...
+
+	// No Calico containers found.
+	if string(containers) == "" {
+		log.Debug("Did not find any Calico containers")
+		return
+	}
+
+	fmt.Println("Copying logs from Calico containers")
+
+	// Grab the log for each container and write it as <containerName>.log.
```
caseydavenport: One thing we had talked about was also saving off the output of `docker ps` so that users could correlate containerName.log with a specific container / timestamp / etc.

i.e. if I'm looking at the directory this code generates, how do I know which of the 100 containers I'm interested in?

calicoctl/commands/node/diags.go
heschlie: That can be a partial match and will return any containers with "calico" in them, so it will match, I tested this on a Vagrant install and got the expected output.

calicoctl/commands/node/diags.go
heschlie: I'll get this setup so we see the returned error from the command,. Before we were only getting an unexplained `exit status 1` here.

calicoctl/commands/node/diags.go
heschlie: I recall talking about it, but at the moment I'm having trouble seeing what this would buy the user, at the moment the only output of the docker command is a list of container names that have "calico" in them, which we use for grabbing the logs and using that name for the logfile.  I could run a similar query with more flags that could include some extra info, like start/creation time and save that out?

caseydavenport: The main thing we're looking for is the ability to correlate a log file with a specific iteration of a container.

So, being able to see the start/creation time is important.  Alternatively we can just encode that into the file name somehow.

heschlie: The names are already pretty long, I'll dump a file with the format:

```
{{.Names}}: {{.CreatedAt}}
```

Which will give us a mapping.

heschlie:
@caseydavenport I think we might be all set?  Let me know if the changes to grabbing the container->created time mapping looks ok.

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,46 @@ such as transfer.sh using curl or similar.  For example:
...
+	}
+
+	// No Calico containers found.
+	if string(containers) == "" {
+		log.Debug("Did not find any Calico containers")
+		return
+	}
+
+	fmt.Println("Copying logs from Calico containers")
+	err = ioutil.WriteFile(logDir+"/"+"container_creation_time", containers,0666)
```
caseydavenport: You're never checking this error

calicoctl/commands/node/diags.go
```diff
@@ -155,7 +160,49 @@ such as transfer.sh using curl or similar.  For example:
...
+		return
+	}
+
+	// No Calico containers found.
+	if string(containers) == "" {
+		log.Debug("Did not find any Calico containers")
+		return
+	}
+
+	fmt.Println("Copying logs from Calico containers")
```
caseydavenport: Some output I got from running on a k8s system:

```
$ docker ps -a --filter name=calico --format "{{.Names}}: {{.CreatedAt}}"
k8s_install-cni_calico-node-6p0xp_kube-system_603c4dd8-98c8-11e7-acf3-42010a800008_0: 2017-09-13 21:13:44 +0000 UTC
k8s_calico-node_calico-node-6p0xp_kube-system_603c4dd8-98c8-11e7-acf3-42010a800008_0: 2017-09-13 21:13:36 +0000 UTC
k8s_calico-etcd_calico-etcd-t6dwl_kube-system_5fe749f4-98c8-11e7-acf3-42010a800008_0: 2017-09-13 21:13:26 +0000 UTC
k8s_POD_calico-node-6p0xp_kube-system_603c4dd8-98c8-11e7-acf3-42010a800008_0: 2017-09-13 21:13:23 +0000 UTC
k8s_POD_calico-etcd-t6dwl_kube-system_5fe749f4-98c8-11e7-acf3-42010a800008_0: 2017-09-13 21:13:23 +0000 UTC
```

We should filter out anything with `k8s_POD_` in it since those are just the infra containers and won't have useful logs.

heschlie: Ah, did not know that, will filter them out.

heschlie:
When trying to trim out `k8s_POD` containers I found that the policy controller log was empty, turns out we are only logging to stderr in that container, so I migrated to `exec.Command.CombinedOutput()` which coincidentally helped me cleanup a few other parts of the code.

PR #1703, opened 2017-08-27
"Types" field support in calicoctl

bcreane:
## Description
This PR is part of the ingress/egress feature:

- Pin libcalico-go to version with 'types' field support.
- Fix STs now that policies with both ingress and egress rules (but no types field) have 'types': 'ingress', 'egress' appended.
- Lastly add two STs that validate libcalico-go appends the `types` field when appropriate.

All STs and UTs pass.

## Todos
- [X] Tests

## Release Note

```release-note
None required
```


tests/st/calicoctl/test_crud.py
```diff
@@ -1098,3 +1108,67 @@ def check_no_data_in_store(testdata):
...
+
+        # Now read it out (yaml format) with calicoctl:
+        self.check_data_in_datastore([policy1_types_dict], "policy")
+
+        # Remove policy1
+        calicoctl("delete -f /tmp/policy1.yaml")
+
+        """
+        Test that a simple policy with only an ingress
+        rule will not have thetypes' field appended.
```
caseydavenport: typo: `thetypes`

tests/st/calicoctl/test_crud.py
```diff
@@ -627,7 +635,8 @@ def test_create_from_file(self, res, data1, data2):
                                      'ports': [1, 2, 3, 4],
                                      'tag': 'web'}}],
                     'order': 6543215.321,
-                    'selector': ''}},
+                    'selector': '',
+                    'types': ['ingress', 'egress']}},
```
caseydavenport: All of these tests are both `ingress` and `egress` - can we add a few new tests to validate ingress-only, and egress-only policies?

Also, a policiy that has no ingress or egress rules defined?

bcreane: The second test `pop`s the egress rule from policy1 and confirms that no `types` field is appended. I could make this more clear by creating another policy_dictionary, if necessary.

I'll add another egress-only test as well as a no-ingress and no-egress policy.

Note that there are already dozens of these tests in other sections that are effectively testing this behavior now as well.

bcreane:
hey @caseydavenport  - thanks for the review. I broke out the tests into clear combinations of egress+ingress rules and improved coverage.

tests/st/calicoctl/test_crud.py
```diff
@@ -1098,3 +1108,213 @@ def check_no_data_in_store(testdata):
...
+
+        self.writeyaml('/tmp/policy1_noegress.yaml', policy1_noegress_dict)
+
+        # Create the no egress policy with calicoctl
+        calicoctl("create -f /tmp/policy1_noegress.yaml")
+
+        # Now read it out (yaml formwat) and ensure no types field is present
+        self.check_data_in_datastore([policy1_noegress_dict], "policy")
+
+class TestTypes(TestBase):
```
caseydavenport: I think you're re-defining the `TestTypes` class here.  These probably should be combined or one of them needs to be renamed.

bcreane: my mistake, I intended to delete the first one. fixing up now.

PR #1711, opened 2017-09-25
Use V2 API for calicoctl

gunjan5:
## Description
- calicoctl using V2 APIs
- still cleaning up some code, but it largely works 
- manually tested for namespaced and non-namespaced resources, get, list, update, -o wide, -o yaml, -o json, get by name, etc.
- tested handling `--namespace`, `--all-namespaces` for different operations and resource kinds

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calicoctl is updated to use Calico V2 data model
```


calicoctl/calicoctl.go
```diff
@@ -89,7 +89,7 @@ Description:
 		case "ipam":
 			commands.IPAM(args)
 		case "config":
-			commands.Config(args)
+			//commands.Config(args)
```
gunjan5: per node config is not yet supported in libcalico V2 so code for that is commented out

calicoctl/commands/apply.go
```diff
@@ -43,6 +43,9 @@ Options:
   -c --config=<CONFIG>      Path to the file containing connection
                             configuration in YAML or JSON format.
                             [default: ` + constants.DefaultConfigPath + `]
+  -n --namespace=<NS>       Namespace of the resource.
+                            Only applicable to NetworkPolicy and WorkloadEndpoint.
+                            [default: default]
```
robbrockbank: Once you take the fixed libcalico-go you should remove the `[default: default]` tag and just state the it will use the default namespace.  As discussed, use the CLI-configured namespace as the one that you use to create the resource-specific client.

calicoctl/commands/printer.go
```diff
@@ -207,17 +207,18 @@ func join(items interface{}, separator string) string {
...
 	return func(name string) string {
 		switch strings.ToLower(name) {
 		case "asnumber":
 			if asValue == "" {
-				if asn, err := client.Config().GetGlobalASNumber(); err != nil {
-					asValue = "unknown"
-				} else {
-					asValue = asn.String()
-				}
+				//if asn, err := client.Config().GetGlobalASNumber(); err != nil {
```
robbrockbank: Could you add a TODO here (I know we don't do that -- but this seems like an easy bit of code to overlook when we come to handle the config options)

calicoctl/commands/resources.go
```diff
@@ -221,15 +208,18 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
...
+		resources = []runtime.Object{r}
 	}
 
 	if len(resources) == 0 {
 		return commandResults{err: errors.New("no resources specified")}
 	}
 
 	if log.GetLevel() >= log.DebugLevel {
-		log.Debugf("Resources: %v", resources)
+		for _, v := range resources {
```
robbrockbank: Resources -> Resource

calicoctl/resourcemgr/bgppeer.go
```diff
@@ -15,42 +15,51 @@
...
 		api.NewBGPPeerList(),
+		false,
+		[]string{"bgppeer", "bgppeers", "bgpp", "bgpps", "bp", "bps"},
 		[]string{"SCOPE", "PEERIP", "NODE", "ASN"},
 		[]string{"SCOPE", "PEERIP", "NODE", "ASN"},
 		map[string]string{
-			"SCOPE":  "{{.Metadata.Scope}}",
-			"PEERIP": "{{.Metadata.PeerIP}}",
-			"NODE":   "{{.Metadata.Node}}",
+			"SCOPE":  "{{ if eq .Spec.Node `` }}global{{ else }}node{{ end }}",
```
robbrockbank: Maybe just get rid of the SCOPE field.  For a global-scoped peer you could use the name `(global)` or similar since that isn't a valid name.

calicoctl/resourcemgr/hostendpoint.go
```diff
@@ -15,15 +15,20 @@
...
 )
 
 func init() {
+	ctx := context.Background()
 	registerResource(
 		api.NewHostEndpoint(),
 		api.NewHostEndpointList(),
+		false,
+		[]string{"hostendpoint", "hostendpoints", "hep", "heps"},
 		[]string{"NODE", "NAME"},
```
robbrockbank: I think the NAME field should be the leading field for each resource now.

I kind feel we should remove name from these lists and add it in automagically (basically just include the Spec fields).  WDYT?

calicoctl/commands/resources.go
```diff
@@ -58,104 +57,92 @@ func convertToSliceOfResources(loaded interface{}) []unversioned.Resource {
...
-	orchestrator := argutils.ArgStringOrBlank(args, "--orchestrator")
-	resScope := argutils.ArgStringOrBlank(args, "--scope")
+	namespace := argutils.ArgStringOrBlank(args, "<NS>")
+
 	switch strings.ToLower(kind) {
 	case "node", "nodes", "no", "nos":
-		p := api.NewNode()
-		p.Metadata.Name = name
-		return *p, nil
+		n := api.NewNode()
```
robbrockbank: shouldn't need this switch statement now - we should be able to do something better using the resource manager.  Maybe something like:

-  For each resource in the resource manager your include a slice of CLI names (and short names).
-  Store these in a map so that you can do a fast locate of the resource-specific manager for the type specified on the CLI
-  Then all this code disappears.

robbrockbank: It looks like you got as far as adding them to the Res, but just aren't using them.

calicoctl/commands/get.go
```diff
@@ -169,7 +165,15 @@ Description:
 		os.Exit(1)
 	}
 
-	results := executeConfigCommand(parsedArgs, actionList)
+	results := commandResults{}
+
+	name := argutils.ArgStringOrBlank(parsedArgs, "<NAME>")
+	if name == "" {
+		results = executeConfigCommand(parsedArgs, actionList)
```
robbrockbank: I don't think this is the correct fix since it won't correctly handle file-based input.

And I assume with file-based input you'll also need to worry about whether each entry specifies a name/namespace and use get/list accordingly.

Probably what I would do is have the ResourceManager expose a Get() interface which actually maps to either the List or Get call depending on whether the namespace or name is wild carded.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -313,42 +344,53 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
-func (rh resourceHelper) Delete(client *client.Client, resource unversioned.Resource) (unversioned.Resource, error) {
+func (rh resourceHelper) Delete(client client.Interface, resource ResourceObject) (ResourceObject, error) {
 	return rh.delete(client, resource)
 }
 
+// Get is an un-typed method to get an existing resource.  This calls directly
+// through to the resource helper specific Get method which will map the untyped call to
+// the typed interface on the client.
+func (rh resourceHelper) Get(client client.Interface, resource ResourceObject) (ResourceObject, error) {
+	return rh.get(client, resource)
```
robbrockbank: This is where I think you should do the logic to do a Get or List depending on whether the resource is qualified with a name/namespace or not.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -313,42 +344,53 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+}
+
 // List is an un-typed method to list existing resources.  This calls directly
 // through to the resource helper specific List method which will map the untyped call to
 // the typed interface on the client.
-func (rh resourceHelper) List(client *client.Client, resource unversioned.Resource) (unversioned.Resource, error) {
+func (rh resourceHelper) List(client client.Interface, resource ResourceObject) (ResourceListObject, error) {
 	return rh.list(client, resource)
```
robbrockbank: And probably just remove this.

robbrockbank: Or add a GetOrList method() that does what I suggested for Get.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -313,42 +344,53 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+func (rh resourceHelper) GetObjectType() reflect.Type {
+	return rh.resourceType
+}
+
 // Apply is an un-typed method to apply (create or update) a resource.  This calls directly
 // through to the resource helper specific Apply method which will map the untyped call to
 // the typed interface on the client.
-func (rh resourceHelper) Apply(client *client.Client, resource unversioned.Resource) (unversioned.Resource, error) {
+func (rh resourceHelper) Apply(client client.Interface, resource ResourceObject) (ResourceObject, error) {
 	return rh.apply(client, resource)
```
robbrockbank: This is where you should do the Create or Update logic.

robbrockbank: and remove the apply parameter from the ResourceManager.

calicoctl/commands/resources.go
gunjan5: I'm guessing you're talking about the log line below this, and not this one itself?

calicoctl/resourcemgr/bgppeer.go
gunjan5: how about something like this?
```
NAME    PEERIP        NODE          ASN
name1     192.168.1.1   rack1-host1   63400
namester   1.1.1.1       (global)              63444
```

calicoctl/resourcemgr/hostendpoint.go
gunjan5: > I think the NAME field should be the leading field for each resource now.

I agree

> I kind feel we should remove name from these lists and add it in automagically (basically just include the Spec fields). WDYT?

Not sure I understand what you mean by adding the name field automagically. Can you give me an example?

calicoctl/resourcemgr/ippool.go
```diff
@@ -29,35 +29,32 @@ func init() {
...
+		[]string{"NAME", "CIDR"},
+		[]string{"NAME", "CIDR", "NAT", "IPIP", "DISABLED"},
 		map[string]string{
-			"CIDR": "{{.Metadata.CIDR}}",
-			"NAT":  "{{.Spec.NATOutgoing}}",
-			"IPIP": "{{if .Spec.IPIP}}{{.Spec.IPIP.Enabled}}{{else}}false{{end}}",
+			"NAME":     "{{.ObjectMeta.Name}}",
+			"CIDR":     "{{.Spec.CIDR}}",
+			"NAT":      "{{.Spec.NATOutgoing}}",
+			"IPIP":     "{{.Spec.IPIP.Mode}}",
```
robbrockbank: IPIP is still an optional struct so probably need to carry across the check that the struct exists, defaulting this value to "Always" if it doesn't exist.

I wonder if we should always set the Mode to Always in libcalico-go, so that querying the resource always returns this with a value?  WDYT?

calicoctl/commands/resources.go
```diff
@@ -304,3 +249,51 @@ func executeResourceAction(args map[string]interface{}, client client.Interface,
 
 	return []runtime.Object{resOut}, err
 }
+
+// handleNamespace fills in the namespace information in the resource (if required),
+// and validates the namespace depending on whether or not a namespace should be
+// provided based on the resource kind.
+func handleNamespace(resource runtime.Object, args map[string]interface{}) (resourcemgr.ResourceObject, string, bool, error) {
```
robbrockbank: Not sure what the string and bool return values are.

The resource input parameter could also be the output parameter, so you don't actually need to include the resourceObject in the results.  In any case, why is the input resource a runtime.Object, but the return value a resurcemgr.ResourceObject?

robbrockbank: Do you need to return the namespace?  Isn't it included in the resource?

calicoctl/commands/resources.go
```diff
@@ -267,24 +208,28 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
...
+	if err != nil {
+		return nil, err
+	}
+
 	var resOut runtime.Object
 
 	switch action {
 	case actionApply:
-		resOut, err = rm.Apply(client, resource.(resourcemgr.ResourceObject))
+		resOut, err = rm.Apply(client, resource, namespace)
```
robbrockbank: If we insist that the namespace is now included in the resource when required, then the namespace parameter in these signatures seems redundant.

robbrockbank: even in the GetOrList - provided you leave either the name or the namespace blank then the GetOrList function can do it's thing and workout what to include in the filter options.

calicoctl/commands/resources.go
robbrockbank: It feels like all of the parameters except the error are redundant:
-  The resource on the input parms can also be used as the output, e.g.  `handleNamespace(inout runtime.Object, arg ...)` ... a useful thing I picked up reading the k8s code
-  The namespace can be extracted from the resource
-  The all namespaces can be inferred from the `resource.Namespace == ""`.  From a Get/LIst perspective, just fill in the Name and Namespace from the values in the resource and the blank values will automatically be treated as wildcards.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
+		} else {
+			resource.GetObjectMeta().SetResourceVersion(ro.(ResourceListObject).GetListMeta().GetResourceVersion())
+		}
+
+		ru, err := rh.update(client, resource, ns)
+		if err != nil {
+			switch err.(type) {
+			// Wait for a second and try again if there was a conflict during the resource update.
+			case cerrors.ErrorResourceUpdateConflict:
+				log.Warnf("Error updating the resource %s: %s. Retrying.", resource.GetObjectMeta().GetName(), err)
```
robbrockbank: I'm not sure this is a warning or just info - it's a perfectly valid scenario.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
+			return ro, err
+		}
+		if resource.GetObjectMeta().GetName() != "" {
+			resource.GetObjectMeta().SetResourceVersion(ro.(ResourceObject).GetObjectMeta().GetResourceVersion())
+		} else {
+			resource.GetObjectMeta().SetResourceVersion(ro.(ResourceListObject).GetListMeta().GetResourceVersion())
+		}
+
+		ru, err := rh.update(client, resource, ns)
+		if err != nil {
```
robbrockbank: Can't you simplify 399-410 just using:

```
if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
    log...
    sleep
    continue
}
```

robbrockbank: you don't need the default branch since it'll drop through to the return anyways.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
+
+			// For any other errors, return the error
+			default:
+				return ru, err
+			}
+		}
+
+		return ru, err
+	}
+	return nil, errors.New("failed to update the resource after 5 retries")
```
robbrockbank: maybe just return the last error, or just return that there was an update conflict.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
 // Get is an un-typed method to get an existing resource.  This calls directly
 // through to the resource helper specific Get method which will map the untyped call to
 // the typed interface on the client.
-func (rh resourceHelper) Get(client client.Interface, resource ResourceObject) (ResourceObject, error) {
-	return rh.get(client, resource)
-}
+func (rh resourceHelper) GetOrList(client client.Interface, resource ResourceObject, ns string, allNs bool) (runtime.Object, error) {
+	if resource.GetObjectMeta().GetName() != "" {
+		if allNs {
+			return nil, fmt.Errorf("cannot use --all-namespace flag for getting a single resource")
```
robbrockbank: what does kubectl do if you use --all-namespaces to get a non-namespaces resource?  I assume it just ignores it?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
 func GetResourceManager(resource runtime.Object) ResourceManager {
 	return helpers[resource.GetObjectKind().GroupVersionKind()]
 }
+
+func GetResourceFromArgs(args map[string]interface{}) (runtime.Object, error) {
+	kind := args["<KIND>"].(string)
+	name := argutils.ArgStringOrBlank(args, "<NAME>")
+	namespace := argutils.ArgStringOrBlank(args, "--namespace")
+
+	res, ok := nameToRes[strings.ToLower(kind)]
```
robbrockbank: should nameToRes actually be kindToRes?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -351,46 +358,108 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
+	if !ok {
+		return nil, fmt.Errorf("Resource type '%s' is not supported", kind)
```
robbrockbank: lower case r?

calicoctl/commands/resources.go
```diff
@@ -304,3 +249,51 @@ func executeResourceAction(args map[string]interface{}, client client.Interface,
...
+	// If --all-namespaces is used then we must set namespace to "" so
+	// list operation can list resources from all the namespaces. Otherwise for
+	// the rest of the operations, we set the namespace to "default" if not specified.
+	if allNs {
+		namespace = ""
+	} else if namespace == "" {
+		namespace = "default"
+	}
+
+	if resourcemgr.IsNamespaced(resource) {
```
robbrockbank: I'm not we should have static methods for things like IsNamespaced.  Really we should be looking up the resource helper and using an accessor method on that to determine IsNamespaced().  I realize that means you have a little extra code, but it keeps things nicely encapsulated.  Besides, you can always pass the resource manager into this function since you have it to have when you call this function.

robbrockbank: Or alternatively we only have static methods and hide the actual resource helper lookup.

robbrockbank: WDYT?

calicoctl/resourcemgr/ippool.go
gunjan5: I think I like setting it in libcalico better because it's explicit. It's not always obvious what the default value is, so libcalico filling that in would be nice. I think then same _might_ go with BGPPeer, node field, we can set it to `(global)` if it;s not specified.

calicoctl/commands/resources.go
robbrockbank: yup

calicoctl/resourcemgr/ippool.go
robbrockbank: I'm up for the IPIP one.  The Node one - displaying a value of (global) is really just a calicoctl nicety., so I don't think that one should.

calicoctl/resourcemgr/resourcemgr.go
gunjan5: seems like they return an error like this:
```
$ kubectl get clusterrolebinding --all-namespaces
NAMESPACE   NAME      AGE
error: clusterRoleBinding is not namespaced
```

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +85,201 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+// through to the resource helper specific Update method which will map the untyped call to
+// the typed interface on the client.
+func (rh resourceHelper) Update(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+	var err error
+	for i := 0; i < 5; i++ {
+		ro, err := rh.GetOrList(client, resource.(ResourceObject))
+		if err != nil {
+			return ro, err
+		}
+		if resource.(ResourceObject).GetObjectMeta().GetName() != "" {
```
robbrockbank: Does this work?  Shouldn't you be checking if this is a ResourceObject or a ResourceList *before* casting it?  Won't this panic when it does a resource list?

robbrockbank: Actually, shouldn't the above call directly into the private Get() function rather than using the GetOrList?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +85,201 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+
+// Create is an un-typed method to create a new resource.  This calls directly
+// through to the resource helper specific Create method which will map the untyped call to
+// the typed interface on the client.
+func (rh resourceHelper) Create(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+	return rh.create(client, resource.(ResourceObject))
+}
+
+// Update is an un-typed method to update an existing resource.  This calls directly
+// through to the resource helper specific Update method which will map the untyped call to
```
robbrockbank: comment is not correct.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,213 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+// Update is an un-typed method to update an existing resource. This calls the resource
+// specific Get method to get the resourceVersion, and then calls resource specific
+// Update method with the resource with the updated resourceVersion.
+// We retry 5 times if there is an update conflict during the Update operation.
+func (rh resourceHelper) Update(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+	var err error
+	for i := 0; i < 5; i++ {
+
+		// Get the resource to get the resourceVersion.
+		ro, err := rh.get(client, resource.(ResourceObject))
```
robbrockbank: Shouldn't we only get the resource version if it wasn't explicitly supplied in the resource?

I assume if the YAML contains the ResourceVersion and we do an update, we expect calicoctl to honor that?

robbrockbank: Which means you probably need to keep track of whether the resource version was initially supplied so that you don't keep retrying in that particular case.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,213 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+
+func (rh resourceHelper) GetObjectType() reflect.Type {
+	return rh.resourceType
+}
+
+// Apply is an un-typed method to apply (create or update) a resource. This calls Create
+// and if the resource already exists then we call the Update method.
+func (rh resourceHelper) Apply(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+
+	// Try to assert the resource type and remove the resourceVersion, because
```
robbrockbank: Did you ascertain what the correct behavior should be for apply?  should it care about the ResourceVersion or not?

calicoctl/commands/node/run.go
```diff
@@ -239,7 +239,7 @@ Description:
 		fmt.Println("Error executing command: invalid config file")
 		os.Exit(1)
 	}
-	if cfg.Spec.DatastoreType != api.EtcdV2 {
+	if cfg.Spec.DatastoreType != apiconfig.EtcdV2 || cfg.Spec.DatastoreType != apiconfig.EtcdV3 {
```
robbrockbank: We will retire etcdv2, so safe to remove that I think.

calicoctl/resourcemgr/resourcemgr.go
gunjan5: not sure I agree. K8s, for example, just ignores ResourceVersion if it's provided for Apply/Update. What is the use case for honoring it for update? I feel like it'll just cause a lot of user issues when users try to save the yaml output and then use apply to update their resources and we reject it because the ResourceVersion is not the latest.

gunjan5: Actually, I take that back. I just tested it, they ignore it for Create it seems like, but not for Update/Apply

calicoctl/resourcemgr/resourcemgr.go
gunjan5: yes they care about it, so what I'm thinking is we should do an update first (which cares about the resourceVersion) and if that fails then we do a Create (which ignores it)

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -157,13 +157,18 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
...
 func (rh resourceHelper) Apply(client client.Interface, resource runtime.Object) (runtime.Object, error) {
 
+	// Store the original ResourceVersion for the Update operation later.
+	originalRV := ""
+
 	// Try to assert the resource type and remove the resourceVersion, because
 	// Create call can't have resourceVersion set and Update automatically
 	// gets and sets the resourceVersion to the latest one from the datastore.
 	_, ok := resource.(ResourceObject)
 	if ok {
```
robbrockbank: We don't apply ResourceLists - so I don't think this logic is necessary.

robbrockbank: and if it was a better syntax would be:

```
if res, ok := resource.(ResourceObject); ok {
  res.GetObjectMeta().SetResourceVersion("")
}
```

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -193,28 +201,45 @@ func (rh resourceHelper) Create(client client.Interface, resource runtime.Object
...
 	var err error
-	for i := 0; i < 5; i++ {
+	rv := ""
 
-		// Get the resource to get the resourceVersion.
-		ro, err := rh.get(client, resource.(ResourceObject))
-		if err != nil {
-			return ro, err
-		}
+	// Check to see if the resourceVersion is specified in the resource object.
```
robbrockbank: ditto.

robbrockbank: and both of these branches are the same!

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -157,13 +157,18 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
 // and if the resource already exists then we call the Update method.
 func (rh resourceHelper) Apply(client client.Interface, resource runtime.Object) (runtime.Object, error) {
```
robbrockbank: Given we can only do these actions on Resources should the resource parm actually just be of type ResourceObject

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -193,28 +201,45 @@ func (rh resourceHelper) Create(client client.Interface, resource runtime.Object
...
-		// Try to update with the resource with the updated resourceVersion.
-		ru, err := rh.update(client, resource.(ResourceObject))
-		if _, ok := err.(cerrors.ErrorResourceUpdateConflict); ok {
-			// Wait for a second and try again if there was a conflict during the resource update.
-			log.Infof("Error updating the resource %s: %s. Retrying.", resource.(ResourceObject).GetObjectMeta().GetName(), err)
-			time.Sleep(1 * time.Second)
-			continue
-		}
+			// Set the resource-to-be-updated's resourceVersion to the one we got from the Get call.
+			resource.(ResourceObject).GetObjectMeta().SetResourceVersion(ro.(ResourceObject).GetObjectMeta().GetResourceVersion())
```
robbrockbank: Lots of casting to type ResourceObject which could be removed if the method signature just used this type rather than runtime.Object.

calicoctl/commands/resources.go
```diff
@@ -41,15 +38,15 @@ const (
...
+	actionGetOrList
 )
 
 // Convert loaded resources to a slice of resources for easier processing.
 // The loaded resources may be a slice containing resources and resource lists, or
 // may be a single resource or a single resource list.  This function handles the
 // different possible options to convert to a single slice of resources.
-func convertToSliceOfResources(loaded interface{}) []unversioned.Resource {
-	r := []unversioned.Resource{}
+func convertToSliceOfResources(loaded interface{}) []runtime.Object {
```
robbrockbank: This is a actually converting to a slice of ResourceObjects, so perhaps that's what this should return.  I think it'll make the rest of the plumbing much simpler

robbrockbank: Remember that a runtime.Object really is just something with a Kind/Version so that covers both Lists and Objects - but in this particular case we should only be returning Objects.  Probably need to beef up the function comment to make that clearer.

robbrockbank:
Semaphore is failing.  I think the STs need to be updated as part of this PR.

calicoctl/commands/get.go
```diff
@@ -47,18 +47,13 @@ Options:
   -o --output=<OUTPUT FORMAT>  Output format.  One of: yaml, json, ps, wide,
```
robbrockbank: Need to include `yml` in this list now.

calicoctl/commands/resources.go
```diff
@@ -41,15 +38,15 @@ const (
...
+	res := []runtime.Object{}
 	log.Infof("Converting resource to slice: %v", loaded)
 
 	switch reflect.TypeOf(loaded).Kind() {
```
robbrockbank: I actually think this processing can be simplified a lot now....

We know that the interface passed in will either be:
-  A runtime.Object
-  A slice of runtime.Objects

And then for each runtime.Object, we know that it will either be
-  A ResourceObject
-  A ResourceList

That's probably much "safer" processing than all the Ptr/Struct stuff which we originally needed because our v1 resources and lists didn't implement an interface.

Have you put this code in it's current guide through a thorough test, testing applying a List or a single resource, or a slice of Lists or slice of Lists and resources etc.  ?

calicoctl/calicoctl.go
robbrockbank: okay

calicoctl/commands/resources.go
robbrockbank: (maybe simplified is not the right word - but certainly made easier to understand).

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+
+	for _, v := range names {
+		kindToRes[v] = res
+	}
+}
+
+func (rh resourceHelper) GetObjectType() reflect.Type {
+	return rh.resourceType
+}
+
```
robbrockbank: Grrr.  Github lost some comments I'd made on this... Oh well, here we go again....

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+// and if the resource already exists then we call the Update method.
+func (rh resourceHelper) Apply(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+
+	// Store the original ResourceVersion for the Update operation later.
+	originalRV := ""
+
+	// Try to assert the resource type and remove the resourceVersion, because
+	// Create call can't have resourceVersion set and Update automatically
+	// gets and sets the resourceVersion to the latest one from the datastore.
+	_, ok := resource.(ResourceObject)
```
robbrockbank: This logic could be better expressed:

```
if res, ok := resource.(ResourceObject); ok {
  res.GetObjectMeta().SetResourceVersion("")
}
```

so that you don't have to cast multiple times.

robbrockbank: However...

The Apply/Update etc. methods deal with ResourceObjects and not ResourceListObjects therefore this check should be redundant.  Instead, it would make more sense to change the signature of this method to make that clearer - so use ResourceObject rather than runtime.Object.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+		resource.(ResourceObject).GetObjectMeta().SetResourceVersion("")
+	} else {
+		originalRV = resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
+		resource.(ResourceListObject).GetListMeta().SetResourceVersion("")
+	}
+
+	// Try to create the resource first.
+	ro, err := rh.Create(client, resource)
+
+	switch err.(type) {
```
robbrockbank: In general a switch statement seems quite a heavyweight way to do a single type check.  What's wrong with the 

```
if _, ok := err.(ErrorResourceAlreadyExists); ok {
}
```

I think that is far more readable when there is really only one branch.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+// Update method with the resource with the updated resourceVersion.
+// We retry 5 times if there is an update conflict during the Update operation.
+func (rh resourceHelper) Update(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+	var err error
+	rv := ""
+
+	// Check to see if the resourceVersion is specified in the resource object.
+	_, ok := resource.(ResourceObject)
+	if ok {
+		rv = resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
```
robbrockbank: Same as with Apply, *and* both of these branches are the same anyway.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+	} else {
+		rv = resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
+	}
+
+	// If the resourceVersion is specified then we don't try to Get it or retry.
+	if rv != "" {
+		return rh.update(client, resource.(ResourceObject))
+	} else {
```
robbrockbank: else is redundant.

calicoctl/resourcemgr/resourcemgr.go
robbrockbank: With a kubectl Apply - does it ignore the RV if the resource doesn't exists (i.e. it's doing a create) and use the RV is it does exists (and is therefore doing an update), or does it basically assume that if an RV is specified it will always be an Update?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -56,78 +83,238 @@ type ResourceActionCommand func(*client.Client, unversioned.Resource) (unversion
...
+// Create is an un-typed method to create a new resource.  This calls directly
+// through to the resource helper specific Create method which will map the untyped call to
+// the typed interface on the client.
+func (rh resourceHelper) Create(client client.Interface, resource runtime.Object) (runtime.Object, error) {
+	return rh.create(client, resource.(ResourceObject))
+}
+
+// Update is an un-typed method to update an existing resource. This calls the resource
+// specific Get method to get the resourceVersion, and then calls resource specific
+// Update method with the resource with the updated resourceVersion.
```
robbrockbank: we only update the RV if it isn't already specified.

calicoctl/commands/get.go
gunjan5: It's not a different option, it's just if someone types `yml` instead of `yaml` since people use them both interchangeably. This is like how we accept `ipp` for `ippool`, we don't list all the short versions we accept, just the right one

robbrockbank: yeah i'm not fussed about this one.

calicoctl/commands/resources.go
```diff
@@ -305,10 +230,52 @@ func executeResourceAction(args map[string]interface{}, client *client.Client, r
...
+		switch {
+		case allNs && cliNs != "":
+			// Check if --namespace and --all-namespaces flags are used together.
+			return fmt.Errorf("cannot use both --namespace and --all-namespaces flags at the same time")
+		case resNs == "" && cliNs != "":
+			// If resource doesn't have a namespace specified
+			// but it's passed in through the -n flag then use that one.
+			resource.GetObjectMeta().SetNamespace(cliNs)
+		case resNs != "" && allNs:
+			// If --all-namespaces is used then we must set cliNs to "" so
```
robbrockbank: nit:  you aren't setting cliNs as per the comment.

calicoctl/resourcemgr/bgppeer.go
```diff
@@ -15,42 +15,46 @@
...
-	"github.com/projectcalico/libcalico-go/lib/client"
+	"context"
+
+	api "github.com/projectcalico/libcalico-go/lib/apiv2"
+	client "github.com/projectcalico/libcalico-go/lib/clientv2"
+	"github.com/projectcalico/libcalico-go/lib/options"
 )
 
 func init() {
+	ctx := context.Background()
```
robbrockbank: ugh - sorry i didn't spot this before.

I don't think declaring the context here is very API-friendly.  Really we should be passing context in to each of the functions.

I realize we could just drop this, but I want to pull the resource manager out into libcalico-go and being able to pass in the context may prove important for the AAPIS.  Generally speaking it's important too if (for example) we later want to add timeouts to the CLI commands - which this approach does not easily allow for.

Could you pull out the context and pass it in the various commands down the stack.  It's lower priority than getting the tests running and can be done as a separate PR - but would be useful to have that done before next sprint.

robbrockbank:
Looks great!  So, just waiting on the FVs now.

I had one comment about contexts but we can pick that up after this PR is checked in.

tests/st/utils/utils.py
```diff
@@ -48,7 +48,7 @@
 IPV6_RE = re.compile(r'inet6 ([a-fA-F\d:]+)/\d{1,3}')
 
 
-def calicoctl(command):
+def calicoctl(input, command):
```
robbrockbank: I'm not sure I see the point of this input parameter.  It makes reading the code a lot less clear and it's really just part of the main command.

Perhaps better to just rename this to `executeCommand()`, or perhaps `commandWithDatastoreEnvs()` if you really wanna describe what it's doing.

robbrockbank: As discussed, I misread this code so, comment above is not sensible.  However, we discussed an improved method signature for this that'll make the command more useful and extensible.  

```
type CalicoctlOptions struct {
	Filename string
	InsertFileAsStdin bool
}

func calicoctl(command string, options... Options) {
	do stuff.  options will be length or 1
}

calicoctl("delete tea pgtips")
calicoctl("create", Options{Filename: "input.yaml"})

```

robbrockbank: Perhaps this should live in a calicoctl sub-module, so you don't need so much repetition.

e.g.

```
calicoctl.Command("apply", Options{Filename: "input.yaml"})
```

Not too fussed though.

robbrockbank: oh silly me.  It's python

Use named parameters.

robbrockbank:
LGTM - one tests pass.

Squash and merge.

PR #1714, opened 2017-10-01
STs and pass-through of ResourceVersion into List/Get/Delete

robbrockbank:
This PR includes the following:
-  When outputting data as YAML or JSON - don't display as a list if there is only one entry in the list.  This makes the output less structured in the mainline case.
-  Tweaks to the ST infrastructure to make it easier for future tests
-  Maintaining a set of resource data to use in the STs
-  Pass ResourceVersion through to Get/List/Delete commands
-  Retire the previous STs in favor of function-specific tests.
-  Tests cover *all* resource types

I'd recommend reviewing the test_crud.py file as it's own thing rather than looking at the diff.


calicoctl/resourcemgr/hostendpoint.go
```diff
@@ -47,14 +47,15 @@ func init() {
...
 		},
 		func(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceObject, error) {
 			r := resource.(*api.HostEndpoint)
-			return client.HostEndpoints().Get(ctx, r.Name, options.GetOptions{})
+			return client.HostEndpoints().Get(ctx, r.Name, options.GetOptions{ResourceVersion: r.ResourceVersion})
 		},
 		func(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceListObject, error) {
-			return client.HostEndpoints().List(ctx, options.ListOptions{})
+			r := resource.(*api.HostEndpoint)
+			return client.HostEndpoints().List(ctx, options.ListOptions{ResourceVersion: r.ResourceVersion, Name: r.Name})
```
gunjan5: why do we need a `r.Name` in the `ListOptions`? this function only gets called when the name is not specified.

tests/st/utils/data.py
```diff
@@ -0,0 +1,414 @@
...
+        'pod': 'abcd',
+        'endpoint': 'eth0',
+        'containerID': 'container1234',
+        'ipNetworks': ['1.2.3.4/32'],
+        'interfaceName': 'cali1234',
+        'profiles': ['prof1', 'prof2'],
+    }
+}
+
+workloadendpoint_name2_rev1 = {
```
gunjan5: where is the `node` `kind`?

tests/st/calicoctl/test_crud.py
```diff
@@ -35,745 +35,1089 @@ class TestPool(TestBase):
...
 
-        ipv4_pool_dict = {'apiVersion': 'projectcalico.org/v2',
-                          'kind': 'IPPool',
-                          'metadata': { 'name': 'ippool-123'},
-                          'spec': {'ipip': {'mode': 'Always'},
-                                   'cidr': str(ipv4_net.cidr)}
-                          }
+        # Add in the ipv4 network with calicoctl, and read out using an exact
+        # get, and a list query.
+        calicoctl("create", data=ippool_name1_rev1_v4)
```
gunjan5: missing `rc.assert_no_error()`?

robbrockbank: fixed

calicoctl/resourcemgr/hostendpoint.go
robbrockbank: it's for future work with AAPIS - they needed the ability to List/Watch specifying name.

tests/st/utils/data.py
robbrockbank: dammit - must have missed one.

robbrockbank: just added

calicoctl/resourcemgr/hostendpoint.go
robbrockbank: incidentally, I think it is still correct for GetOrList to fan out to Get when the resource is fully qualified - because in that case we don't want to display a ResourceList.

tests/st/calicoctl/test_crud.py
```diff
@@ -35,745 +35,1091 @@ class TestPool(TestBase):
...
+        rc.assert_no_error()
+        rev0 = rc.decoded
 
-        # Assert that deleting the pool again fails.
-        self.assertRaises(CommandExecError, calicoctl, "delete", "/tmp/ipv4.yaml")
+        # Update the BGP Peer and get it to assert the resource version is not
+        # the same.
+        rc = calicoctl("apply", data=bgppeer_name1_rev2_v4)
+        rc.assert_no_error()
+        rc = calicoctl("get bgppeer %s -o yaml" % name(bgppeer_name1_rev1_v4))
```
gunjan5: I think this seems a bit confusing that you're doing a get with `rev1` data. I know Get only cares about metadata which hasn't changed but getting with `rev2` would be clearer

tests/st/calicoctl/test_crud.py
```diff
@@ -35,745 +35,1091 @@ class TestPool(TestBase):
...
+        rc.assert_no_error()
+        rev1 = rc.decoded
+        self.assertNotEqual(rev0, rev1)
```
gunjan5: how is this asserting that **resourceVersion** is not the same? `rev0` and `rev1` seem to have the whole resource and they have different spec fields so they're going to be different. So this assertion isn't specifically checking for different `resourceVersion`, is that correct? Comment suggests this test checks if the RVs are different

tests/st/calicoctl/test_crud.py
```diff
@@ -35,745 +35,1091 @@ class TestPool(TestBase):
...
-        logger.debug("Testing %s" % res_type)
-        # Write out the files to load later
-        self.writeyaml('/tmp/%s-1.yaml' % res_type, data)
+        # Replace using the original data, but with the new resource version.
+        rev0['metadata']['resourceVersion'] = rev1['metadata']['resourceVersion']
+        rc = calicoctl("replace", data=rev0)
+        rc.assert_no_error()
 
-        calicoctl("create", "/tmp/%s-1.yaml" % res_type)
+        # Delete the resource without using a resource version.
```
gunjan5: how does this `rev0` not have a `resourceVersion`? Didn't you set it in line 135?

tests/st/calicoctl/test_crud.py
robbrockbank: Ha!  you're quite correct.  My bad.

tests/st/calicoctl/test_crud.py
robbrockbank: because we are only using the name to do the delete.

robbrockbank: i've tweaked the comment to spell this out.

tests/st/calicoctl/test_crud.py
```diff
@@ -35,745 +35,1087 @@ class TestPool(TestBase):
...
-        self.writeyaml('/tmp/%s-1.yaml' % res_type, data)
+        # Replace the Network Policy (with no resource version) and get it to
+        # assert the resource version is not the same.
+        rc = calicoctl("replace", data=networkpolicy_name1_rev2)
+        rc.assert_no_error()
+        rc = calicoctl(
+            "get networkpolicy %s -o yaml" % name(networkpolicy_name1_rev2))
+        rc.assert_no_error()
+        rev1 = rc.decoded
+        self.assertNotEqual(rev0, rev1)
```
gunjan5: the same comment applies here (and rest of the tests), `resourceVersion` is not explicitly being asserted (different from what the comment says)

robbrockbank: ah oops.

I've fixed other instances and squashed.  just waiting for tests to pass

PR #1718, opened 2017-10-12
add two named-port CRUD tests

bcreane:
## Description
Add a couple of named-port CRUD tests for calicoctl.

## Todos
- [x] Tests

## Release Note

```release-note
None required
```


fasaxc:
Both look like good tests, is there a reason that you broke them out into their own functions instead of adding them to the list of table-driven tests.  It's certainly easier to see what's going on with these but the table driven ones have their advantages too (any enhancement to the test gets applied to all tests).

bcreane:
Mostly simplicity and readability. The table driven tests are approaching an unwieldy scale. I can move over to the table-driven if you think its important.

PR #1719, opened 2017-10-15
Rev libcalico-go to latest 3.0-integration

robbrockbank:
Rev libcalico-go.


glide.yaml
```diff
@@ -42,3 +42,50 @@ import:
   - lib/validator
 - package: github.com/satori/go.uuid
 - package: github.com/termie/go-shutil
+- package: k8s.io/apimachinery
```
gunjan5: don't think we need to pin this

PR #1726, opened 2017-10-18
Apply didn't handle unsupported create operations

robbrockbank:

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -169,7 +169,10 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 	// Try to create the resource first.
 	ro, err := rh.Create(ctx, client, resource)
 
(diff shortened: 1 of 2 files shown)


calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -169,7 +169,8 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 	// Try to create the resource first.
 	ro, err := rh.Create(ctx, client, resource)
 
-	if _, ok := err.(cerrors.ErrorResourceAlreadyExists); ok {
+	switch err.(type) {
+	case cerrors.ErrorResourceAlreadyExists, cerrors.ErrorOperationNotSupported:
```
gunjan5: can you add a comment explaining why we're doing this? Not clear from reading the code why we're doing this

robbrockbank: i'll add to the comment that was already there shall I?

PR #1733, opened 2017-10-23
Re-enable Cluster Version and Type output

### LINKED ISSUE #1732: calicoctl version has stopped displaying calico version / cluster types
```
$ kubectl exec -ti -n kube-system calicoctl -- /calicoctl version
Client Version:    v2.0.0-alpha1-rc2
Build date:        2017-10-21T04:01:51+0000
Git commit:        87e8bd1f
```

tmjd:
## Description
Fixes #1732 

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/version.go
+++ b/calicoctl/commands/version.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2017 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,13 +15,15 @@
 package commands
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
 
 	"github.com/docopt/docopt-go"
-	//"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
 	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
+	"github.com/projectcalico/libcalico-go/lib/options"
 )
 
 var VERSION, BUILD_DATE, GIT_REVISION string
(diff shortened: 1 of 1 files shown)


calicoctl/commands/version.go
```diff
@@ -58,27 +61,19 @@ Description:
...
+	}
+	ctx := context.Background()
+	ci, err := client.ClusterInformation().Get(ctx, "default", options.GetOptions{})
+	if err != nil {
+		fmt.Println("Unable to retrieve Cluster Version or Type: ", err)
+	} else {
+		fmt.Println("Cluster Version:  ", ci.Spec.CalicoVersion)
+		fmt.Println("Cluster Type:     ", ci.Spec.ClusterType)
+	}
+
```
bcreane: nit - extra new line

calicoctl/commands/version.go
```diff
@@ -15,13 +15,16 @@
...
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
 
 	"github.com/docopt/docopt-go"
 	//"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
```
gunjan5: nit: should delete line 24 now.

calicoctl/commands/version.go
```diff
@@ -58,27 +61,18 @@ Description:
...
+	cf := parsedArgs["--config"].(string)
+	client, err := clientmgr.NewClient(cf)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	ctx := context.Background()
+	ci, err := client.ClusterInformation().Get(ctx, "default", options.GetOptions{})
+	if err != nil {
+		fmt.Println("Unable to retrieve Cluster Version or Type: ", err)
```
gunjan5: I think we should os.Exit(1) here if we get an error getting the ClusterInfo. WDYT?

calicoctl/commands/version.go
```diff
@@ -58,27 +61,18 @@ Description:
...
+	} else {
+		fmt.Println("Cluster Version:  ", ci.Spec.CalicoVersion)
```
gunjan5: We need to check if the values are assigned manually for both `ClusterVersion` and `ClusterType` print `"unknown"` if it's `""` like we use to before.

calicoctl/commands/version.go
tmjd: It would be nice I guess to have that return code. I had purposefully not exited there because I was thinking about more code being added after but I think I'd rather get that return code than plan for something that might never happen.

calicoctl/commands/version.go
tmjd: Is that what the previous code equated to? IIRC, previously the "" is what was returned as val if there was not a value set, so you could have "" in the fields but `assigned` would still be true and nothing would be printed.  I'll go check.

tmjd: Ok so the "" before was the node.
Previously if the `value` in the GetFelixConfig was nil then assigned would be false, otherwise true (it could be set empty).
So I don't know if it matters but it doesn't look like we really had this specifically as a case before. I feel like the fact that the Get on the ClusterInformation would give an error of `cerrors.ErrorResourceDoesNotExist` would equate to the previous check if the Config field was assigned or not. WDYT?

gunjan5: Now that we have a bunch of fields aggregated into a single resource `ClusterInformation` we can have some of the fields un-assigned (ClusterType) even when the `default` ClusterInformation resource exists, and the only way to know is to see if it's a `""` now since they're all strings

calicoctl/commands/version.go
gunjan5: Yes, I think we should exit here with the right exit code, if we add something else later we can return the OR'd exit code later

calicoctl/commands/version.go
```diff
@@ -58,27 +60,19 @@ Description:
...
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	ctx := context.Background()
+	ci, err := client.ClusterInformation().Get(ctx, "default", options.GetOptions{})
+	if err != nil {
+		fmt.Println("Unable to retrieve Cluster Version or Type: ", err)
+		os.Exit(1)
+	} else {
```
gunjan5: also, if we're exiting at line 73, we don't need this else

gunjan5:
just verified running old calicoctl version on an cluster without clusterType or version:
```
calicoctl-v1.4 version
Client Version:    v1.4.1
Build date:        2017-08-08T23:33:34+0000
Git commit:        a3e3c130
Cluster Version:   unknown
Cluster Type:      unknown
```

calicoctl/commands/version.go
```diff
@@ -58,27 +60,28 @@ Description:
...
+		os.Exit(1)
+	}
+
+	v := ci.Spec.CalicoVersion
+	if v == "" {
+		v = "unknown"
+	}
+	t := ci.Spec.ClusterType
+	if t == "" {
+		v = "unknown"
```
gunjan5: this should be `t = "unknown"` not `v`

gunjan5:
one last change then we're good to go

calicoctl/commands/version.go
tmjd: Good catch.

PR #1735, opened 2017-10-23
bump libcalico-go pin, update STs

bcreane:
## Description
[WIP]. fixup STs. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -369,6 +369,10 @@ class TestCalicoctlCommands(TestBase):
         if namespaced:
             data1['metadata']['namespace'] = "default"
 
+
+        if kind == "WorkloadEndpoint":
+            data1['metadata']['labels']['projectcalico.org/namespace'] = 'default'
+
         # Get the resource with name1 and namespace2.  For a namespaced
         # resource this should match the modified data to default the
         # namespace.  For non-namespaced resources this will error.
@@ -377,6 +381,9 @@ class TestCalicoctlCommands(TestBase):
             rc.assert_data(data1)
         else:
             rc.assert_error(NOT_NAMESPACED)
+
+        if kind == "WorkloadEndpoint":
+            data2['metadata']['labels']['projectcalico.org/namespace'] = 'namespace2'
 
         # Get the resource type for all namespaces.  For a namespaced resource
         # this will return everything.  For non-namespaced resources this will
--- a/tests/st/utils/data.py
+++ b/tests/st/utils/data.py
@@ -138,11 +138,9 @@ networkpolicy_name1_rev1 = {
                     'notNets': ['10.3.0.0/16'],
                     'notPorts': ['110:1050'],
                     'notSelector': "type=='apples'",
-                    'notTag': "bananas",
                     'nets': ['10.2.0.0/16'],
                     'ports': ['100:200'],
                     'selector': "type=='application'",
-                    'tag': 'alphatag'
                 },
                 'icmp': {'type': 10, 'code': 6},
                 'protocol': 'tcp',
@@ -150,11 +148,9 @@ networkpolicy_name1_rev1 = {
                     'notNets': ['10.1.0.0/16'],
                     'notPorts': [1050],
                     'notSelector': "type=='database'",
-                    'notTag': 'bartag',
                     'nets': ['10.0.0.0/16'],
                     'ports': [1234, '10:1024'],
                     'selector': "type=='application'",
-                    'tag': 'footag'
                 }
             }
         ],
@@ -217,11 +213,9 @@ globalnetworkpolicy_name1_rev1 = {
                     'notNets': ['10.3.0.0/16'],
                     'notPorts': ['110:1050'],
                     'notSelector': "type=='apples'",
-                    'notTag': "bananas",
                     'nets': ['10.2.0.0/16'],
                     'ports': ['100:200'],
                     'selector': "type=='application'",
-                    'tag': 'alphatag'
                 },
                 'icmp': {'type': 10, 'code': 6},
                 'protocol': 'tcp',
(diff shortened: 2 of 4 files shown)


tests/st/calicoctl/test_crud.py
```diff
@@ -443,9 +447,11 @@ def test_bgpconfig(self):
         Test CRUD commands behave as expected on the BGP configuration resource:
         """
         # Create a new default BGPConfiguration and get it to determine the current
-        # resource version.
+        # resource version. Ignore creation error since it's possible to create
+        # a default bgpconfig resource, but not to delete (so it may be hanging
+        # around).
         rc = calicoctl("create", data=bgpconfig_name1_rev1)
-        rc.assert_no_error()
+        #rc.assert_no_error()
```
caseydavenport: Do we need to command out these changes? I think we shouldn't be.

caseydavenport: (It's possible to create a BGP config, so it shouldn't error, yeah?)

tests/st/calicoctl/test_crud.py
```diff
@@ -486,8 +492,9 @@ def test_bgpconfig(self):
         rc.assert_error(SET_DEFAULT)
 
         # Delete the resource by name (i.e. without using a resource version).
+        # Note that deleting the default bgpconfig resource is disallowed.
         rc = calicoctl("delete bgpconfig %s" % name(rev3))
-        rc.assert_no_error()
+        #rc.assert_no_error()
```
caseydavenport: Is there an "assert_error" method we can call?

tests/st/calicoctl/test_crud.py
bcreane: yep, fixing - I was running inside the same `etcd` server instance, but `make st` tears that down for each run.

tests/st/calicoctl/test_crud.py
bcreane: actually this works. I think my error and fix were caused by re-running with the same `etcd` server instance.

tests/st/calicoctl/test_crud.py
```diff
@@ -378,6 +378,10 @@ def test_namespace(self, data, namespaced):
         else:
             rc.assert_error(NOT_NAMESPACED)
 
+        # The label will be updated with the new namespace 
+        if kind == "WorkloadEndpoint" and namespaced:
+            data2['metadata']['labels']['projectcalico.org/namespace'] = "default"
```
robbrockbank: don't we also have the orchestrator label too?

robbrockbank: workload endpoint is always namespaced

tests/st/utils/data.py
```diff
@@ -375,9 +363,13 @@
     'apiVersion': API_VERSION,
     'kind': 'WorkloadEndpoint',
     'metadata': {
+        'labels': {
```
robbrockbank: namespace should be namespace1 according to the spec

tests/st/utils/data.py
```diff
@@ -395,9 +387,13 @@
...
```
robbrockbank: ditto.

and orchestrator should be cni.

robbrockbank: in fact calicoctl should basically overwrite any value for orchestrator and namespace that you assign and fix them to the values in the values specified in the Metadata and Spec.

tests/st/calicoctl/test_crud.py
robbrockbank: ah i see we're tweaking the namespaces in the test.

ok from what I can see the namespace for data2 should be "namespace2"

bcreane: will fix

tests/st/utils/data.py
bcreane: will fix

tests/st/calicoctl/test_crud.py
robbrockbank: it feels like we should just update the data1 and data2 to have what is specified in the Metadata and Spec.

```
data1['metadata']['labels']['projectcalico.org/orchestrator'] = data1['spec']['orchestrator']
data1['metadata']['labels']['projectcalico.org/namespace'] = data1['metadata']['namespace']
data2['metadata']['labels']['projectcalico.org/orchestrator'] = data2['spec']['orchestrator']
data2['metadata']['labels']['projectcalico.org/namespace'] = data2['metadata']['namespace']
```

This would allow us to put in some more test data to check that the values in the supplied data are being overwritten with the correct values.

tests/st/utils/data.py
robbrockbank: well it may not be necessary - since the test could actually check that the values are updated to their correct values... not sure.

tests/st/utils/data.py
```diff
@@ -395,9 +387,13 @@
...
+            'projectcalico.org/namespace': 'namespace2',
```
robbrockbank: namespace1

tests/st/utils/data.py
robbrockbank: actually scrap that - i think both of these WEP definitions need to be valid because there are several tests that use these defs - so unless you wanna fix up a bunch of tests it's prolly easier to have these valid.

tests/st/utils/data.py
```diff
@@ -217,23 +213,19 @@
                     'notNets': ['10.3.0.0/16'],
                     'notPorts': ['110:1050'],
                     'notSelector': "type=='apples'",
-                    'notTag': "bananas",
```
gunjan5: 💯

robbrockbank: hmmm, are we not testing with this resource def?  If not why is this not failing tests?

robbrockbank: oh lordy I misread this.  Ignore me

PR #1736, opened 2017-10-25
Metadata handling for immutable fields

mgleung:
## Description
Added the handling for metadata fields that are populated by the datastore so they should not be overwritten by any calicoctl operations. The fields that are the exception to this rule is `Labels` and `Annotations`.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -188,7 +188,8 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 // through to the resource helper specific Create method which will map the untyped call to
 // the typed interface on the client.
 func (rh resourceHelper) Create(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceObject, error) {
-	return rh.create(ctx, client, resource)
+	resourceCopy := prepareMetadataForCreate(resource)
+	return rh.create(ctx, client, resourceCopy)
 }
 
 // Update is an un-typed method to update an existing resource. This calls the resource
@@ -201,8 +202,18 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 	// Check to see if the resourceVersion is specified in the resource object.
 	rv := resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
 
-	// If the resourceVersion is specified then we don't try to Get it or retry.
+	// If the resourceVersion is specified then we use it to try and update the resource.
+	// Do not attempt to retry if the resource version is specified.
 	if rv != "" {
+		// Validate the metadata is not changed for the the resource.
+		ro, err := rh.get(ctx, client, resource)
+		if err != nil {
+			return ro, err
+		}
+		// Set the specific resource version we are attempting to update.
+		ro.GetObjectMeta().SetResourceVersion(rv)
+		resource = mergeMetadataForUpdate(ro, resource)
+
 		return rh.update(ctx, client, resource)
 	}
 
@@ -216,8 +227,7 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 			return ro, err
 		}
 
-		// Set the resource-to-be-updated's resourceVersion to the one we got from the Get call.
-		resource.(ResourceObject).GetObjectMeta().SetResourceVersion(ro.GetObjectMeta().GetResourceVersion())
+		resource = mergeMetadataForUpdate(ro, resource)
 
 		// Try to update with the resource with the updated resourceVersion.
 		ru, err := rh.update(ctx, client, resource)
@@ -507,4 +517,44 @@ func (rh resourceHelper) GetTableTemplate(headings []string, printNamespace bool
 	}
 
 	return buf.String(), nil
+}
+
+// mergeMetadataForUpdate merges the Metadata for a stored ResourceObject and a potentail
+// update. All metadata in the potential update will be overwritten by the stored object
+// except for Labels and Annotations. This prevents accidental modifications to the metadata
+// fields by forcing updates to those fields to be handled by internal or more involved
+// processes.
+func mergeMetadataForUpdate(old, new ResourceObject) ResourceObject {
+	sm := old.GetObjectMeta()
+	cm := new.GetObjectMeta()
+
+	// Set the fields that are allowed to be overwritten (Labels and Annotations)
+	// so that they will not be overwritten.
+	sm.SetAnnotations(cm.GetAnnotations())
+	sm.SetLabels(cm.GetLabels())
+
+	sm.(*v1.ObjectMeta).DeepCopyInto(cm.(*v1.ObjectMeta))
+	return new
+}
+
+// prepareMetadataForCreate removes the metadata fields that should not be set from
+// calicoctl. Only the metadata fields Name, Namespace, ResourceVersion, Labels,
+// and Annotations will be kept. All other fields will be set elsewhere if required.
+// This prevents accidental modifications to the metadata fields by forcing updates
+// to those fields to be handled by internal or more involved processes.
+func prepareMetadataForCreate(r ResourceObject) ResourceObject {
+	rom := r.GetObjectMeta()
+	meta := &v1.ObjectMeta{}
+
+	// Save the important fields in the meta before everything gets wiped out.
+	meta.Name = rom.GetName()
+	meta.Namespace = rom.GetNamespace()
+	meta.ResourceVersion = rom.GetResourceVersion()
+	meta.Labels = rom.GetLabels()
+	meta.Annotations = rom.GetAnnotations()
+
+	// Make a copy of the resource so the input does not get modified
+	resOut := r.DeepCopyObject().(ResourceObject)
+	meta.DeepCopyInto(resOut.GetObjectMeta().(*v1.ObjectMeta))
+	return resOut
 }
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -378,6 +378,10 @@ class TestCalicoctlCommands(TestBase):
             data2['metadata']['name'] = data1['metadata']['name'][:len(data1['metadata']['name'])-1] + "1"
             # Change endpoint to eth1 so the validation works on the WEP
             data2['spec']['endpoint'] = "eth1"
+        elif kind == "IPPool":
+            data1['metadata']['name'] = "name1"
+            data2['metadata']['name'] = "name2"
+            data2['spec']['cidr'] = "10.10.1.0/24"
         else:
             data1['metadata']['name'] = "name1"
             data2['metadata']['name'] = "name2"
@@ -477,8 +481,8 @@ class TestCalicoctlCommands(TestBase):
         self.assertNotEqual(rev0['metadata']['resourceVersion'], rev1['metadata']['resourceVersion'])
 
         # Attempt to delete the default resource by name (i.e. without using a resource version).
-        rc = calicoctl("delete bgpconfig %s" % name(rev0))
-        rc.assert_no_error()
+        # rc = calicoctl("delete bgpconfig %s" % name(rev0))
+        # rc.assert_error(DELETE_DEFAULT)
 
         rc = calicoctl("create", data=bgpconfig_name2_rev1)
         rc.assert_no_error()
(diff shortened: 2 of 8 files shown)


calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +518,103 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+	cm := changed.GetObjectMeta()
+
+	// Set the fields that are allowed to be overwritten (Labels and Annotations)
+	// so that they will not be overwritten.
+	// TODO: Figure out if a warning is needed
+	sm.SetAnnotations(cm.GetAnnotations())
+	sm.SetLabels(cm.GetLabels())
+
+	// Make a copy of the resource so we do not need to modify the input
+	resOut := changed.DeepCopyObject().(ResourceObject)
```
robbrockbank: Does it matter if we modify the input in this case, I don't think anything else will need the unprocessed "changed" resource.

robbrockbank: or is this thinking ahead to the editor or something?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +518,103 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+	/*
+		if sm.GetNamespace() != cm.GetNamespace() {
+			log.Warning("Cannot change Metadata.Namespace. Reverting to past value.")
+			cm.SetNamespace(sm.GetNamespace())
+		}
+		if sm.GetName() != cm.GetName() {
+			log.Warning("Cannot change Metadata.Name. Reverting to past value.")
+			cm.SetName(sm.GetName())
+		}
+		if sm.GetGenerateName() != cm.GetGenerateName() {
```
robbrockbank: Personally I'd keep things simple and not bother warning about fields they can't change.  I'd probably just add some comments to the docs to list which fields can and cannot be modified through calicoctl.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +518,103 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
 
 	return buf.String(), nil
 }
+
+// validateMetadata compares the Metadata for a ResourceObject and warns if anything was
+// changed that will not be changed as a part of the update. Any discrepancies to the metadata
+// will be changed to match the metadata available in the stored object except for Labels and
+// Annotations.
+func validateMetadata(stored, changed ResourceObject) ResourceObject {
```
robbrockbank: This isn't just validate (and if we remove the warnings won't be validate at all).

I think the method name should indicate that this is merging ObjectMetadata for an Update.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +518,103 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+			cm.SetClusterName(sm.GetClusterName())
+		}
+		return changed
+	*/
+}
+
+// clearInvalidMetadata removes the metadata fields that should not be set from
+// calicoctl. Any metadata fields that need to be populated except for Name, Namespace,
+// ResourceVersion, Labels, and Annotations will be set elsewhere.
+func clearInvalidMetadata(r ResourceObject) ResourceObject {
```
robbrockbank: Similarly, I think this method might be better off indicating that it's preparing the ObjectMetadata for a Create (I don't think this is called from anywhere else).

robbrockbank: Also the comment says ResourceVersion, Labels etc. are set elsewhere, but that is not true.

calicoctl/resourcemgr/resourcemgr.go
mgleung: Fair enough. I'll get rid of the warnings.

calicoctl/resourcemgr/resourcemgr.go
mgleung: I did not have a specific feature in mind when doing this (though the editor may definitely run into this at a future date). It shouldn't make a functional difference now, just thought it would keep things cleaner if we didn't have to worry about the side effects of copying to and from the metadata.  I can remove it, but I think it might be nice to leave the return value so its obvious which resource is the one to use in the update step.

calicoctl/resourcemgr/resourcemgr.go
mgleung: The comment should say "except" for the relevant fields, but its not a very clear comment, I'll rewrite it so it is easier to read.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +515,40 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
 
 	return buf.String(), nil
 }
+
+// overwriteMetadata compares the Metadata for a ResourceObject and warns if anything was
```
robbrockbank: This comment is not true.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +515,40 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+// changed that will not be changed as a part of the update. Any discrepancies to the metadata
+// will be changed to match the metadata available in the stored object except for Labels and
+// Annotations.
+func overwriteMetadata(stored, changed ResourceObject) ResourceObject {
```
robbrockbank: How about `old, new` rather than `stored, changed` - it's a little more standard.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -505,3 +515,40 @@ func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
...
+	sm.SetLabels(cm.GetLabels())
+
+	sm.(*v1.ObjectMeta).DeepCopyInto(cm.(*v1.ObjectMeta))
+	return changed
+}
+
+// prepCreateMetadata removes the metadata fields that should not be set from
+// calicoctl. Only the metadata fields Name, Namespace, ResourceVersion, Labels,
+// and Annotations will be kept. All other fields will be set elsewhere if required.
+func prepCreateMetadata(r ResourceObject) ResourceObject {
```
robbrockbank: Given that these methods are for Update and Create, can we just make that a bit clearer.  How about `mergeMetadataForUpdate` and `prepareMetadataForCreate`

robbrockbank: Both of these methods serve a related purpose and that is to disallow modifying any of the metadata fields except the Labels and Annotations.  All other fields may be set internally or through other mechanism, but calicoctl will disregard any values supplied through the YAML/JSON files and ensure the existing values for the other other fields is maintained.

I think the comments could do with beefing up with the details.

tests/st/calicoctl/test_crud.py
```diff
@@ -557,6 +557,47 @@ def test_clusterinfo(self):
...
+        """
+        Test that the metadata fields other than labels and annotations cannot be changed
+        in create and update operations.
+        """
+        # Use apply to create a new Host Endpoint and get it to determine the
+        # current resource version (first checking that it doesn't exist).
+        rc = calicoctl(
+            "get hostendpoint %s -o yaml" % name(hostendpoint_name1_rev2))
+        rc.assert_error(text=NOT_FOUND)
+
```
robbrockbank: Probably have the same test using create/update rather than apply/apply.

tests/st/utils/utils.py
```diff
@@ -15,6 +15,7 @@
 import os
 import socket
 import sys
+from datetime import datetime
```
robbrockbank: the clean_calico_data should be updated to remove the UUID and creation timestamp fields - that way the assert_data doesn't need to worry about those values being set.

tests/st/calicoctl/test_crud.py
```diff
@@ -557,6 +557,47 @@ def test_clusterinfo(self):
...
+        rc = calicoctl("apply", data=hostendpoint_name1_rev3)
+        rc.assert_no_error()
+        rc = calicoctl(
+            "get hostendpoint %s -o yaml" % name(hostendpoint_name1_rev3))
+        rc.assert_no_error()
+        rev1 = rc.decoded
+        self.assertNotEqual(rev0['metadata']['resourceVersion'], rev1['metadata']['resourceVersion'])
+
+        # Validate that only annotations and labels were changed in the metadata
+        self.assertNotEqual(rev1['metadata'].get('selfLink', ''), hostendpoint_name1_rev3['metadata'].get('selfLink', ''))
```
robbrockbank: i think it's only worth checking the UID and Creation Timestamp for now.  We can add in the other checks if we manage to come up with a way to set them - but for now it makes the code harder to read.

robbrockbank: I think we should be asserting that UID and creation timestamp are both assigned in rev0 and rev1.

tests/st/calicoctl/test_crud.py
```diff
@@ -557,6 +557,47 @@ def test_clusterinfo(self):
...
+        self.assertNotEqual(rev1['metadata'].get('uid', ''), hostendpoint_name1_rev3['metadata'].get('uid', ''))
+        self.assertNotEqual(rev1['metadata'].get('generation', ''), hostendpoint_name1_rev3['metadata'].get('generation', ''))
```
robbrockbank: I don't see the point in comparing to the original data since we know calicoctl is not going to set any of the metadata fields from that data except the annotations and labels.

tests/st/utils/data.py
```diff
@@ -289,6 +289,25 @@
...
 }
 
+hostendpoint_name1_rev3 = {
+    'apiVersion': API_VERSION,
+    'kind': 'HostEndpoint',
+    'metadata': {
+        'name': 'endpoint1',
+        'labels': {'type': 'frontend', 'misc': 'version1'},
+        'annotations': {'key': 'value'},
+        'selfLink': 'test-self-link',
```
robbrockbank: I think this should include uuid and creation timestamp to check that the values in the YAML are not stored as well.

calicoctl/resourcemgr/resourcemgr.go
mgleung: Oops I didn't read closely enough

tests/st/calicoctl/test_crud.py
mgleung: The thing I was trying to prove here was that they weren't set so that they wouldn't match the original but that's a little hokey.  I think I'll change the assertion to just say that the keys don't exist in the `metadata`

tests/st/calicoctl/test_crud.py
```diff
@@ -565,6 +569,109 @@ def test_clusterinfo(self):
...
+
+        # Validate that creationTimestamp and UID are unchanged from when they were created
+        self.assertEqual(rev1['metadata']['creationTimestamp'], rev0['metadata']['creationTimestamp'])
+        self.assertEqual(rev1['metadata']['uid'], rev0['metadata']['uid'])
+
+        # Delete the resource without using a resource version.
+        rc = calicoctl("delete hostendpoint %s" % name(rev1))
+        rc.assert_no_error()
+
+    def test_metadata_unchanged_create_apply(self):
```
robbrockbank: I was thinking one this one would be create/replace rather than using apply.

robbrockbank: hmmm, in k8s - does a `replace` act as if the resource is a completely new object with a new UUID and new creation timestamp?

tests/st/calicoctl/test_crud.py
```diff
@@ -565,6 +569,109 @@ def test_clusterinfo(self):
...
+        rc = calicoctl("apply", data=hostendpoint_name1_rev3)
+        rc.assert_no_error()
+        rc = calicoctl(
+            "get hostendpoint %s -o yaml" % name(hostendpoint_name1_rev3))
+        rc.assert_no_error()
+        rev1 = rc.decoded
+        self.assertNotEqual(rev0['metadata']['resourceVersion'], rev1['metadata']['resourceVersion'])
+
+        # Validate that only annotations and labels were changed in the metadata
+        self.assertEqual('selfLink' in rev1['metadata'], False)
```
robbrockbank: There are better assert methods to use here, see:  https://docs.python.org/2/library/unittest.html#assert-methods

PR #1737, opened 2017-10-25
Update release instructions to be more clear

caseydavenport:
## Description

Make it abundantly clear not to push the latest images for an RC or alpha release. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -241,15 +241,25 @@ endif
...
+	@echo "- dist/calicoctl-darwin-amd64"
+	@echo "- dist/calicoctl-windows-amd64.exe"
+	@echo "# To find commit messages for the release notes:  git log --oneline <old_release_version>...$(VERSION)"
+	@echo ""
+	@echo "# Now push the newly created release images."
+	@echo "  docker push calico/ctl:$(VERSION)"
+	@echo "  docker push quay.io/calico/ctl:$(VERSION)"
+	@echo ""
+	@echo "# For the final release only, push the latest tag"
+	@echo "# DO NOT PUSH THESE IMAGES FOR RELEASE CANDIDATES OR ALPHA RELEASES"
```
robbrockbank: I would go one step further and prefix these lines with a "- "  so you can't blindly copy and paste it in.  But I'm ok without that too.

PR #1739, opened 2017-10-26
Bcreane bump libcalico go

caseydavenport:
## Description
A few sentences describing the overall goals of the pull request's commits.
Please include
- the type of fix - (e.g. bug fix, new feature, documentation)
- some details on _why_ this PR should be merged
- the details of the testing you've done on it (both manual and automated)
- which components are affected by this PR
- links to issues that this PR addresses

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -20,7 +20,7 @@ from nose_parameterized import parameterized
 from tests.st.test_base import TestBase
 from tests.st.utils.utils import log_and_run, calicoctl, \
     API_VERSION, name, ERROR_CONFLICT, NOT_FOUND, NOT_NAMESPACED, \
-    DELETE_DEFAULT, SET_DEFAULT, NOT_SUPPORTED
+    SET_DEFAULT, NOT_SUPPORTED
 from tests.st.utils.data import *
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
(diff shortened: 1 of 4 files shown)


bcreane:
@caseydavenport - thanks for fixing up tests. I only checked `test-containerized` but skipped UTs.

tests/st/calicoctl/test_crud.py
```diff
@@ -470,7 +470,7 @@ def test_bgpconfig(self):
 
         # Attempt to delete the default resource by name (i.e. without using a resource version).
         rc = calicoctl("delete bgpconfig %s" % name(rev0))
-        rc.assert_error(DELETE_DEFAULT)
+        rc.assert_no_error()
```
gunjan5: How would deleting the default bgpconfig succeed? I thought we don't let anyone delete it from calicoctl, did we change that?

bcreane: yeah, just checked and it's now possible to remove the default bgp resource. maybe @robbrockbank has the info.

gunjan5: just talked to Rob, we have disabled it for now, but for some tests. I'm not really convinced that's a good reason to disable the check, but for alpha release, it's not necessary and we have a backlog item to rethink this, so I'm not going to block this

bcreane:
Was going to squash, but okay.

PR #1746, opened 2017-10-30
Change namespace output to more closely match K8s

tmjd:
Print namespaces as the first column.
Only print namespace column if getting 'all-namespaces' or a specific namespace.

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/printer.go
+++ b/calicoctl/commands/printer.go
@@ -88,6 +88,10 @@ type resourcePrinterTable struct {
 	// Wide format.  When headings have not been explicitly specified, this is used to
 	// determine whether to the resource-specific default wide or narrow headings.
 	wide bool
+
+	// Namespace included. When a resource being printed is namespaced, this is used
+	// to determine if the namespace column should be printed or not.
+	printNamespace bool
 }
 
 func (r resourcePrinterTable) print(client client.Interface, resources []runtime.Object) error {
@@ -104,7 +108,7 @@ func (r resourcePrinterTable) print(client client.Interface, resources []runtime
 		}
 
 		// Look up the template string for the specific resource type.
-		tpls, err := rm.GetTableTemplate(headings)
+		tpls, err := rm.GetTableTemplate(headings, r.printNamespace)
 		if err != nil {
 			return err
 		}
--- a/calicoctl/resourcemgr/networkpolicy.go
+++ b/calicoctl/resourcemgr/networkpolicy.go
@@ -29,7 +29,8 @@ func init() {
 		true,
 		[]string{"networkpolicy", "networkpolicies", "policy", "np", "policies", "pol", "pols"},
 		[]string{"NAME"},
-		[]string{"NAME", "ORDER", "SELECTOR", "NAMESPACE"},
+		[]string{"NAME", "ORDER", "SELECTOR"},
+		// NAMESPACE may be prepended in GrabTableTemplate so needs to remain in the map below
 		map[string]string{
 			"NAME":      "{{.ObjectMeta.Name}}",
 			"NAMESPACE": "{{.ObjectMeta.Namespace}}",
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -44,7 +44,7 @@ import (
 //	-  Commands to manage resource instances through an un-typed interface.
 type ResourceManager interface {
 	GetTableDefaultHeadings(wide bool) []string
-	GetTableTemplate(columns []string) (string, error)
+	GetTableTemplate(columns []string, printNamespace bool) (string, error)
 	GetObjectType() reflect.Type
 	IsNamespaced() bool
 	Apply(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceObject, error)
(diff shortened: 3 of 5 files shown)


tmjd:
Here is some sample output with those changes.

```
$ calicoctl get networkpolicy -o wide -n production
NAMESPACE    NAME             ORDER   SELECTOR             
production   allow-tcp-6379   <nil>   role == 'database' 

$ calicoctl get networkpolicy -o wide
NAME   ORDER   SELECTOR   

$calicoctl get networkpolicy -o wide --all-namespaces
NAMESPACE    NAME             ORDER   SELECTOR             
production   allow-tcp-6379   <nil>   role == 'database'   

$ calicoctl get networkpolicy --all-namespaces
NAMESPACE    NAME             
production   allow-tcp-6379   

$ calicoctl get networkpolicy
NAME   

```

calicoctl/resourcemgr/networkpolicy.go
```diff
@@ -29,7 +29,7 @@ func init() {
 		true,
 		[]string{"networkpolicy", "networkpolicies", "policy", "np", "policies", "pol", "pols"},
 		[]string{"NAME"},
-		[]string{"NAME", "ORDER", "SELECTOR", "NAMESPACE"},
```
gunjan5: Is there a good reason for removing the namespace from `-o wide` besides that's what k8s does? I find having namespace field in `-o wide` very useful

calicoctl/commands/printer.go
```diff
@@ -88,6 +88,10 @@ type resourcePrinterTable struct {
 	// Wide format.  When headings have not been explicitly specified, this is used to
 	// determine whether to the resource-specific default wide or narrow headings.
 	wide bool
+
+	// Namespace included. When a resource is being printed is namespaced this is used
+	// to determine if the namespace column should be printed or not.
+	namespace bool
```
gunjan5: I think this should be called `printNamespace`

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -467,7 +467,11 @@ func (rh resourceHelper) GetTableDefaultHeadings(wide bool) []string {
 // GetTableTemplate constructs the go-lang template string from the supplied set of headings.
 // The template separates columns using tabs so that a tabwriter can be used to pretty-print
 // the table.
-func (rh resourceHelper) GetTableTemplate(headings []string) (string, error) {
+func (rh resourceHelper) GetTableTemplate(headings []string, namespace bool) (string, error) {
+
```
gunjan5: nit: extra newline

calicoctl/resourcemgr/networkpolicy.go
gunjan5: If we keep the `NAMESPACE` field in `-o wide` output, this could be super simple (just move `NAMESPACE` first on this list, and we're done (since calicoctl already prints the correct namespace, just not in the first column). And, I think having the namespace in `-o wide` would be super useful, we can also add the namespace field for the normal `ps` output.

gunjan5:
@tmjd besides my comments, can you make sure this works fine if you specify `--all-namespaces` for a non-namespaced resource such as bgppeer? (It should fail with an error saying you can't use this flag for a non-namespaced resource)

calicoctl/resourcemgr/networkpolicy.go
tmjd: Yeah the only reason is to copy how K8s does it. As you can see in the output you basically only get no namespace when you aren't specifying the namespace or requesting all-namespaces. My initial thought was just to move Namespace to the first column, @caseydavenport requested it behave like kubectl.

tmjd:
Output if specifying `--all-namespaces` for a non-namespaced resource.
```
$ calicoctl get bgppeer --all-namespaces
Failed to get resources: BGPPeer is not namespaced
```

calicoctl/resourcemgr/networkpolicy.go
```diff
@@ -29,7 +29,7 @@ func init() {
...
+		[]string{"NAME", "ORDER", "SELECTOR"},
 		map[string]string{
 			"NAME":      "{{.ObjectMeta.Name}}",
 			"NAMESPACE": "{{.ObjectMeta.Namespace}}",
```
gunjan5: can you add a quick comment here saying NAMESPACE is implicitly added to the template in  GetTableTemplate so humans don't delete this in future thinking it's not in the headings so we don't need it

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -44,7 +44,7 @@ import (
 //	-  Commands to manage resource instances through an un-typed interface.
 type ResourceManager interface {
 	GetTableDefaultHeadings(wide bool) []string
-	GetTableTemplate(columns []string) (string, error)
+	GetTableTemplate(columns []string, namespace bool) (string, error)
```
gunjan5: this should also be renamed to `printNamespace` for consistency

calicoctl/resourcemgr/networkpolicy.go
tmjd: Another option I considered was to have namespace in the ps and wide and then in the GetTableTemplate remove the field from the headings if printNamespace was false. Would that be preferable?

gunjan5: I think this is fine, not too fussed either way tho.

PR #1747, opened 2017-10-31
Pin go-build to v0.9 to pick up CVE fixes and go 1.9.2.

fasaxc:
## Description

Pin the calico/go-build container to v0.9 to pick up go 1.9.2 and CVE fixes.

I do think we should pin to particular releases of go-build to ensure build consistency and repeatability of older builds.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ OS := $(shell uname -s | tr A-Z a-z)
 ARCH := amd64
 ###############################################################################
 # Subcomponent versions:
-GO_BUILD_VER:=latest
+GO_BUILD_VER:=v0.9
 ###############################################################################
 
 CALICOCTL_VERSION?=$(shell git describe --tags --dirty --always)


Makefile
```diff
@@ -15,7 +15,7 @@ OS := $(shell uname -s | tr A-Z a-z)
 ARCH := amd64
 ###############################################################################
 # Subcomponent versions:
-GO_BUILD_VER:=latest
+GO_BUILD_VER:=v0.9
```
bcreane: @fasaxc - it looks like `quay.io/calico/go-build:latest` was updated on quay.io earlier today, but the hash doesn't match up with anything else. Do you know what's been in `latest` and how long calicoctl has been building with it?

fasaxc: Looks like that's `master`

PR #1750, opened 2017-11-01
Fill in all felixConfiguration field names for CRUD test.

bcreane:
## Description
Fill in all felixConfiguration field names for CRUD test.

Note this test will fail until https://github.com/projectcalico/libcalico-go/pull/645 is accepted.
## Release Note

```release-note
None required
```


tests/st/utils/data.py
```diff
@@ -496,9 +496,43 @@
         'name': 'felixconfiguration1',
     },
     'spec': {
+        'chainInsertMode': 'append',
+        'defaultEndpointToHostAction': 'ACCEPT',
+        'failsafeInboundHostPorts': [
+            {'Protocol': 'tcp', 'Port': 666},
```
mgleung: We may want to add json tags to the `ProtoPort` objects in libcalico-go so that we don't have to use the go representation of the variables since the capitalization in the JSON looks a little funny.

caseydavenport: +1

keys should start with a lowercase

PR #1756, opened 2017-11-08
Add the export option in calicoctl to allow users to be able to apply calicoctl get output

mgleung:
## Description
Add the "export" option modeled after the K8s option. This will allow the calicoctl "get" output to be applied without needing modification.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note
```release-note
Added --export flag to the calicoctl get command to allow for resources to be applied across clusters strictly from the calicoctl output.
```

### DIFF
--- a/calicoctl/commands/resources.go
+++ b/calicoctl/commands/resources.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 
+	"context"
 	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
 	"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
 	"github.com/projectcalico/calicoctl/calicoctl/resourcemgr"
@@ -27,8 +28,8 @@ import (
 	calicoErrors "github.com/projectcalico/libcalico-go/lib/errors"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"context"
 )
 
 type action int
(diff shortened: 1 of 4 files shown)


bcreane:
Definitely worth waiting for @gunjan5 's review, however this PR looks good to me. If he doesn't get to it before vacation, I'll approve.

calicoctl/commands/resources.go
```diff
@@ -188,6 +189,21 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
 			results.err = err
 			break
 		}
+
+		for i, _ := range res {
+			// Remove the cluster specific metadata if the "--export" flag is specified
+			if argutils.ArgBoolOrFalse(args, "--export") {
```
gunjan5: I think you can store this in a variable outside the loop so you don't have to keep parsing it for every loop iteration

tests/st/calicoctl/test_crud.py
```diff
@@ -624,6 +624,55 @@ def test_metadata_unchanged(self, create_cmd, update_cmd):
...
+        self.assertIn('uid', rev0['metadata'])
+        self.assertIn('creationTimestamp', rev0['metadata'])
+        self.assertEqual(rev0['metadata']['name'], networkpolicy_name2_rev1['metadata']['name'])
+        self.assertEqual(rev0['metadata']['namespace'], networkpolicy_name2_rev1['metadata']['namespace'])
+        self.assertIn('resourceVersion', rev0['metadata'])
+
+        # Retrieve the Network Policy with the export flag and
+        # verify that cluster-specific information is not present
+        rc = calicoctl(
+            "get networkpolicy %s -o yaml --export" % name(networkpolicy_name2_rev1))
```
gunjan5: does this work with the list operation as well? (when you don't specify the resource name), looks like k8s doesn't do anything when `--export` is specified on a list (get without a resource name/namespace)

calicoctl/commands/resources.go
```diff
@@ -188,6 +189,21 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
...
+				rom := res[i].(v1.ObjectMetaAccessor).GetObjectMeta()
+				rom.SetNamespace("")
```
gunjan5: didn't know k8s also removes the namespace with `--export`, but I just checked it and it does, should be ok because both of our namespaced resources are generated from k8s resources

gunjan5:
@mgleung looking pretty good! You'll also need to update this https://docs.projectcalico.org/master/reference/calicoctl/commands/get

tests/st/calicoctl/test_crud.py
mgleung: Good point, this would currently work with the list operation (at least based off my understanding of the code) and it probably shouldn't to match k8s.

mgleung: I changed this functionality to silently ignore the `--export` flag if a name is not provided. From my testing, this should mirror k8s.

calicoctl/commands/resources.go
```diff
@@ -182,12 +183,34 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
...
 		res, err := executeResourceAction(args, client, r, action)
 		if err != nil {
 			results.err = err
 			break
 		}
+
+		for i, _ := range res {
+			// Remove the cluster specific metadata if the "--export" flag is specified
+			if export {
+				if nameSpecified == "" {
```
gunjan5: line 195 to 202 can be simplified to:
```go
if exported && nameSpecified != "" {
     for i, _ := range res {
          rom := res[i].(v1.ObjectMetaAccessor).GetObjectMeta()
          ...
          ...
     }
}
```

mgleung: oops wrote it before I had tested if kubectl gives a log or anything for specifying `--export` without a name. I'll fix that up.

calicoctl/commands/resources.go
```diff
@@ -182,12 +183,31 @@ func executeConfigCommand(args map[string]interface{}, action action) commandRes
...
 		if err != nil {
 			results.err = err
 			break
 		}
+
+		for i, _ := range res {
+			// Remove the cluster specific metadata if the "--export" flag is specified
+			// Skip removing cluster specific metadata if this is is called as a "list"
+			// operation (no specific name is specified).
+			if export && nameSpecified != "" {
```
gunjan5: this should still go before line 195 since these things won't change during the execution of this code (you will either have the flag and name or not for the calicoctl command, it's not going to change for each resources)

PR #1759, opened 2017-11-10
Fix STs after reenabling validation

mgleung:
## Description
Fixing up the STs after reenabling the validations added in https://github.com/projectcalico/libcalico-go/pull/661 (which should now be moved to https://github.com/projectcalico/libcalico-go/pull/667)

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #1760, opened 2017-11-10
Fix STs after turning validation back on

heschlie:
## Description

Pulling from PR #1759 to continue fixing the validation rules after they have been fixed in libcalico-go.

Currently I've pinned libcalico-go to @bcreane PR to fixup the validations in the new API. We'll need to change this to his commit when ready.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -25,17 +25,17 @@ import (
 	"strings"
 	"time"
 
+	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
+	yamlsep "github.com/projectcalico/calicoctl/calicoctl/util/yaml"
+	yaml "github.com/projectcalico/go-yaml-wrapper"
+	client "github.com/projectcalico/libcalico-go/lib/clientv3"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+	validator "github.com/projectcalico/libcalico-go/lib/validator/v3"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-
-	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
-	yamlsep "github.com/projectcalico/calicoctl/calicoctl/util/yaml"
-	"github.com/projectcalico/go-yaml-wrapper"
-	client "github.com/projectcalico/libcalico-go/lib/clientv3"
-	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 )
 
 // The ResourceManager interface provides useful function for each resource type.  This includes:
@@ -366,10 +366,9 @@ func unmarshalResource(tm unstructured.Unstructured, b []byte) ([]runtime.Object
 	}
 
 	log.Infof("Type of unpacked data: %v", reflect.TypeOf(unpacked))
-	// TODO: Remember to uncommment this once libcalico has validation code.
-	// if err = validator.Validate(unpacked); err != nil {
-	// 	return nil, err
-	// }
+	if err = validator.Validate(unpacked); err != nil {
+		return nil, err
+	}
 
 	log.Infof("Unpacked: %+v", unpacked)
 
@@ -399,12 +398,11 @@ func unmarshalSliceOfResources(tml []unstructured.Unstructured, b []byte) ([]run
 
 	// Validate the data in the structures.  The validator does not handle slices, so
 	// validate each resource separately.
-	// TODO: uncomment this once libcalico has validation
-	// for _, r := range unpacked {
-	// 	if err := validator.Validate(r); err != nil {
-	// 		return nil, err
-	// 	}
-	// }
+	for _, r := range unpacked {
+		if err := validator.Validate(r); err != nil {
+			return nil, err
+		}
+	}
 
 	log.Infof("Unpacked: %+v", unpacked)
 
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -20,7 +20,7 @@ from nose_parameterized import parameterized
 from tests.st.test_base import TestBase
 from tests.st.utils.utils import log_and_run, calicoctl, \
     API_VERSION, name, ERROR_CONFLICT, NOT_FOUND, NOT_NAMESPACED, \
-    SET_DEFAULT, NOT_SUPPORTED, writeyaml
+    SET_DEFAULT, NOT_SUPPORTED, KUBERNETES_NP, writeyaml
 from tests.st.utils.data import *
 
 logging.basicConfig(level=logging.DEBUG, format="%(message)s")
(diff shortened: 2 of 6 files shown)


heschlie:
@bcreane I had only uncommented the tests that had a TODO for Validation, are we able to reactivate the other tests? More than happy to get them working if that's the case.

bcreane:
@heschlie - you're right, that looks like all the validation tests which are all that is required for this pr. thanks for getting this working. as soon as the libcalico-go fix goes in, you can fix up your glide.yaml and we should be able to merge it.

tests/st/calicoctl/test_crud.py
```diff
@@ -1455,369 +1455,354 @@ def test_export_flag(self):
...
-#             for data in testdata:
-#                 check_no_data_in_store(data)
-#         else:
-#             check_no_data_in_store(testdata)
-#
-#         # Cover the case where no data got stored, but calicoctl didn't fail:
-#         assert commanderror is True, "Failed - calicoctl did not fail to add invalid config"
-#
+class InvalidData(TestBase):
+    testdata = [
```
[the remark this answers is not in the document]
heschlie: Thanks @robbrockbank I've given it a once over for those cases, going to comb over it some more and make sure we're all up to date.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -25,17 +25,17 @@ import (
 	"strings"
 	"time"
 
+	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
```
robbrockbank: nit:  I think it's nice to organize imports in 3 groups.   Built-ins, External, projectcalico

robbrockbank: shouldn't stop this going in though

heschlie: Ah, I have my IDE setup to just run `go imports` on all modified files I think, do we need this before merging?

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                                'scope': 'node'}
+                   }),
+                   # See issue https://github.com/projectcalico/libcalico-go/issues/222
+                   ("bgpPeer-longname", {
+                       'apiVersion': API_VERSION,
+                       'kind': 'BGPPeer',
+                       'metadata': {'name':'bgppeer8'},
+                       'spec': {'asNumber': 64590,
+                                'node':
+                                    'TestTestTestTestTestTestTestTestTestTestTest'
```
robbrockbank: This won't be testing the field length.

Since node is also a name field, we don't allow caps.

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                                'selector': ""}}),
+                   ("policy-invalidReversedRange", {
+                       'apiVersion': API_VERSION,
+                       'kind': 'NetworkPolicy',
+                       'metadata': {'name': 'policy2'},
+                       'spec': {'Egress': [{'action': 'Deny',
+                                            'destination': {},
+                                            'source': {
+                                                'protocol': 'TCP',
+                                                'ports': [65535-1]  # range should be low-high
```
robbrockbank: I don't think that is valid format for port range, range is separated by a colon.

And i think it would need to be in quotes too.

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                                'order': 100000,
+                                'selector': ""}}),
+                   ("policy-invalidLowPortinRange", {
+                       'apiVersion': API_VERSION,
+                       'kind': 'NetworkPolicy',
+                       'metadata': {'name': 'policy2'},
+                       'spec': {'Egress': [{'action': 'Deny',
+                                            'destination': {},
+                                            'source': {
+                                                'ports': [0-65535],  # Min port is 1
```
robbrockbank: invalid port format (see below for comment), and other tests for culprits

robbrockbank:
Without checking the error message from the validation failures I reckon we're failing a lot of the validations for the wrong reason (i.e. not the reason intended by the test).

Strictly we'd be better searching the error message for some particular text to check the reason.

Perhaps worth raising an issue/backlog item to address that.

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                   ("pool-invalidNet2", {'apiVersion': API_VERSION,
+                                         'kind': 'IPPool',
+                                         'metadata': {},
+                                         'spec': {'ipip': {'mode': 'Always'},
+                                                  'cidr': "10.0.256.0/24"}  # invalid octet
+                                         }),
+                   ("pool-invalidNet3", {'apiVersion': API_VERSION,
+                                         'kind': 'IPPool',
+                                         'metadata': {},
+                                         'spec': {'ipip': {'mode': 'Always'},
```
robbrockbank: This is no longer the format for the IPPool.

There is now a top-level ipipMode field, rather than this sub-struct.

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                                         # nothing
+                                         }),
+                   ("pool-invalidNet7", {'apiVersion': API_VERSION,
+                                         'kind': 'IPPool',
+                                         'metadata': {},
+                                         'spec': {'cidr': "192.168.0.0/27"}  # invalid mask
+                                         }),
+                   ("pool-invalidNet8", {'apiVersion': API_VERSION,
+                                         'kind': 'IPPool',
+                                         'metadata': {},
```
robbrockbank: not having a name in the metadata will likely be the reason for this failing.  Same above.

robbrockbank: and below.

tests/st/calicoctl/test_crud.py
```diff
@@ -1479,369 +1479,366 @@ def test_get(self):
...
+                                                          'ICMP': {'type': 19,
+                                                                   'code': 256},  # max value 255
+                                                          'action': 'Deny',
+                                                          'destination': {},
+                                                          'source': {}}],
+                                             }}),
+                   ("compound-config", [{
+                       'apiVersion': API_VERSION,
+                       'kind': 'BGPPeer',
+                       'metadata': {},
```
robbrockbank: none of your metadata should really be empty.  unless that's what we're testing

heschlie:
Yea I wasn't the closest to a lot of these changes, I'm more than happy to get them fixed up ASAP after we merge or before, whichever seems best.

heschlie:
@robbrockbank @caseydavenport in the interest of getting things moving should we go ahead and get this merged? I've added a small fix from @neiljerram but as Rob mentioned a bunch of these tests need more digging into to be sure they are failing for the right reason, which could take awhile.

I can create an issue to fix those tests properly to get them sorted post merge. WDYGT?

tests/st/calicoctl/test_crud.py
heschlie: `scope` is also no longer a valid field in the `Spec` right?

robbrockbank:
I'm ok with this going in - I think I mentioned that above - as long as we have a task to fix them up after.

heschlie:
I _think_ I've fixed the tests so they are failing for the correct reasons (one is commented out at the moment for name length) going to open an issue to improve the tests to check the proper errors were returned.

PR #1765, opened 2017-11-16
Validation changes

neiljerram:


neiljerram:
D'oh, everything here is also already in #1760

PR #1779, opened 2017-12-05
Updating the InvalidData ST to check the returned error

### LINKED ISSUE #1766: Calicoctl STs for checking invalid manifests are rejected needs improvement
The STs will "pass" during the `test_invalid_profiles_rejected` tests if the data is rejected for the wrong reason. We should improve the framework to check the correct error was given.

## Expected Behavior

Invalid data tests tell you when the data is rejected, and verify that it was rejected for the correct reason.

## Current Behavior

If the data is rejected for any reason the test passes.

## Possible Solution

Pass in some info for what error should be returned?

## Steps to Reproduce (for bugs)

1. Adjust one of the cases to be invalid for any other reason than what it is tested for
2. Run `make st`
3. Watch tests still succeed

## Your Environment

* calicoctl version (e.g. `calicoctl version`):
* calico/node version (e.g. `docker run --rm calico/node versions`)
* Orchestrator type and version (e.g. kubernetes, mesos, rkt):
* Operating System and version:
* Link to your project (optional):

heschlie:
## Description

Fixes #1766
This PR adds the error messages to check for each failure in the `InvalidData` test so we can be sure that the test is passing for the correct reason.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -1479,6 +1479,7 @@ class TestCalicoctlCommands(TestBase):
 #         self.assert_same(exp_data, output)
 #
 
+
 class InvalidData(TestBase):
     testdata = [
                    ("bgpPeer-invalidASnum", {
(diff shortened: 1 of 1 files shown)


tests/st/calicoctl/test_crud.py
```diff
@@ -1670,7 +1676,8 @@ class InvalidData(TestBase):
                                              'protocol': 'UDP',
                                              'source': {}}],
                                 'order': 100000,
-                                'selector': ""}}),
+                                'selector': ""}
+                   }, "error with field action = 'jumpupanddown'"),
```
gunjan5: ⬆️ ⬇️  👍

PR #1780, opened 2017-12-05
Manifests for v1 to v3 migration testing

tmjd:
These manifests along with the ones already present in `test-data/v1` should be used to test out the offline migration functionality.

```release-note
None required
```

### DIFF
--- a/test-data/v1/migration/README.md
+++ b/test-data/v1/migration/README.md
@@ -1,0 +1,2 @@
+The test data here is to be used when testing out the offline migration from
+v1 to v3.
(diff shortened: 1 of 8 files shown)


tmjd:
Most (all) of these manifests I took from the v2.6 Resource Reference docs.
I had to modify the workloadendpoint manifest from what the example had, due to https://github.com/projectcalico/calico/issues/1459. Depending on the result of that issue we may want to update the workloadendpoint manifest.

tmjd:
Instead of reviewing this PR it might be better to checkout https://github.com/gunjan5/calico-containers/pull/2.  I've added these manifests there so I can use them in STs for the `calicoctl convert` command.

PR #1782, opened 2017-12-06
Add calicoctl convert command for manifest offline conversions

gunjan5:
## Description

example multi-resource manifest conversion:
Original:
```yaml
cat ../test-data/v1/multi-resource.yaml
- apiVersion: v1
  kind: bgpPeer
  metadata:
    node: Node1
    peerIP: aa:bb::ff
    scope: node
  spec:
    asNumber: 64514
- apiVersion: v1
  kind: bgpPeer
  metadata:
    node: Node2
    peerIP: 1.2.3.4
    scope: node
  spec:
    asNumber: 6455
---
apiVersion: v1
kind: bgpPeer
metadata:
  scope: global
  peerIP: 192.20.30.40
spec:
  asNumber: 64567

---
- apiVersion: v1
  kind: ipPool
  metadata:
    cidr: 192.168.0.0/16
  spec:
    ipip:
      enabled: true

- apiVersion: v1
  kind: ipPool
  metadata:
    cidr: 2001::00/120
  spec:
```

Conversion command:
```bash
calicoctl convert -f ../test-data/v1/multi-resource.yaml -o yaml
```
```yaml
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    creationTimestamp: null
    name: node1.00aa-00bb-0000-0000-0000-0000-0000-00ff
  spec:
    asNumber: 64514
    node: node1
    peerIP: aa:bb::ff
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    creationTimestamp: null
    name: node2.1-2-3-4
  spec:
    asNumber: 6455
    node: node2
    peerIP: 1.2.3.4
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    creationTimestamp: null
    name: 192-20-30-40
  spec:
    asNumber: 64567
    peerIP: 192.20.30.40
- apiVersion: projectcalico.org/v3
  kind: IPPool
  metadata:
    creationTimestamp: null
    name: 192-168-0-0-16
  spec:
    cidr: 192.168.0.0/16
    ipipMode: Always
- apiVersion: projectcalico.org/v3
  kind: IPPool
  metadata:
    creationTimestamp: null
    name: 2001---120
  spec:
    cidr: 2001::/120
    ipipMode: Never
```

Another example:

original yaml file:
```yaml
- apiVersion: v1
  kind: bgpPeer
  metadata:
    node: Node1
    peerIP: aa:bb::ff
    scope: node
  spec:
    asNumber: 64514
- apiVersion: v1
  kind: bgpPeer
  metadata:
    node: Node2
    peerIP: 1.2.3.4
    scope: node
  spec:
    asNumber: 6455
```

conversion:
```yaml
calicoctl convert -f ./bgpplist.yaml -o yaml
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    creationTimestamp: null
    name: node1.00aa-00bb-0000-0000-0000-0000-0000-00ff
  spec:
    asNumber: 64514
    node: node1
    peerIP: aa:bb::ff
- apiVersion: projectcalico.org/v3
  kind: BGPPeer
  metadata:
    creationTimestamp: null
    name: node2.1-2-3-4
  spec:
    asNumber: 6455
    node: node2
    peerIP: 1.2.3.4
```

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Add calicoctl convert command for manifest offline conversions
```

### DIFF
--- a/calicoctl/commands/convert.go
+++ b/calicoctl/commands/convert.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/docopt/docopt-go"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/v1resourceloader"
+	"github.com/projectcalico/libcalico-go/lib/apis/v1/unversioned"
+	conversion "github.com/projectcalico/libcalico-go/lib/upgrade/etcd/conversionv1v3"
+	validator "github.com/projectcalico/libcalico-go/lib/validator/v3"
+)
+
+func Convert(args []string) {
+	doc := constants.DatastoreIntro + `Usage:
+  calicoctl convert --filename=<FILENAME>
+                [--output=<OUTPUT>] [--ignore-validation]
+
+Examples:
+  # Convert the contents of policy.yaml to v3 policy.
+  calicoctl convert -f ./policy.yaml -o yaml
+
+  # Convert a policy based on the JSON passed into stdin.
+  cat policy.json | calicoctl convert -f -
+
+Options:
+  -h --help                     Show this screen.
+  -f --filename=<FILENAME>      Filename to use to create the resource. If set to
+                                "-" loads from stdin.
+  -o --output=<OUTPUT FORMAT>   Output format. One of: yaml or json.
+                                [Default: yaml]
+  --ignore-validation           Skip validation on the converted manifest.
+
+
+Description:
+  Convert config files from Calico v1 to v3 API versions. Both YAML and JSON formats are accepted.
+
+  The default output will be printed to stdout in YAML format.
+`
+	parsedArgs, err := docopt.Parse(doc, args, true, "", false, false)
+	if err != nil {
+		fmt.Printf("Invalid option: 'calicoctl %s'. Use flag '--help' to read about a specific subcommand.\n", strings.Join(args, " "))
+		os.Exit(1)
+	}
+	if len(parsedArgs) == 0 {
+		return
+	}
+
+	var rp resourcePrinter
+	output := parsedArgs["--output"].(string)
+	// Only supported output formats are yaml (default) and json.
+	switch output {
+	case "yaml", "yml":
+		rp = resourcePrinterYAML{}
+	case "json":
+		rp = resourcePrinterJSON{}
+	default:
+		fmt.Printf("unrecognized output format '%s'\n", output)
+		os.Exit(1)
+	}
+
+	filename := argutils.ArgStringOrBlank(parsedArgs, "--filename")
+
+	// Load the V1 resource from file and convert to a slice
+	// of resources for easier handling.
+	resV1, err := v1resourceloader.CreateResourcesFromFile(filename)
+	if err != nil {
+		fmt.Printf("Failed to execute command: %v\n", err)
+		os.Exit(1)
+	}
+
+	var results []runtime.Object
+	for _, v1Resource := range resV1 {
+		v3Resource, err := convertResource(v1Resource)
+		if err != nil {
+			fmt.Printf("Failed to execute command: %v\n", err)
+			os.Exit(1)
+		}
+
+		// Remove any extra metadata the object might have.
+		rom := v3Resource.(v1.ObjectMetaAccessor).GetObjectMeta()
+		rom.SetNamespace("")
+		rom.SetUID("")
+		rom.SetResourceVersion("")
+		rom.SetCreationTimestamp(v1.Time{})
+		rom.SetDeletionTimestamp(nil)
+		rom.SetDeletionGracePeriodSeconds(nil)
+		rom.SetClusterName("")
+
+		ignoreValidation := argutils.ArgBoolOrFalse(parsedArgs, "--ignore-validation")
+		if !ignoreValidation {
+			if err := validator.Validate(v3Resource); err != nil {
+				fmt.Printf("Converted manifest resource(s) failed validation: %s\n", err)
+				fmt.Printf("Re-run the command with '--ignore-validation' flag to see the converted output.\n")
+				os.Exit(1)
+			}
+		}
+
+		results = append(results, v3Resource)
+	}
+
+	log.Infof("results: %+v", results)
+
+	err = rp.print(nil, results)
+	if err != nil {
+		fmt.Printf("Failed to execute command: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+// convertResource converts v1 resource into a v3 resource.
+func convertResource(v1resource unversioned.Resource) (conversion.Resource, error) {
+	// Get the type converter for the v1 resource.
+	convRes, err := getTypeConverter(v1resource.GetTypeMetadata().Kind)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert v1 API resource to v1 backend KVPair.
+	kvp, err := convRes.APIV1ToBackendV1(v1resource)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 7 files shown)


calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+Options:
+  -h --help                     Show this screen.
+  -f --filename=<FILENAME>      Filename to use to create the resource.  If set to
+                                "-" loads from stdin.
+  -o --output=<OUTPUT FORMAT>   Output format. One of: yaml or json.
+                                [Default: yaml]
+
+
+Description:
+  Convert config files between different API versions. Both YAML and JSON formats are accepted.
```
robbrockbank: between different implies this is more flexible than it is.  I think it's just that it converts v1 manifests to v3.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+
+	log.Infof("results: %+v", results)
+
+	err = rp.print(nil, results)
+	if err != nil {
+		fmt.Println(err)
+	}
+}
+
+func convertResource(v1resource unversioned.Resource) (runtime.Object, error) {
```
robbrockbank: missing a fn description.  Perhaps should make clear it's converting v1 to v3.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+		return convert(conversion.IPPool{}, v1resource)
+	case "bgppeer":
+		return convert(conversion.BGPPeer{}, v1resource)
+
+	default:
+		return nil, fmt.Errorf("conversion for the resource type '%s' is not supported", v1resource.GetTypeMetadata().Kind)
+	}
+}
+
+func convert(convRes conversion.Converter, v1resource unversioned.Resource) (conversion.Resource, error) {
```
robbrockbank: missing a fn description.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert v1 backend KVPair to v3 API resource.
+	res, err := convRes.BackendV1ToAPIV3(kvp)
+	if err != nil {
+		return nil, err
+	}
+
```
robbrockbank: We should probably validate the final result resource too.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+	switch output {
+	case "yaml", "yml":
+		rp = resourcePrinterYAML{}
+	case "json":
+		rp = resourcePrinterJSON{}
+	default:
+		rp = nil
+	}
+
+	if rp == nil {
```
robbrockbank: nit: can the following two lines just be put in the default: branch above and then we can do away with this if statement.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+			os.Exit(1)
+		}
+		results = append(results, v3Resource)
+	}
+
+	log.Infof("results: %+v", results)
+
+	err = rp.print(nil, results)
+	if err != nil {
+		fmt.Println(err)
```
robbrockbank: should exit with non-zero RC here.

calicoctl/commands/convert.go
robbrockbank: Not sure you gain much from this extra method, other than making it difficult to explain the difference between convertResource() and convert().

I'm not too fussed if this stays as it is, but another way to structure this would be to have the resource-specific switch statement select a converter to use and then just join two methods together, or have a method that returns the converter for a specific resource type (which is easier to explain than the subtle different between convertResource and convert).

Anyways, it's a nit and I'm talking more than I should.

lwr20:
Please add STs :)

tmjd:
I've got a PR against gunjan's branch https://github.com/gunjan5/calico-containers/pull/2 to add some STs for this.

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+func Convert(args []string) {
+	doc := constants.DatastoreIntro + `Usage:
+  calicoctl convert --filename=<FILENAME>
+                [--output=<OUTPUT>]
+
+Examples:
+  # Create a policy using the data in policy.yaml.
+  calicoctl convert -f ./policy.yaml -o yaml
+
+  # Create a policy based on the JSON passed into stdin.
```
bcreane: Probably "Create" -> "Convert"

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+  cat policy.json | calicoctl convert -f -
+
+Options:
+  -h --help                     Show this screen.
+  -f --filename=<FILENAME>      Filename to use to create the resource.  If set to
```
bcreane: nit, single space after period.

calicoctl/commands/convert.go
bcreane: 👍

calicoctl/commands/convert.go
bcreane: 👍

calicoctl/commands/convert.go
bcreane: :

calicoctl/commands/v1resourceloader/v1resourceloader.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Create a new concrete resource structure based on the type.  If the type is
+// a list, this creates a concrete Resource-List of the required type.
+func newResource(tm unversioned.TypeMetadata) (unversioned.Resource, error) {
+	rh, ok := resourceToType[tm]
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("Unknown resource type (%s) and/or version (%s)", tm.Kind, tm.APIVersion))
+	}
+	log.Infof("Found resource helper: %s", rh)
```
bcreane: debug level log?

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,148 @@
...
+                                "-" loads from stdin.
+  -o --output=<OUTPUT FORMAT>   Output format. One of: yaml or json.
+                                [Default: yaml]
+
+
+Description:
+  Convert config files between different API versions. Both YAML and JSON formats are accepted.
+
+  The default output will be printed to stdout in YAML format.
```
bcreane: suggest: "calicoctl sends YAML to standard output (stdout) by default." 

"will be printed" is future tense/passive voice.

calicoctl/commands/convert.go
tmjd: I think it would be a good idea, that would catch some of the problems I've seen in testing when taking converted data and applying it.

gunjan5: hmmm, not sure if it makes sense to fail validation on something that user can't control, I'd say it's better to fail when a user tries to apply the config. Failing validation on the config we generated seems a bit weird (from the UX PoV)

calicoctl/commands/convert.go
gunjan5: hehe, i know i know i complain about this and now i have this in my code too. I guess blind copy paste :em

calicoctl/commands/convert.go
gunjan5: I mostly copied the description from `kubectl convert` command help to be consistent, I can change it if you like.

calicoctl/commands/v1resourceloader/v1resourceloader.go
gunjan5: hmm I've used `Info` to be consistent with the other calicoctl commands, but I think you're right, it/they should be debug

gunjan5: o nvm i thought you were talking about the `log.Infof("results: %+v", results)` line in `convert.go`, but this makes more sense, I'll change it to debug

gunjan5:
@robbrockbank @bcreane @tmjd @lwr20 I've added validation for the converted resources as well as a new `--ignore-validation` flag. It's in the last commit. I could use some help around the wording of the help text and the error message when we fail validation.

robbrockbank:
I assume we'll look into fixing the `creationTimestamp: null` or just live-with?  I think it'll be a pain to get rid of - we'd need to convert the struct to a map[string]interface{} and then remove the field from the map.

gunjan5:
yeah it's hard to get rid of, we also print that with `--export` flag. Would've been easier to add `omitempty` if we didn't import the ObjectMeta field from k8s. I think we've already adjusted our tests to deal with it

calicoctl/commands/convert.go
```diff
@@ -0,0 +1,176 @@
...
+  -h --help                     Show this screen.
+  -f --filename=<FILENAME>      Filename to use to create the resource. If set to
+                                "-" loads from stdin.
+  -o --output=<OUTPUT FORMAT>   Output format. One of: yaml or json.
+                                [Default: yaml]
+  --ignore-validation           Skip validation on the converted manifest.
+
+
+Description:
+  Convert config files from Calico V1 to V3 API versions. Both YAML and JSON formats are accepted.
```
bcreane: nit: is it v[13] or V[13]? Thought the former ...

calicoctl/commands/convert.go
bcreane: Agree with @gunjan5 on this - our bad v3 manifest will be caught when the user applies it. As G5 mentions, there's nothing the user can do about our botched conversion (other than edit and try to apply again).

PR #1788, opened 2017-12-14
Revving libcalico-go and updating tests

heschlie:
## Description

Revving libcalico-go to pickup the FelixConfig changes, and updating the tests to use the new values.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/utils/data.py
+++ b/tests/st/utils/data.py
@@ -610,15 +610,15 @@ felixconfig_name1_rev1 = {
             {'protocol': 'UDP', 'port': 422}, ],
         'interfacePrefix': 'humperdink',
         'ipipMTU': 1521,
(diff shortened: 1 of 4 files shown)


tests/st/utils/data.py
```diff
@@ -648,7 +647,21 @@
     'spec': {
         'ipv6Support': False,
         'logSeverityScreen': 'Debug',
-        'netlinkTimeoutSecs': 11,
+        'netlinkTimeout': '11s',
+    }
+}
+
+felixconfig_name1_rev3 = {
```
robbrockbank: Maybe worth a quick comment above this that the large durations are returned on a get with different units - so it doesn't trip up anyone else using this

PR #1797, opened 2018-01-05
Disallow any updates on old resource versions to fix etcd compaction errors

mgleung:
In cases where a resource version has been compacted in etcd and is then attempted to be updated, it gives the unclear error message `resource: etcdserver: mvcc: required revision has been compacted` instead of notifying the user that they are passing in an old resource version. This fixes the error message and makes it so updates are always done against the newest version of the resource.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -195,16 +195,30 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 	// Check to see if the resourceVersion is specified in the resource object.
 	rv := resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
 
+	// Copy the resource to prevent modifying the input resource metadata.
+	resource = resource.DeepCopyObject().(ResourceObject)
+
 	// If the resourceVersion is specified then we use it to try and update the resource.
 	// Do not attempt to retry if the resource version is specified.
 	if rv != "" {
+		// Clean out the resource version to always get the latest revision.
+		resource.(ResourceObject).GetObjectMeta().SetResourceVersion("")
 		// Validate the metadata is not changed for the the resource.
 		ro, err := rh.get(ctx, client, resource)
 		if err != nil {
 			return ro, err
 		}
-		// Set the specific resource version we are attempting to update.
-		ro.GetObjectMeta().SetResourceVersion(rv)
+		// Check that the resource version is the latest
+		if rv != ro.GetObjectMeta().GetResourceVersion() {
+			id := fmt.Sprintf("%s(%s)", ro.GetObjectKind().GroupVersionKind().GroupKind().Kind, ro.GetObjectMeta().GetName())
+			if ro.GetObjectMeta().GetNamespace() != "" {
+				id = fmt.Sprintf("%s(%s/%s)", ro.GetObjectKind().GroupVersionKind().GroupKind().Kind, ro.GetObjectMeta().GetNamespace(), ro.GetObjectMeta().GetName())
+			}
+			return ro, cerrors.ErrorResourceUpdateConflict{
+				Err:        fmt.Errorf(fmt.Sprintf("Resource version '%s' is out of date (latest: %s). Update the resource YAML/JSON in order to make changes.", rv, ro.GetObjectMeta().GetResourceVersion())),
+				Identifier: id,
+			}
+		}
 		resource = mergeMetadataForUpdate(ro, resource)
 
 		return rh.update(ctx, client, resource)
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -736,6 +736,38 @@ class TestCalicoctlCommands(TestBase):
         rc = calicoctl("delete", data=k8s_np)
         rc.assert_error(text=NOT_SUPPORTED)
         rc.assert_error(text=KUBERNETES_NP)
+
+    @parameterized.expand([
+        ('replace'),
+        ('apply'),
+    ])
+    def test_disallow_update_old_resource_version(self, update_cmd):
+        """
+        Test that we disallow updates on resources with old resource versions.
+        """
+        rc = calicoctl("create", data=ippool_name1_rev1_v4)
+        rc.assert_no_error()
+
+        rc = calicoctl(
+                "get ippool %s -o yaml" % name(ippool_name1_rev1_v4))
+        rc.assert_no_error()
+        rev1 = rc.decoded
+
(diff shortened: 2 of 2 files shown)


calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -198,13 +198,17 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
...
 		ro, err := rh.get(ctx, client, resource)
 		if err != nil {
 			return ro, err
 		}
-		// Set the specific resource version we are attempting to update.
-		ro.GetObjectMeta().SetResourceVersion(rv)
+		// Check that the resource version is the latest
+		if rv != ro.GetObjectMeta().GetResourceVersion() {
+			return ro, fmt.Errorf("Resource version '%s' is out of date (latest: %s). Update the resource in order to make changes.", rv, ro.GetObjectMeta().GetResourceVersion())
+		}
```
robbrockbank: To ensure you get the same error messages, I think you either want to:
-  return a ErrorResourceUpdateConflict type here
or
-  set the resource version in the merged settings and let the update return the error message as appropriate.

robbrockbank: As an aside - I'm not sure what you mean by "Update the resource" in the error - I assume you mean "update the YAML/JSON".  Anyways, if you return the correct error type you don't need to worry about crafting the correct message.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -198,13 +198,17 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 	// If the resourceVersion is specified then we use it to try and update the resource.
 	// Do not attempt to retry if the resource version is specified.
 	if rv != "" {
+		// Clean out the resource version to always get the latest revision.
+		resource.(ResourceObject).GetObjectMeta().SetResourceVersion("")
```
fasaxc: This change means that the method now modified the input data.  I feel pretty strongly that we shouldn't be modifying what's passed in.  It makes for some pretty sneaky bugs down the line and the signature of this method makes it look like it shouldn't touch its input (because it _returns_ a copy of the object with the new values filled in).  

For an example of where the bugs would pop out: a similar issue tripped me up a couple of days ago: my canned unit test data was being modified when I passed it into libcalico-go so the first test that used the canned data worked but later tests weren't sending in the data that they expected.

I think the answer is simply to take a copy of the input struct if you need to modify it.  I think there's a DeepCopy() method on the object that you could use for that.

calicoctl/resourcemgr/resourcemgr.go
mgleung: I was hoping to avoid the error returned by the update since in the case of etcd compaction, it gives an unhelpful error message back about compaction which isn't the real issue the user should fix so I think using an `ErrorResourceUpdateConflict` might be the option I go with. 

Question about `ErrorResourceUpdateConflict`: is a mesage of "update conflict: <resource identifier>" enough to point out to a user? Is there a way that I could use `ErrorResourceUpdateConflict` other than just putting in a small blurb about resource version in the <identifier> portion of the message? It will work according to the code, it just doesn't look like the way the code was meant to be used...

calicoctl/resourcemgr/resourcemgr.go
mgleung: Good point, I overlooked the fact that it will affect our input values.

calicoctl/resourcemgr/resourcemgr.go
robbrockbank: I think it's fine - basically an old resource version means that someone has modified the resource since you took your snapshot.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -198,16 +198,28 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 	// If the resourceVersion is specified then we use it to try and update the resource.
 	// Do not attempt to retry if the resource version is specified.
 	if rv != "" {
+		// Clean out the resource version to always get the latest revision.
+		ru := resource.DeepCopyObject().(ResourceObject)
```
robbrockbank: It looks as though mergeMetadataForUpdate() adjusts the resource that you give it, so I think the branch of code below this if `if rv != ""` conditional will also modify the input resource.

Therefore, probably best just to move this DeepCopy to outside this conditional.  Maybe just have `resource = resource.DeepCopyObject().(ResourceObject)` so it's not possible to accidentally modify the original resource.

mgleung: Good point. Definitely better to make it safer.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -195,11 +195,13 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
 	// Check to see if the resourceVersion is specified in the resource object.
 	rv := resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
 
+	// Copy the resource to prevent modifying the input resource metadata.
+	ru := resource.DeepCopyObject().(ResourceObject)
```
robbrockbank: actually I would have done `resource =` which I think is safer since it means you won't be able to accidentally beat over the original.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -195,35 +195,49 @@ func (rh resourceHelper) Update(ctx context.Context, client client.Interface, re
...
 			return ro, err
 		}
-		// Set the specific resource version we are attempting to update.
-		ro.GetObjectMeta().SetResourceVersion(rv)
-		resource = mergeMetadataForUpdate(ro, resource)
+		// Check that the resource version is the latest
+		if rv != ro.GetObjectMeta().GetResourceVersion() {
+			id := ro.GetObjectMeta().GetName()
+			if ro.GetObjectMeta().GetNamespace() != "" {
+				id = fmt.Sprintf("%s/%s", ro.GetObjectMeta().GetNamespace(), ro.GetObjectMeta().GetName())
```
robbrockbank: sorry just spotted this:  I think this should also include the resource kind in it, so of the format `NetworkPolicy(namespace/name)`

mgleung: Currently the error message ends up looking like this:
`Failed to apply 'IPPool' resource: update conflict: default-ipv4-ippool`
If it already lists the resource type, should we still make the changes so it looks like:
`Failed to apply 'IPPool' resource: update conflict: IPPool(default-ipv4-ippool)`

PR #1799, opened 2018-01-22
Add support for GlobalNetworkSets.

fasaxc:
## Description

Plumb through global networks sets into calicoctl.

## Todos
- [x] Tests
- [x] Documentation: https://github.com/projectcalico/calico/pull/1630
- [x] Release note (will be covered in another component)
- [ ] Switch to main libcalico-go repo after merge there
- [x] Add test/fix large YAML parsing

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/apply.go
+++ b/calicoctl/commands/apply.go
@@ -57,6 +57,7 @@ Description:
     * bgpPeer
     * felixConfiguration
     * globalNetworkPolicy
+    * globalNetworkSet
     * hostEndpoint
     * ipPool
     * networkPolicy
--- a/calicoctl/resourcemgr/globalnetworkset.go
+++ b/calicoctl/resourcemgr/globalnetworkset.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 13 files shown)


calicoctl/resourcemgr/globalnetworkset.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
```
gunjan5: it will be nice if we also have an example GlobalNetworkSet in: https://github.com/projectcalico/calicoctl/tree/master/test-data/v3 since we use it for manual testing and as a reference

calicoctl/commands/apply.go
```diff
@@ -57,6 +57,7 @@ Description:
     * bgpPeer
     * felixConfiguration
     * globalNetworkPolicy
+    * globalNetworkSet
```
caseydavenport: ick, I think these are all wrong :/

All the resource types start uppercase now - `BGPPeer`, `GlobalNetworkPolicy`, etc...

happy to ignore for now and we can fix in another PR.

calicoctl/resourcemgr/globalnetworkset.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func init() {
+	registerResource(
+		api.NewGlobalNetworkSet(),
+		api.NewGlobalNetworkSetList(),
+		false,
+		[]string{"globalnetworkset", "globalnetworksets"},
+		[]string{"NAME"},
+		[]string{"NAME", "NETS"},
```
caseydavenport: Hm, do we have any way to limit the length of what gets put in the NETS column? It could potentially get really really big.

fasaxc: Good point; I've seen kubectl truncate with ellipsis in that sort of situation.  I'll look into it

calicoctl/commands/apply.go
fasaxc: Raised #1800

fasaxc:
@caseydavenport I've added a truncation feature, please can you give that a look?

PR #1807, opened 2018-02-16
Added support for s390x

guirish:
## Description

This PR is raised to provide support for s390x 

## Todos
- [ ] Tests
Have run 
```bash
ARCH=s390x make test
```
All tests are passing.


calicoctl/Dockerfile.calicoctl.build-s390x
```diff
@@ -0,0 +1,8 @@
+FROM s390x/golang:1.8.3
```
caseydavenport: Hm, looking at the Makefile now I can't see that the `Dockerfile.calicoctl.build` series is ever used - looks like it's all using `calico/go-build` now.

So, I think we can remove this file.

fasaxc:
I manually confirmed the @guirish signed our CLA.  Seems the bot is wrong.

PR #1810, opened 2018-02-28
glide.yaml: Remove grpc and x/sys pins

tomdee:
It gets rid of some glide warnings

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -20,7 +20,7 @@ import:
 - package: golang.org/x/net
   version: 66aacef3dd8a676686c7ae3716979581e8b03c47
 - package: golang.org/x/sys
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -46,7 +46,6 @@ import:
   - lib/upgrade
 - package: github.com/termie/go-shutil
 - package: google.golang.org/grpc
-  version: 8050b9cbc271307e5a716a9d782803d09b0d6f2d
```
bcreane: pinning to a particular version seems more robust than floating against `master`. Is there a specific release we could use for `version`?

tomdee: No - the model we have is that we don't pin versions in glide.yaml unless we have a specific need to, they are pinned in glide.lock.

PR #1814, opened 2018-03-01
Bump libcalico go

ozdanborne:
## Description

- Bumps libcalico-go to latest version
- fixes some sample policies that violated new icmp validation

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/utils/data.py
+++ b/tests/st/utils/data.py
@@ -132,7 +132,6 @@ networkpolicy_name1_rev1 = {
         ],
         'ingress': [
             {
-                'notICMP': {'type': 19, 'code': 255},
                 'ipVersion': 4,
                 'action': 'Deny',
                 'destination': {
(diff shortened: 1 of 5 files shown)


tests/st/utils/data.py
```diff
@@ -260,7 +258,6 @@
         ],
         'ingress': [
             {
-                'notICMP': {'type': 19, 'code': 255},
```
caseydavenport: Why are we removing these? Which part of validation does this fail?

matthewdupre: There's a proto: TCP just below.

ozdanborne: These rules have both ICMP rules and Ports rules, which is invalid now:
- If includes port rules, protocol needs to be TCP. 
- If it includes ICMP rules, protocol needs to be ICMP

caseydavenport:
@ozdanborne could you write a release note?

PR #1818, opened 2018-03-08
Add a mount for /var/lib/calico, which is now required.

fasaxc:
## Description

Add /var/lib/calico to the standard list of mounts since the CNI plugin and calico/node now rely on sharing that directory.

## Todos
- [x] Tests -> covered by the calico/node STs
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -299,10 +299,11 @@ Description:
 		containerPath string
 	}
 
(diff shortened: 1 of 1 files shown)


fasaxc:
OK, I've finally had a good run on the calico repo CI: https://semaphoreci.com/calico/calico/branches/pull-request-1738/builds/7

PR #1831, opened 2018-03-30
dont fallback to Update if operation is NotSupported

ozdanborne:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -162,10 +162,9 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 	// Try to create the resource first.
 	ro, err := rh.Create(ctx, client, resource)
 
-	// Fall back to an Update if the resource already exists, or the datastore does not support
-	// create operations for that resource.
+	// Fall back to an Update if the resource already exists.
 	switch err.(type) {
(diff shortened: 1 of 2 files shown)


calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -165,7 +165,7 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 	// Fall back to an Update if the resource already exists, or the datastore does not support
 	// create operations for that resource.
 	switch err.(type) {
-	case cerrors.ErrorResourceAlreadyExists, cerrors.ErrorOperationNotSupported:
+	case cerrors.ErrorResourceAlreadyExists:
```
robbrockbank: Comment above is now incorrect.

robbrockbank: The comment and code seem to be deliberate, so I'm wondering if there are some resources for which Create is not supported, but Apply is - I'm thinking  maybe the global config or cluster info types.  Could you do a quick check in the code?

ozdanborne: I believe it is only k8s network policy and clusterinfo which can return NotSupported, and its consistent across all CRUD operations for them. So no, afaict, there are no resources where Create is not supported but Apply is.

PR #1836, opened 2018-04-05
Handle unsupported applies

ozdanborne:
## Description

~This PR fixes the issue where running an Apply against a resource where Create is not supported returns a "not found error". We now will return "operation not supported".~

~Implemented by changing error logic for `Apply` such that if the `Create` is Unsupported and `Update` is NotFound, return the Create's Unsupported error.~

~I needed to revert my original fix for this (#1831 ) since there are resources that you can Apply but not Create: Node resources in KDD mode.~

Added a special case for the "clusterinfo apply" so that we return `unsupported` error. This fixes an issue where running `apply` against a clusterinfo with name != default returns a `not found` error (since the `create` is unsupported and the `update` needs to run a `get` first).

I'm not happy about adding clusterinfo specific code to the abstract resourcemgr class but I also don't want to refactor this system until we are certain there is more than this scenario.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -28,6 +28,7 @@ import (
 	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
 	yamlsep "github.com/projectcalico/calicoctl/calicoctl/util/yaml"
 	yaml "github.com/projectcalico/go-yaml-wrapper"
+	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	validator "github.com/projectcalico/libcalico-go/lib/validator/v3"
@@ -151,6 +152,15 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
 // Apply is an un-typed method to apply (create or update) a resource. This calls Create
 // and if the resource already exists then we call the Update method.
 func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceObject, error) {
+	// Block operations on ClusterInfo as calico/node is responsible for managing it.
+	if _, ok := resource.(*api.ClusterInformation); ok {
+		return nil, cerrors.ErrorOperationNotSupported{
+			Operation:  "apply",
+			Identifier: "ClusterInformation",
+			Reason:     "resource is readonly",
+		}
+	}
+
 	// Store the original ResourceVersion for the Update operation later.
 	originalRV := resource.(ResourceObject).GetObjectMeta().GetResourceVersion()
 
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -602,7 +602,7 @@ class TestCalicoctlCommands(TestBase):
         # Apply an update to the cluster information and assert not found (we need the node to
         # create it).
         rc = calicoctl("apply", data=clusterinfo_name1_rev2)
-        rc.assert_error(NOT_FOUND)
+        rc.assert_error(NOT_SUPPORTED)
 
         # Delete the resource by name (i.e. without using a resource version) - assert not
         # supported.


calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -162,15 +162,26 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
...
-		// Try updating if the resource already exists.
+		// Fall back to an Update if the resource already exists.
 		return rh.Update(ctx, client, resource)
+
+	case cerrors.ErrorOperationNotSupported:
+		// Insert the original ResourceVersion back into the object before trying the Update.
+		resource.(ResourceObject).GetObjectMeta().SetResourceVersion(originalRV)
+
+		// Try updating if the resource already exists.
+		_, nerr := rh.Update(ctx, client, resource)
```
robbrockbank: `_, nerr -> ro, nerr`

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -162,15 +162,26 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
...
+		if _, dne := nerr.(cerrors.ErrorResourceDoesNotExist); dne {
```
robbrockbank: it's more common to use `ok` to indicate successful casting (rather than using `dne`

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -162,15 +162,26 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
...
+			// If the resource doesn't exist, we should return the original "not supported"
+			// error. Otherwise, the user will get "not found" for a failed apply-create.
+			return ro, err
+		}
+		return ro, nerr
```
caseydavenport: This should return the updated object rather than the original one.

calicoctl/resourcemgr/resourcemgr.go
caseydavenport: I feel like this is wrong.

Take the kdd Node case as an example:
- Create node fails with unsupported
- Update node fails with "node does not exist"

The error the use should receive is "Node does not exist", not "operation unsupported"

ozdanborne: I think the UX you've outlined contradicts the one outlined here:

>If calicoctl were to do an "apply" action on a ClusterInformation resource, it will get the standard error that the resource is "not found" (since it will try an update). We should make sure that this error message gets switched to give out a "not supported" error messaged instead to make it clearer that you should not be updating a ClusterInformation resource.

caseydavenport: I don't see how.

For ClusterInformation, any operation should return "Not supported"

For nodes, "Create" should return not supported, but "Apply" and "Update" should not return "Not supported"

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -151,6 +152,15 @@ func (rh resourceHelper) GetObjectType() reflect.Type {
 // Apply is an un-typed method to apply (create or update) a resource. This calls Create
 // and if the resource already exists then we call the Update method.
 func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, resource ResourceObject) (ResourceObject, error) {
+	// Block operations on ClusterInfo as calico/node is responsible for managing it.
+	if _, ok := resource.(*api.ClusterInformation); ok {
+		return nil, cerrors.ErrorOperationNotSupported{
+			Operation:  "apply",
+			Identifier: resource,
+			Reason:     "clusterinfo is managed by calico/node",
```
caseydavenport: I think the reason should be that "resource is read-only"

caseydavenport:
@ozdanborne this LGTM - what does the output from calicoctl look like?

I'd expect something like this: 

```
Failed to create 'ClusterInformation' resource: operation create or apply is not supported on ClusterInformation
```

ozdanborne:
Yep:

`Failed to apply 'ClusterInformation' resource: operation apply is not supported on ClusterInformation: resource is readonly`

PR #1851, opened 2018-05-04
Fix dist/calicoctl for building on host arch

mkumatag:
This PR is for fixing dist/calicoctl target for building this project on host architecture. Previously is was building only `adm64` even we run this target on any machines like `ppc64le,arm64 etc...`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,8 +118,8 @@ binary: $(CALICOCTL_FILES) vendor
 
 
 dist/calicoctl: $(CALICOCTL_FILES) vendor
-	$(MAKE) dist/calicoctl-linux-amd64
-	cp dist/calicoctl-linux-amd64 dist/calicoctl
+	$(MAKE) dist/calicoctl-linux-$(ARCH)
+	cp dist/calicoctl-linux-$(ARCH) dist/calicoctl
 
 ## build for linux on amd64
 dist/calicoctl-linux-amd64: $(CALICOCTL_FILES) vendor


deitch:
@fasaxc and I discussed this when doing #1838. The decision as I recall it (correct me please as needed, @fasaxc) was:

1. `make build` would build the binary for the current os/arch as `dist/calicoctl-$(OS)-$(ARCH)`
2. `make dist/calicoctl-$(OS)-$(ARCH)` would build the binary for the given os/arch as `dist/calicoctl-$(OS)-$(ARCH)`
3. `make dist/calicoctl` would build the binary `dist/calicoctl` for `linux/amd64`. 

Essentially, `dist/calicoctl` is a "legacy" target. One should always be building an explicit os/arch binary.

mkumatag:
> @fasaxc and I discussed this when doing #1838. The decision as I recall it (correct me please as needed, @fasaxc) was:
> 
> make build would build the binary for the current os/arch as dist/calicoctl-$(OS)-$(ARCH)
> make dist/calicoctl-$(OS)-$(ARCH) would build the binary for the given os/arch as dist/calicoctl-$(OS)-$(ARCH)
> make dist/calicoctl would build the binary dist/calicoctl for linux/amd64.

I don't see a harm in using arch there, having said that I'm okay using `make build` target instead :)
> Essentially, dist/calicoctl is a "legacy" target. One should always be building an explicit os/arch binary.

deitch:
> I don't see a harm in using arch there, having said that I'm okay using make build target instead :)

Yeah, definitely is a push to standardize the make targets to `build` `image` `push` etc.

mkumatag:
In favor of above standard, I'm closing this PR. Thanks @deitch

deitch:
> Thanks @deitch

Always a pleasure working together @mkumatag

PR #1867, opened 2018-06-05
Update Makefile and fix static checks

tomdee:


Dockerfile.amd64
```diff
@@ -1,7 +1,7 @@
 FROM amd64/alpine:3.7
 MAINTAINER Tom Denham <tom@projectcalico.org>
 
-ADD dist/calicoctl-linux-amd64 /calicoctl
+ADD bin/calicoctl-linux-amd64 /calicoctl
```
deitch: Thank you! I wanted to do this when I did the large cross-build and standardization PR, but thought it would raise too many hackles and mix too much together.

Makefile
```diff
@@ -89,139 +102,104 @@ vendor: glide.yaml
...
+      -e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+      -v $(CURDIR)/.go-pkg-cache:/go-cache/:rw \
+      -e GOCACHE=/go-cache \
+	    $(GO_BUILD_CONTAINER) sh -c '\
+          cd /go/src/$(PACKAGE_NAME) && \
+          go build -v -o bin/calicoctl-$(OS)-$(ARCH) $(LDFLAGS) "./calicoctl/calicoctl.go"'
+
+# Overrides for the binaries that need different output names
+bin/calicoctl: bin/calicoctl-linux-amd64
+	mv $< $@
```
deitch: I think this should be 

```
cp $< $@
```

We want to keep around `bin/calicoctl-linux-amd64` but _add_ the archless/OSless default of `bin/calicoctl`

PR #1871, opened 2018-06-12
descriptive error for profile apply failure on kdd

dmmcquay:
## Description

Add a descriptive error for attempting to apply a `profile` with kdd. Error message used to be:

```bash
Failed to apply 'Profile' resource: Revision  invalid
```

This PR makes it more clear:
```bash
Failed to apply 'Profile' resource: operation Create is not supported on Profile(profile1) 
```

fixes https://github.com/projectcalico/calicoctl/issues/1827

## Todos
- [x] Tests

## Release Note

```release-note
None required
```
Signed-off-by: derek mcquay <derek@tigera.io>

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -172,10 +172,9 @@ func (rh resourceHelper) Apply(ctx context.Context, client client.Interface, res
 	// Try to create the resource first.
 	ro, err := rh.Create(ctx, client, resource)
 
-	// Fall back to an Update if the resource already exists, or the datastore does not support
-	// create operations for that resource.
+	// Fall back to an Update if the resource already exists.
 	switch err.(type) {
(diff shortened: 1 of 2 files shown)


ozdanborne:
There may be room to clean up this error message, but last time I touched code in this area it took me 3 tries to get right:

- https://github.com/projectcalico/calicoctl/pull/1831
- https://github.com/projectcalico/calicoctl/pull/1834
- https://github.com/projectcalico/calicoctl/pull/1836

The original bug as mentioned in 1836 is that when running an Apply against a resource where Create is not supported returned a "not found error", which was a misleading UX. This is because `apply` does a `create` and if that fails, an `update`. The create error is gobbled and the update error is returned.

The fix I implemented was a bit hacky as it put resource specific code in the resourcemanager. But it was the best option at the time without a huge refactor.

caseydavenport:
I suggest we put this PR on hold for now.

We may improve Profile apply by supporting the creation of new Profiles in KDD in the near future.

tomdee:
@caseydavenport Shall we just close this then?

PR #1886, opened 2018-07-20
Makefile: Fully containerize update-libcalico

tomdee:
In order to make it MacOS compatible


tomdee:
I like your confidence!

PR #1902, opened 2018-09-09
Replace docopt with spf13/cobra

briansan:
## Description
This is a pretty nasty looking PR so I apologize, but I promise it isn't as bad as the github diff viewer is making it look; none of the actual functionality was changed, just the command-line interface parts. 
The main bulk of the change comes from the `args.go` file that defines the arguments for each command and the flags that they receive.
It seems like the default behavior for most if not all flags was to return blank (for strings) or false (for bools) which the cobra replacement provides pretty nicely.
If it passes the ci tests, that means it's good to go, right? 😉 

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note
```release-note
Replaced docopt with spf13/cobra for command-line argument parsing
```

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -16,84 +16,51 @@ package main
 
 import (
 	"fmt"
-
 	"os"
 
-	"github.com/docopt/docopt-go"
-	"github.com/projectcalico/calicoctl/calicoctl/commands"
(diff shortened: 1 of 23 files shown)


caseydavenport:
Given the size of this PR, I think we'll have to close this for now. It's not obvious to me that this is worthwhile, just yet.

PR #1904, opened 2018-09-14
Enable junit reports and write to report directory

lwr20:
## Description
Output test result reports for UTs and STs to ./report

## Todos
None

## Release Note
```release-note
None required
```


caseydavenport:
@tomdee @lwr20 can this be merged now?

PR #1917, opened 2018-10-22
Remove build date from binary

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
It's something that Tom and I had been discussing for a while as a step towards reproducible builds. As it is, you can't build twice and confirm it's the same binary.

> if a user rebuilds the binary rather than using our published version

I've never seen this before - is this something we're actually worried will happen? If they do, it should have a different commit ID / version / or report a version with a `-dirty` on the end, so it should be easy to spot.

In theory if they are changing the binary they could just hard-code the version text to whatever they want anyway!

PR #1924, opened 2018-11-06
Update the URL of alpine glibc signing key

doublek:
## Description

Update our Dockerfile per [the note at sgerrand/alpine-pkg-glibc](https://github.com/sgerrand/alpine-pkg-glibc/tree/fc20c32a048060a5478aa579b648bace7364b38f#please-note) they've updated the URL for the signing key.

## Release Note

```release-note
None required
```


doublek:
Combining this with #1923 so that CI can actually run.

PR #1935, opened 2018-11-13
Support calicoctl --version

neiljerram:
We already support 'calicoctl version', which prints information like
this:

    Client Version:    v3.4.0-0.dev-15-g553bb080-dirty
    Git commit:        553bb080
    no etcd endpoints specified

However we don't support the standard 'calicoctl --version', and it
would be better if we did.  (Even our own Makefiles expect this; see
https://github.com/projectcalico/node/blob/master/Makefile#L459)

With this change, 'calicoctl --version' will print information likes
this:

    calicoctl version v3.4.0-0.dev-15-g553bb080-dirty, build 553bb080

We could refactor more to make this output the same as 'calicoctl
version', but I think this is good enough as is and that the further
work isn't clearly justified.

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -53,7 +53,7 @@ Description:
 
   See 'calicoctl <command> --help' to read about a specific subcommand.
 `
-	arguments, _ := docopt.Parse(doc, nil, true, "", true, false)
+	arguments, _ := docopt.Parse(doc, nil, true, commands.VERSION_SUMMARY, true, false)
 
 	if logLevel := arguments["--log-level"]; logLevel != nil {
 		parsedLogLevel, err := log.ParseLevel(logLevel.(string))


caseydavenport:
@fasaxc @neiljerram We explicitly removed this a ways back: https://github.com/projectcalico/calicoctl/pull/1808

Do we need both options here?

neiljerram:
Sorry, I wasn't aware of that. But I have personally hit a road hump several times when typing 'calicoctl --version' to find out version information, and as I pointed out we even have a Makefile that expects that to work. And, when I'm just trying to see the version, it makes no difference what the layout/format is.

So in my view #1808 had the wrong balance of concerns, and we're better off with this support in.

If folks are really worried about layout consistency, I can do a bit more work to fix that.

caseydavenport:
@neiljerram It's a bit more than just layout consistency - there is also information that is missing from the `--version` variant - namely the "Cluster version".

If the Makefiles are using `--version`, then IMO that's a makefile bug rather than a calicoctl one.

I think if we do this, we need to at least have it perform the same logic as `calicoctl version`, but I'm still personally in favor of supporting only `calicoctl version` (this is also consistent with `kubectl` behavior, which is something we've tried to maintain in general).

fasaxc:
Sounds like everyone can live with`--version` if it prints the same output as `version` so let's do that if Neil has the bandwidth to implement?

@caseydavenport I've found this a gotcha in the past too.  Seems a shame to throw away a perfectly good convention that's been around for decades just because kubectl doesn't do it.

neiljerram:
Unfortunately we can't make `--version` identical to `version` because `--version` has to be a string that is precomputed before `docopt.Parse` is called, whereas `version` includes information about the cluster that is obtained dynamically at runtime and depends on other args or environment, notably `--config`.

We could make `--version` identical to the non-cluster-derived part of `version`, i.e. make it like this:
```
Client Version:   v3.4.0-0.dev-18-g338531f8
Git commit:       338531f8
```
but that doesn't feel like a big win to me because it still wouldn't be the same as the whole `version` output, and in any case I'm happy accepting that `--version` and `version` do different things.

- `--version` just tells you the version of the binary you're running (per widespread Linux convention)

- `version` tells you info about the calicoctl binary and the Calico cluster that you're pointing at.

I think it _is_ worth drawing out that difference a bit more with this addition to `--version`, as in #1937:
```
calicoctl v3.4.0-0.dev-18-g338531f8, build 338531f8
Run 'calicoctl version' to see cluster version information as well.
```

OK, with all that said, I'm happy to do whatever the consensus thinks.  My personal order of preference is:

1. #1937 

2. Any formatting improvement / alignment that the consensus wants, bearing in mind the static vs dynamic difference noted above.

3. Current master (i.e. #1935)

4. Revert to printing an unhelpful error message for `calicoctl --version`.

WDYT?

caseydavenport:
I'd be happy to print something like "Please use the version command instead" and just not print any version information at all when typing `calicoctl --version`.

But, I don't want multiple commands which provide different types of version output. I see only negatives with doing so.

neiljerram:
@caseydavenport Doesn't this count as a positive: someone (e.g. me) typed the standard command for getting the version of a program, and got a useful answer ?

caseydavenport:
Yes, I suppose you're right and I shouldn't have said I only see negatives :D 

It does count as some bit of positive, but I guess my feeling is that it doesn't outweigh the negative of "I typed a command and got an output that isn't the full set of information"

caseydavenport:
> I've found this a gotcha in the past too. Seems a shame to throw away a perfectly good convention that's been around for decades just because kubectl doesn't do it.

@fasaxc FWIW, it's not _just_ kubectl that does it - another example is the `go` tool which uses `go version` as well. It's not like it's a crazy departure from other tools.

caseydavenport:
> typed the standard command for getting the version of a program, and got a useful answer ?

I think the best path forward, given technical limitations, would be this:

- `calicoctl version` prints version information as it does today.
- `calicoctl --version` prints a helpful usage text explaining that `calicoctl version` is the correct command.

This will avoid the main issues we're debating, which are:
- Two version commands with different output is confusing.
- The current `--version` command provides an unhelpful and confusing error.

PR #1937, opened 2018-11-14
Improve 'calicoctl --version' output

neiljerram:
Output is like this:
```
calicoctl v3.4.0-0.dev-18-g338531f8, build 338531f8
Run 'calicoctl version' to see cluster version information as well.
```


caseydavenport:
Can we drop the version output altogether and simply refer people to the "one true version command", `calicoctl version`?

caseydavenport:
@neiljerram agreed, thanks.

Issue #1910, opened 2018-09-28
calicoctl --unknown-flag wrongly returns 0


mirwan:
When calling `calicoctl` with an unknown flag, the return code is 0. For example, when wrongly placing a command flag before the command, the command is not executed and usage is displayed (partially as it seems) but the return code is 0.
NB: This does not occur when passing unknown flag to the command.

## Expected Behavior
The return code should not be 0 when an unknown flag is passed to calicoctl and the command fails (is not executed).

## Current Behavior
```console
mirwan$ docker run -it --rm quay.io/calico/ctl:v3.1.3 --unknown-flag
Usage:
  calicoctl [options] <command> [<args>...]
mirwan$ echo $?
0
mirwan$ echo '{}' | docker run -i --rm quay.io/calico/ctl:v3.1.3 --skip-exists create -f -
Usage:
  calicoctl [options] <command> [<args>...]
mirwan$ echo $?
0
mirwan$ echo '{}' | docker run -i --rm quay.io/calico/ctl:v3.1.3 create --skip-exists -f -
Failed to execute command: error parsing document: cannot parse object '{}' into value of type []unstructured.Unstructured
mirwan$ echo $?
1
mirwan$ echo '{}' | docker run -i --rm quay.io/calico/ctl:v3.1.3 create --unknown-flag -f -
Usage:
  calicoctl create --filename=<FILENAME> [--skip-exists] [--config=<CONFIG>] [--namespace=<NS>]
Invalid option: 'calicoctl create --unknown-flag -f -'. Use flag '--help' to read about a specific subcommand.
mirwan$ echo $?
1
```
As it not related to docker:
```console
mirwan$ docker run -it --rm --entrypoint=/bin/sh quay.io/calico/ctl:v3.1.3
~ # /calicoctl --unknown-flag
Usage:
  calicoctl [options] <command> [<args>...]
~ # echo $?
0
~ #
```
## Steps to Reproduce (for bugs)
1. docker run -it --rm quay.io/calico/ctl:v3.1.3 --unknown-flag
2. echo $?

## Context
When using calicoctl in a ansible playbook, the command/shell actually fails but is reported as OK which is a problem

## Your Environment

* calicoctl version (e.g. `calicoctl version`):
```
Client Version:    v3.1.3
Build date:        2018-05-30T17:15:59+0000
Git commit:        231083c2
no etcd endpoints specified
```
* calico/node version (e.g. `docker run --rm calico/node versions`): N/A
[...]

PR #1977, opened 2019-03-04
Catch and exit on bad command-line flags

jpalardy:
## Description

Addresses #1910. This fix looks for the returning `err` and exits if present.

If the `err` is a `UserError`, we remind the user of the `--help` option.

I tested positive and negative tests and added some checks to the Python scripts.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calicoctl will now return a non-zero exit code when passed an unknown flag
```

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -55,7 +55,6 @@ Description:
   See 'calicoctl <command> --help' to read about a specific subcommand.
 `
 	arguments, err := docopt.Parse(doc, nil, true, commands.VERSION_SUMMARY, true, false)
-
 	if err != nil {
 		if _, ok := err.(*docopt.UserError); ok {
 			// the user gave us bad input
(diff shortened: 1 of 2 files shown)


calicoctl/calicoctl.go
```diff
@@ -54,7 +54,15 @@ Description:
 
   See 'calicoctl <command> --help' to read about a specific subcommand.
 `
-	arguments, _ := docopt.Parse(doc, nil, true, commands.VERSION_SUMMARY, true, false)
+	arguments, err := docopt.Parse(doc, nil, true, commands.VERSION_SUMMARY, true, false)
+
```
caseydavenport: just a minor formatting thing, but generally would remove this newline between the call which generates the error, and the handling of the error.

e.g.,

```go
retVal, err := someFunc()
if err !=nil {
  // Handle it
  ...
}
```

jpalardy: Ah, ok.

Let me patch that up tomorrow morning and send another commit.

jpalardy: I added that whitespace commit, let me know.

caseydavenport: Looks nice

PR #1985, opened 2019-03-08
Add NetworkSets support

lmm:
## Description
Add support for NetworkSets.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add NetworkSets
```

### DIFF
--- a/calicoctl/resourcemgr/networkset.go
+++ b/calicoctl/resourcemgr/networkset.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


lmm:
Tests to come

calicoctl/resourcemgr/networkset.go
```diff
@@ -0,0 +1,59 @@
...
+	client "github.com/projectcalico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/libcalico-go/lib/options"
+)
+
+func init() {
+	registerResource(
+		api.NewNetworkSet(),
+		api.NewNetworkSetList(),
+		true,
+		[]string{"networkset", "networksets", "ns"},
```
caseydavenport: Hrm, the abbreviation "ns" seems nice but is also what Kubernetes uses for "Namespace" so might be confusing... especially if we decide to add a Namespace object some day.

Maybe `netset`? WDYT?

caseydavenport:
This looks like it's probably right to me- once we get the libcalico PR merged we can bump the version in here to refer to master and hopefully it all just works!

calicoctl/resourcemgr/networkset.go
lmm: Ahh that's a good point. `netset` is nice. And for GlobalNetworkSets, `gnetset`?

PR #2004, opened 2019-05-06
Fix regression in reporting errors in ipam show

tomastigera:
Do not return error code in ipam show if ip does not exist

## Description

```
DATASTORE_TYPE=kubernetes KUBECONFIG=~/.kube/config bin/calicoctl-linux-amd64 ipam show --ip=1.1.1.1
echo $?
1
```

is back to 

```
DATASTORE_TYPE=kubernetes KUBECONFIG=~/.kube/config bin/calicoctl-linux-amd64 ipam show --ip=1.1.1.1
echo $?
0
```

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/ipam/show.go
+++ b/calicoctl/commands/ipam/show.go
@@ -69,7 +69,8 @@ Description:
 	// `IP 192.168.71.1 is not assigned in block`. This is not exactly an error,
 	// so not returning it to the caller.
 	if err != nil {
-		return err
+		fmt.Println(err)
+		return nil
 	}
 
 	// IP address is assigned with attributes.


calicoctl/commands/ipam/show.go
```diff
@@ -69,7 +69,8 @@ Description:
 	// `IP 192.168.71.1 is not assigned in block`. This is not exactly an error,
 	// so not returning it to the caller.
 	if err != nil {
-		return err
+		fmt.Println(err)
```
jpalardy: use stderr?

tomastigera: I stick with pure revert atm

caseydavenport: Yeah, this isn't an error case so I think stdout is probably right,

PR #1088, opened 2016-08-03
added link to OS X kubelet binary

cainelli:
I think is good to reference OS X binary as well.

https://coreos.com/kubernetes/docs/latest/configure-kubectl.html#download-the-kubectl-executable

PR #2006, opened 2019-05-14
Setting defaults before validation for resources

asincu:
## Description

Validation of resources needs to be done after setting the proper default values. For example, for VXLAN mode this should be `Never` rather than an empty string.

When applying an ip pool policy with IPV6 CIDR and no VXLAN mode selected via `calicoctl apply -f`

```
---
kind: IPPool
apiVersion: projectcalico.org/v3
metadata:
  name: ippool-2
spec:
  cidr: 2002::/64
  ipipMode: Never
  natOutgoing: true
Calicoctl crashes with (VXLANMode other than 'Never' is not supported on an IPv6 IP pool)
```

Related PR: https://github.com/projectcalico/libcalico-go/pull/1084

## Todos
- [ ] Tests

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -400,6 +400,7 @@ func unmarshalResource(tm unstructured.Unstructured, b []byte) ([]runtime.Object
 		return nil, err
 	}
 
+	// Setting default values to resource before un marshalling
 	defaults.SetDefaults(unpacked)
 	if err = yaml.UnmarshalStrict(b, unpacked); err != nil {
 		return nil, err
@@ -429,10 +430,12 @@ func unmarshalSliceOfResources(tml []unstructured.Unstructured, b []byte) ([]run
 		if err != nil {
 			return nil, err
 		}
+
 		unpacked[i] = r
+		// Setting default values to resource before un marshalling
+		defaults.SetDefaults(unpacked[i])
 	}
 
-	defaults.SetDefaults(unpacked)
 	if err := yaml.UnmarshalStrict(b, &unpacked); err != nil {
 		return nil, err
 	}
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	defaults "github.com/mcuadros/go-defaults"
 	"github.com/projectcalico/calicoctl/calicoctl/commands/argutils"
 	yamlsep "github.com/projectcalico/calicoctl/calicoctl/util/yaml"
 	yaml "github.com/projectcalico/go-yaml-wrapper"
@@ -403,6 +404,7 @@ func unmarshalResource(tm unstructured.Unstructured, b []byte) ([]runtime.Object
 		return nil, err
 	}
 
+	defaults.SetDefaults(unpacked)
 	log.Infof("Type of unpacked data: %v", reflect.TypeOf(unpacked))
 	if err = validator.Validate(unpacked); err != nil {
 		return nil, err
@@ -434,6 +436,7 @@ func unmarshalSliceOfResources(tml []unstructured.Unstructured, b []byte) ([]run
 		return nil, err
 	}
 
+	defaults.SetDefaults(unpacked)
 	// Validate the data in the structures.  The validator does not handle slices, so
 	// validate each resource separately.
 	for _, r := range unpacked {
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017,2015-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tomastigera:
Doesnt this mean that the unset value now has meaning "Never" ?So if someones forget, instead of getting an error, an unintended value is set?

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -383,6 +384,7 @@ func unmarshalResource(tm unstructured.Unstructured, b []byte) ([]runtime.Object
 		return nil, err
 	}
 
+	defaults.SetDefaults(unpacked)
```
tomastigera: isn't this overriding the actually set values?

asincu: From what I can tell it only overrides empty values.

```
	func (f *Filler) SetDefaultValues(fields []*FieldData) {
	for _, field := range fields {
		if f.isEmpty(field) {
			f.SetDefaultValue(field)
		}
	}
```

jpalardy: What if the value was _explicitly_ set to the "empty value"? And the empty value is valid, but not the default…

I don't know if we have cases like that. :thinking:

caseydavenport: I'm not sure why we need this, we should be hitting this code in libcalico IIUC: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/ippool.go#L58

tomastigera: Even if you do not have that usecase now, pulling in a lib that has such corner cases? If so, it should be used broadly. What if you have an int that by default should be 25 but 0 is a valid value? The annotation is cool, if only it was used by the yaml unmarshaller.

If used this way, should it go to before line 383? First create an uninitialized object, set default then unmarshall into this partially initialized struct?

asincu: > I'm not sure why we need this, we should be hitting this code in libcalico IIUC: https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/ippool.go#L58

@caseydavenport From my digging, it seems that calicoctl invoked apply via
https://github.com/projectcalico/calicoctl/blob/master/calicoctl/commands/resources.go#L129. `CreateResourcesFromFile` reads the file and unmarshals the content at https://github.com/projectcalico/calicoctl/blob/2140631e726895836474beb1e5ffcbbf6a0eb737/calicoctl/resourcemgr/resourcemgr.go#L457-L497.  Then the validation fails at https://github.com/projectcalico/calicoctl/blob/master/calicoctl/resourcemgr/resourcemgr.go#L407.

tomastigera: If validation fails, because something is not set properly, that is correct, right? Or at least you can see it that way ...

asincu: @tomastigera  Looking over the docs for ippool at https://github.com/projectcalico/calico/blob/60e262b977e5338242bb9b3c6670626364f79f5e/v3.7/reference/calicoctl/resources/ippool.md the default value should be `Never`. 

I applied your suggestions about setting the defaults before unmarshalling. Please take a look.

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -430,6 +432,7 @@ func unmarshalSliceOfResources(tml []unstructured.Unstructured, b []byte) ([]run
 		unpacked[i] = r
 	}
 
+	defaults.SetDefaults(unpacked)
```
tomastigera: I believe you need to do it per object before line 432

tomastigera:
I think I start liking this change :-) :+1:

calicoctl/resourcemgr/resourcemgr.go
```diff
@@ -399,6 +400,7 @@ func unmarshalResource(tm unstructured.Unstructured, b []byte) ([]runtime.Object
 		return nil, err
 	}
 
+	defaults.SetDefaults(unpacked)
```
caseydavenport: I'm a little bit hesitant to introduce a new library here - to be honest, I feel like this must already be working for the equivalent ipipMode field which has the same semantics. I haven't had time to dig into it yet, but feel like this should behave the same as the `ipipMode` field does.

glide.yaml
```diff
@@ -15,7 +15,8 @@ import:
   - json
 - package: github.com/projectcalico/go-yaml-wrapper
 - package: github.com/projectcalico/libcalico-go
-  version: 97f275dffda5fcbad25248aec09b556173e09e92
+  repo: https://github.com/asincu/libcalico-go.git
```
briansan: we probably don't want to merge this change into master :)

asincu: @briansan  I have a PR opened on libcalico-go that needs to be merged before this one: https://github.com/projectcalico/libcalico-go/pull/1084

calicoctl/resourcemgr/resourcemgr_test.go
```diff
@@ -0,0 +1,216 @@
...
+	const CidrV4 = "192.168.0.0/16"
+	const PoolName = "my-ippool"
+	const AnotherPoolName = "another-ippool"
+
+	const VxlanModeNever = string(api.VXLANModeNever)
+	const VxlanModeAlways = string(api.VXLANModeAlways)
+	const IpipModeNever = string(api.IPIPModeNever)
+	const IpipModeAlways = string(api.IPIPModeAlways)
+
+	ipPoolV6WithNeverVxlan := ipPoolSpec(DefaultIpPoolTemplate, CidrV6, PoolName, VxlanModeNever, IpipModeNever)
```
tomastigera: Are all these test supposed to test the generic feature of setting defaults and use ippool to prove that it works, or should the tests check all sort of combination of ipv4/6+vxlan+ipipmode ? Given the nature of the change, I would expect the former, but it looks like the latter.

calicoctl/resourcemgr/resourcemgr.go
asincu: @caseydavenport I added a test for missing IpipMode field and I am seeing this error: `error with field IPpool.IPIPMode = '' (IPIPMode other than 'Never' is not supported on an IPv6 IP pool)`

asincu:
Closing this as we decided to go with projectcalico/calicoctl#2015 and #1088

PR #2010, opened 2019-05-18
add glide mirror

shigy123456:
## Description

/kind documentation

When we build a calicoctl project, we need to add the mirror information of glide before install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,13 +78,35 @@ To list all possible targets, run `make help`.
 
 2. [Install Glide](https://github.com/Masterminds/glide#install).
 
-3. Populate the `vendor/` directory in the project's root with this project's dependencies:
+3. Add mirror:
+
+   ```
+   glide mirror set https://k8s.io/client-go https://github.com/kubernetes/client-go --vcs git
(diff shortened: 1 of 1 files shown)


caseydavenport:
@shigy123456 I don't think this is correct - what is the purpose of this change?

shigy123456:
@caseydavenport 
When building calicoctl, you need to refer to packages such as golang. org/x. The code of these packages can not be downloaded in China, and you need to add a mirror of github.
Even in the United States, adding this mirror is no problem.
In order to ensure that calicoctl can be built in all countries, I recommend adding mirror information in the build steps.

caseydavenport:
>  The code of these packages can not be downloaded in China, and you need to add a mirror of github.

Ok, that makes sense. However, I don't think this should be the mainline experience. We could maybe have a FAQ section at the bottom which is "If you're trying to do this in China, then..."

However, we're planning to move to `go mod` in the near future, and this won't apply there, so I'm going to close this for now. Thanks!

timyininspur:
@caseydavenport
Okay, that's all right. Thank you for your attention.

PR #2011, opened 2019-05-23
IPAM utilization reporting

neiljerram:
## Todos
- [x] Tests
- [x] Documentation - at https://github.com/projectcalico/calico/pull/2636
- [x] Release note - at https://github.com/projectcalico/libcalico-go/pull/1087

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -398,7 +398,7 @@ foss-checks: vendor
 ###############################################################################
 .PHONY: ci
 ## Run what CI runs
-ci: clean build-all static-checks ut st image-all
+ci: clean build-all static-checks test image-all
 
 ###############################################################################
 # CD
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ default: build
 all: build
 
 ## Run the tests for the current platform/architecture
-test: ut st
+test: ut fv st
 
 ###############################################################################
 # Both native and cross architecture builds are supported.
@@ -318,6 +318,20 @@ ut: bin/calicoctl-linux-amd64
 		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
 		$(LOCAL_BUILD_MOUNTS) \
 		$(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && ginkgo -cover -r --skipPackage vendor calicoctl/* $(GINKGO_ARGS)'
+
+###############################################################################
+# FVs
+###############################################################################
+.PHONY: fv
+## Run the tests in a container. Useful for CI, Mac dev.
+fv: bin/calicoctl-linux-amd64 run-etcd-host
+	docker run --rm -v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
+		--net=host -e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+		$(LOCAL_BUILD_MOUNTS) \
+	        -v $(CURDIR)/.go-pkg-cache:/go-cache/:rw \
+                -e GOCACHE=/go-cache \
+		$(CALICO_BUILD) sh -c 'cd /go/src/$(PACKAGE_NAME) && go test ./tests/fv'
+	$(MAKE) stop-etcd
 
 ###############################################################################
 # STs
--- a/Makefile
+++ b/Makefile
@@ -324,7 +324,8 @@ ut: bin/calicoctl-linux-amd64
 ###############################################################################
 .PHONY: fv
 ## Run the tests in a container. Useful for CI, Mac dev.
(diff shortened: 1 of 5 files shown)


calicoctl/commands/test_ipam.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/docopt/docopt-go"
+
+	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/test_ipam"
+)
+
+func TestIPAM(args []string) error {
```
caseydavenport: This is neat and seems useful to have.

however, I worry about adding it to calicoctl itself since users might not really understand it.


I think at a minimum we should have it be under some clearly named subcommand - e.g., `calicoctl unsupported test-ipam` or something like that.

Even better might just be to have a second tool that this repo builds that is only used for test purposes and not actually supplied to users.

And even better, I think my dream would be that instead of building a tool at all, we simply write the tests that require this in golang so they can just import the logic they need. WDYT?

caseydavenport:
I'm really excited about this!

A few quick thoughts on the output format

```
IP Pool 10.65.0.0/16 'ipam-test-v4'
    Block 10.65.79.0/26    59/64 IPs still available (92%)
IP Pool fd5f:a1b3:64::/48 'ipam-test-v6'
    Block fd5f:a1b3:64:4f2c:ec1b:27b9:1989:77c0/122    57/64 IPs still available (89%)
IP Pool 0.0.0.0/0 'orphaned allocation blocks'
    (no blocks)
```

I wonder if we should use a table format, similar to other calicoctl commands, with headings, columns, rows, etc. e.g,

```
POOL          CIDR                    UTILIZATION
my-pool       192.168.0.0/24          200/256 (78%)
```

Note the above also switches to counting the number of used IPs rather than the number of free IPs. Not sure which is better... it felt more intuitive to me that 100% would mean "full" rather than "empty". 

This is different from the current format of `calicoctl ipam show --ip`, but I think is probably easier to read and more consistent with other commands. 

We could then do a `-o yaml` or something to show the per-block stuff (though note that per-block isn't required for this first drop).

e.g.,

```
calicoctl ipam show my-pool -o yaml

- pool: my-pool
  cidr: 192.168.0.0/16
  blocks:
  - cidr: 192.168.0.0/26
    node: somenode
    usedIPs: 53
```

fasaxc:
Maybe `s/UTILIZATION/USED/`, that's a common idiom used by commands like `df` and `free`

```
              total        used        free      shared  buff/cache   available
Mem:       32568508     4424304    21417956      443144     6726248    27232104
Swap:       1003516           0     1003516
```

neiljerram:
@fasaxc @caseydavenport On the point about whether to report (primarily) the number in use, or the number available, I would personally be happy either way, but was swayed by the language in our customer request (RFE-137), which specifically asks for "how many IPs we still have available in a pool." Hence I broke the tie on that side. WDYT?

caseydavenport:
I like the idea of reporting both "USED" and "FREE", so the user can take their pick :)

neiljerram:
Note: needs glide update after https://github.com/projectcalico/libcalico-go/pull/1087 has merged.

neiljerram:
Sample output for `calicoctl ipam show`:
```
+----------+-------------------+------------+---------------+
| GROUPING |       CIDR        | IPS IN USE | IPS AVAILABLE |
+----------+-------------------+------------+---------------+
| IP Pool  | 10.65.0.0/16      | 5/64 (7%)  | 59/64 (92%)   |
| IP Pool  | fd5f:abcd:64::/48 | 7/64 (10%) | 57/64 (89%)   |
+----------+-------------------+------------+---------------+
```
and for `calicoctl ipam show --show-blocks`:
```
+----------+-------------------------------------------+------------+---------------+
| GROUPING |                   CIDR                    | IPS IN USE | IPS AVAILABLE |
+----------+-------------------------------------------+------------+---------------+
| IP Pool  | 10.65.0.0/16                              | 5/64 (7%)  | 59/64 (92%)   |
| Block    | 10.65.79.0/26                             | 5/64 (7%)  | 59/64 (92%)   |
| IP Pool  | fd5f:abcd:64::/48                         | 7/64 (10%) | 57/64 (89%)   |
| Block    | fd5f:abcd:64:4f2c:ec1b:27b9:1989:77c0/122 | 7/64 (10%) | 57/64 (89%)   |
+----------+-------------------------------------------+------------+---------------+
```

caseydavenport:
Docs completed here, updating label: https://github.com/projectcalico/calico/pull/2636#pullrequestreview-243285554

PR #2012, opened 2019-05-23
Fix "Unknown" ASN being shown. 

jadech32:
## Description

Fixed unknown value by replacing hardcoded value with default ASN (64512).

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Show default ASN when no BGP configuration exists
```

### DIFF
--- a/calicoctl/commands/printer.go
+++ b/calicoctl/commands/printer.go
@@ -230,7 +230,8 @@ func config(client client.Interface) func(string) string {
 		case "asnumber":
 			if asValue == "" {
 				if bgpConfig, err := client.BGPConfigurations().Get(context.Background(), "default", options.GetOptions{}); err != nil {
-					// Use the default ASNumber of 64512 when there is none configured.
+					// Use the default ASNumber of 64512 when there is none configured (first ASN reserved for private use).
+					// https://en.m.wikipedia.org/wiki/Autonomous_system_(Internet)#ASN_Table
 					asValue = "64512"
 				} else {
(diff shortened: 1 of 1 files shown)


calicoctl/commands/printer.go
```diff
@@ -230,7 +230,7 @@ func config(client client.Interface) func(string) string {
 		case "asnumber":
 			if asValue == "" {
 				if bgpConfig, err := client.BGPConfigurations().Get(context.Background(), "default", options.GetOptions{}); err != nil {
-					asValue = "unknown"
+					asValue = "64512"
```
caseydavenport: Worth adding a comment here explaining where this value comes from, I think.

calicoctl/commands/printer.go
```diff
@@ -230,7 +230,9 @@ func config(client client.Interface) func(string) string {
 		case "asnumber":
 			if asValue == "" {
 				if bgpConfig, err := client.BGPConfigurations().Get(context.Background(), "default", options.GetOptions{}); err != nil {
-					asValue = "unknown"
```
tmjd: Shouldn't we check if the error was due to the non-existence of this or an actual error accessing the data?  If the error was a failure to connect or something like that then I think the value should be unknown.

caseydavenport: Hrm, yeah I think you're right.

@jadech32 could you adjust to check the error type, and only show the default value if it's an `ErrorResourceDoesNotExist` (or whatever its called)?

For any other errors, we should continue showing "unknown"

PR #2015, opened 2019-05-28
Remove unmarshalling when creating a resource

asincu:
## Description
Remove validation when unmarshalling a resource. The validation will be handled by lib calico, when calling create or update. Libcalico will also set the defaults for missing fields in the process of update/create.

Validation of resources needs to be done after setting the proper default values. For example, for VXLAN mode this should be Never rather than an empty string.

For example, when applying an ip pool policy with IPV6 CIDR and no VXLAN mode selected via calicoctl apply -f
```
---
kind: IPPool
apiVersion: projectcalico.org/v3
metadata:
  name: ippool-2
spec:
  cidr: 2002::/64
  ipipMode: Never
  natOutgoing: true

```
Calicoctl crashes with (VXLANMode other than 'Never' is not supported on an IPv6 IP pool)

## Todos
- [x] Update lib-calico version
- [x] Release note

## Release Note

```release-note
Set resource defaults before performing validation in calicoctl. Note: when creating multiple resources from a single file, calicoctl will now successfully create all valid resources up to the first invalid one before exiting with an error.  
```

### DIFF
--- a/calicoctl/resourcemgr/resourcemgr.go
+++ b/calicoctl/resourcemgr/resourcemgr.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Tigera, Inc. All rights reserved.
+// Copyright (c) 2017-2019 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


caseydavenport:
@asincu would you mind cherry-picking this fix to the release-v3.7 branch? (as well as the libcalico PR)

PR #2020, opened 2019-06-05
Fix crash in ipam when NewClient fails

tomastigera:
Does not panic

Adds assert_output_not_contains in test infra

fixes OS-4514

## Description

Adds some testing about broken store configuration. However, those tests take ~10s to timeout and fail.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/tests/st/calicoctl/test_crud.py
+++ b/tests/st/calicoctl/test_crud.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import logging
 import copy
+import os
 
 from nose_parameterized import parameterized
 
@@ -38,6 +39,7 @@ class TestCalicoctlCommands(TestBase):
         """
         Test that a basic CRUD flow for pool commands works.
         """
+
         # Create the ipv6 pool using calicoctl, and read it out using an
         # exact get and a list query.
         rc = calicoctl("create", data=ippool_name2_rev1_v6)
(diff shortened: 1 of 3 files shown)


tests/st/calicoctl/test_crud.py
```diff
@@ -76,6 +78,19 @@ def test_get(self):
...
         rc.assert_error(text=NOT_FOUND)
 
+    def test_broke_store_config(self):
+        """
+        Test that broken store configuration does not crash
+        """
+
+        rc = calicoctl("get policy", broken_store=True)
+        rc.assert_error()
+        rc.assert_output_not_contains("panic")
```
neiljerram: Do we need to have these `rc.assert_output_not_contains("panic")` calls separately from `rc.assert_error()`?  I think it would make sense for `rc.assert_error()` *always* to check that the output does not include a panic.  WDYT?

tests/st/utils/utils.py
```diff
@@ -168,8 +168,20 @@ def assert_output_contains(self, text):
...
+        """
+        Assert the calicoctl command output does not contain the supplied text.
+        Args:
+            text:   Expected text in the command output.
+        """
+        if not text:
+            return
+        assert not text in self.output, "Unxpected text in output; \n" + \
+            "command=" + self.command + "\noutput=\n" + self.output + \
+            "\nunexpected=\n" + text
```
neiljerram: As above, I suggest adding this check in `assert_error`, instead of creating and calling this new method.

tests/st/utils/utils.py
```diff
@@ -209,6 +221,10 @@ def calicoctl(command, data=None, load_as_stdin=False, format="yaml", only_stdou
         etcd_auth = "%s:2379" % ETCD_HOSTNAME_SSL
     else:
         etcd_auth = "%s:2379" % get_ip()
+
+    if broken_store :
+        etcd_auth = "broken.xyz:54321"
```
neiljerram: The scenario that discovered this problem was with no etcd config at all, which might be a bit different from having a broken config - and which might also solve your problem with the failure taking a long time.  Could you adjust so that the new testing is with no etcd config at all, instead of a broken config?

tests/st/utils/utils.py
tomastigera: we can do the check for no panic in the assert_error for sure. If there was a panic, the assertion would fail.

tests/st/utils/utils.py
tomastigera: will look into that, thanks

tests/st/utils/utils.py
tomastigera: done

tests/st/utils/utils.py
tomastigera: yup, done, added a whole bunch of test for other commands too

tests/st/calicoctl/test_crud.py
tomastigera: indeed, done

PR #2021, opened 2019-06-10
Replaces blank map[] output with blank column instead (-o wide)

jadech32:
## Description

When printing (e.g. -o wide) and the selected metric / data doesn't exist, print "" (blank) instead of default of `map[]`.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ default: build
 all: build
 
 ## Run the tests for the current platform/architecture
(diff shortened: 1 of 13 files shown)


jadech32:
Closing as changes need to be made (commit clean up)

PR #2022, opened 2019-06-10
Output empty string instead of map[] if attribute doesn't exist

jadech32:
## Description

When printing (e.g. -o wide) and the selected metric / data doesn't exist, print "" (blank) instead of default of `map[]`.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


jadech32:
@fasaxc , should there be a unit test for this? I'm not sure where I would put it in `calicoctl`.

fasaxc:
Yes, there should be UTs.  Please create a new `printer_test.go`.  I think, by convention, we use Ginkgo in this repo.

Unfortunately, I see some problems with the fix.  Fortunately, they're all good  learning opportunities :-)

- If you pass it an interface with a `nil` type, it will panic (I tried it in the go playground to be sure!):
```
fmt.Println(joinAndTruncate(nil, ",", 0))
---
panic: reflect: call of reflect.Value.IsNil on zero Value

goroutine 1 [running]:
reflect.Value.IsNil(...)
```
   This is a bit subtle; interfaces actually have two values inside; one is a pointer to a type, the other is a pointer to a value.  So, you can have an interface that has a nil value but a valid type `interface{}(map[string]string(nil))`, or, you can have an interface with nil type and nil value`interface{}(nil)`.  They're not the same!  In practice, it might not be possible to hit this case in calicoctl, but it'd be a nasty gotcha to leave in a formatting function that might be used in dozens of places.

- It's very specific; it only handles a non-nil `interface{}` containing a `nil` value.  After seeing the bad behaviour for that value, it's worth taking a step back and asking if other values might be handled poorly.  In this case, it looks like this function was never designed to handle maps at all(!); it was designed for slices.  What other types of map could we throw in?

  - If we try an empty map `map[string]string{}`, what does it output?  I think we'll get `"map[]"` just like before since it'll bypass your reflect check but the built-in stringifier for maps treats nil and empty maps equally.

  - If we try a map with some keys `map[string]string{"a": "b"}`, I think we get this: `"map[a:b]"`, which isn't very user-friendly (how many of our users know/care what a map is).

I think the right fix is to add proper handling for maps to this function.  Since we have to use reflection, I think that'll mean checking to see if you have something of `Kind() == reflect.Map` and then using `reflect.MapRange` to iterate over the map.  Then, format each KV pair as, say, `key=value` and append it to a slice.  _Then_ fall through to the existing logic to stringify the slice.  I think there may be one more issue to solve if we go down that route: map iteration is random, you'll probably want to sort the slice before falling through or the output will be non-deterministic (which looks odd in a CLI command makes it harder to test).

UTs that we should have here:

- nil interface
- nil map
- empty map
- map with one kv
- map with several kvs
- map with enough kvs that they get truncated
- nil slice
- empty slice
- slice with one value
- slice with multiple values
- slice that gets truncated
- tests of truncation to make sure that we get the expected number of chars; e.g. if we truncate "123456" to 5 chars, do we get "12345..." "12..." or something else?  If we truncate "12345" to 5 chars, do we still get "12345".

Using a Ginkgo DescribeTable to make a table-driven test would make it very easy to add all these.

calicoctl/commands/printer.go
```diff
@@ -198,12 +199,24 @@ func join(items interface{}, separator string) string {
...
-	// If the interface is empty, return an empty string. Otherwise, a raw map would
-	// show in the output.
-	if reflect.ValueOf(items).IsNil() {
+	// Nil types.
+	if items == nil {
 		return ""
 	}
 
+	// If it is a map, just convert key,value pairs into slice.
+	mapSlice := []string{}
```
jpalardy: this can move down into the `if` scope

calicoctl/commands/printer.go
```diff
@@ -198,12 +199,24 @@ func join(items interface{}, separator string) string {
...
+	if reflect.TypeOf(items).Kind() == reflect.Map {
+		iterator := reflect.ValueOf(items).MapRange()
+		for iterator.Next() {
+			k := iterator.Key().Interface().(string)
+			v := iterator.Value().Interface().(string)
+			mapSlice = append(mapSlice, k+"="+v)
+		}
+		sort.Strings(mapSlice)
+		items = mapSlice
```
jpalardy: if we'll end up re-assigning to `items`, maybe we don't need `mapSlice` at all?  (append to empty items)

jadech32: We'd need to initialize `items` to `[]string{}` anyway as it is coming in as `interface{}` which results in the same lines.

PR #2030, opened 2019-07-05
Remove references to libnetwork, which is no longer part of calico/node

rene-dekker:
Remove references to libnetwork, which is no longer part of v3.
Related to this PR for the /node project is https://github.com/projectcalico/node/pull/266

- Removed references to CALICO_LIBNETWORK in /node/run.go as Calico Libnetwork is no longer part of the node project.
- Tests have been run using `make test`
- No new tests have been added, nor was new code.

## Todos
- Related, but not required: https://github.com/projectcalico/node/pull/266

### DIFF
--- a/calicoctl/commands/node/run.go
+++ b/calicoctl/commands/node/run.go
@@ -247,7 +247,6 @@ Description:
 	envs := map[string]string{
 		"NODENAME":                  name,
 		"CALICO_NETWORKING_BACKEND": backend,
-		"CALICO_LIBNETWORK_ENABLED": fmt.Sprint(!disableDockerNw),
 	}
 
 	// Validate the ifprefix to only allow alphanumeric characters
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/run.go
```diff
@@ -259,16 +258,6 @@ Description:
 		return fmt.Errorf("Error executing command: invalid to disable Docker Networking and enable Container labels")
 	}
 
-	// Set CALICO_LIBNETWORK_IFPREFIX env variable if Docker network is enabled and set to non-default value.
-	if !disableDockerNw && ifprefix != DEFAULT_DOCKER_IFPREFIX {
```
lmm: Can we remove the `disableDockerNw` var too?

rene-dekker: Thank you for your suggestion. I double-checked it with Casey and it can indeed be removed. I did this in a new commit.

PR #2036, opened 2019-08-06
[master] Automatic update of libcalico-go to bf00054ed881fc4023e6e748c5faba3ad0757028

fasaxc:


fasaxc:
Looks like the automated approach won't work here, needs a fix.

PR #2037, opened 2019-08-12
Update to latest libcalico-go

neiljerram:


calicoctl/commands/ipam/show.go
```diff
@@ -73,11 +73,12 @@ Description:
 		ip := argutils.ValidateIP(passedIP.(string))
 		attr, err := ipamClient.GetAssignmentAttributes(ctx, ip)
 
-		// IP address is not assigned, this prints message like `IP 192.168.71.1
-		// is not assigned in block`. This is not exactly an error, so not
-		// returning it to the caller.
 		if err != nil {
-			fmt.Println(err)
+			// IP address is not assigned.  The detailed error message here is either
```
caseydavenport: We should probably be checking for a "Does not exist" error, specifically here. Right?

otherwise won't things like connection errors show up as "Is not assigned"?

neiljerram: Good point; please see #2038

Issue #2047, opened 2019-08-27
Add calicoctl subcommand patch


mattymo:
Issue type: Feature request

As calicoctl evolves to work more and more like kubectl, adding the subcommand patch will improve end user experience. With kubectl patch, a user can quickly modify an object and not have to parse it. Without a patch function, the user needs to write some relatively sophisticated yaml/json parser and then replace the object.

Here is an example that is both unreadable and unpleasant to maintain:
https://github.com/kubernetes-sigs/kubespray/pull/5114/files#diff-68a99474f0a01df786a37dd82338e083R15

What would be much easier could be this:
```
calicoctl patch node mynode01 -p '{"spec":{"bgp": {"routeReflectorClusterID": "CLUSTER_ID"}}}' 
```

mattymo:
@syscll 
Strategy: I'm not too particular, so long as the scenario I described initially is possible.
JSON and YAML: of course. Feature parity with kubectl patch should be the goal
I agree that calicoctl should reject the entire patch if it partially fails.

syscll:
SGTM - I am actively working on this so I'll post a link to the PR when I have a base implementation ready.

PR #2054, opened 2019-08-30
Implement calicoctl patch

syscll:
## Description

This is a base implementation of a new `calicoctl patch` command as requested in #2047 

The end goal is to best match the implementation of `kubectl patch`. However, to start, the functionality of this PR will be limited to JSON data only and will perform a strategic merge patch as described [here](https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-strategic-merge-patch-to-update-a-deployment).

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
calicoctl now supports patching resources using `calicoctl patch`
```


calicoctl/commands/patch.go
```diff
@@ -0,0 +1,111 @@
...
+  # Partially update a node using a json merge patch.
+  calicoctl patch node node-0 --patch '{"spec":{"unschedulable":true}}' --type json
+
+Options:
+  -h --help                  Show this screen.
+  -p --patch=<PATCH>         Spec to use to patch the resource.  If set
+                             to "-" loads from stdin.
+  -t --type=<TYPE>           Format of patch type:
+                                json        JSON Patch, RFC 6902 (not yet implemented)
+								merge       JSON Merge Patch, RFC 7386 (not yet implemented)
```
caseydavenport: Indentation here maybe looks a bit funny

syscll: Fixed

syscll:
Will add some unit and integration tests, as well as perform some manual testing.

calicoctl/commands/patch.go
```diff
@@ -0,0 +1,111 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/docopt/docopt-go"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
+	log "github.com/sirupsen/logrus"
+)
+
+func Patch(args []string) error {
+	// TODO: write coherent description.
```
caseydavenport: Is this TODO still valid?

syscll: Ah, yes, I suppose it was more of a reminder to get it proofread by someone else. So if you're happy, I'll remove the TODO

caseydavenport: Yeah, this looks fine to me!

syscll:
What's the procedure around release notes and docs? Do I just need to add a `patch` section to the `calicoctl` reference for v3.10?

caseydavenport:
@syscll yep - I've added a release note to the description which will get picked up during the next release.

We should add new documentation to the `master/` directory of the Calico repo (since that will become v3.10). We'll definitely want the reference changes, we might also want to update some of our other docs which have users do a get->modify->update to just use patch instead, since it's way simpler.

caseydavenport:
Let's make sure to squash into one commit before merging.

PR #2062, opened 2019-09-27
Upgrade to go v1.13

chenrui333:
## Description

Upgrade to go v1.13.

Relates to https://github.com/projectcalico/go-build/pull/82, #2061 

Currently, the brew formula is already built with go v1.13, https://github.com/Homebrew/homebrew-core/pull/44132

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
@chenrui333 looks like you've got some test failures.

chenrui333:
Circle back on this. :)

chenrui333:
Should be covered with the latest `go-build` container.

chenrui333:
@caseydavenport it should be good now! Thanks!

.golangci.yml
```diff
@@ -0,0 +1,4 @@
+issues:
+  exclude:
```
caseydavenport: @chenrui333 could you add a comment explaining why we need to exclude these checks? It's not obvious to me, and I don't want the context to get lost :)

chenrui333: Sure, I am doing that now.

chenrui333: Done. :)

chenrui333: Let me know if you need more things from my side.

PR #2066, opened 2019-10-02
need to harden go mod downloads against network errors

rafaelvanoni:


Makefile
```diff
@@ -306,18 +313,16 @@ sub-tag-images-%:
...
-	$(DOCKER_RUN) $(CALICO_BUILD) golangci-lint run --deadline 5m
+	$(DOCKER_RUN) $(CALICO_BUILD) golangci-lint run $(LINT_ARGS)
 
 .PHONY: fix
-## Fix static checks
 fix:
 	goimports -w calicoctl/*
 
 .PHONY: install-git-hooks
-## Install Git hooks
```
caseydavenport: FYI these `##` commands are so that the `make help` magic picks up these targets.

PR #2082, opened 2019-11-12
Clean up a few things from go.sum

tmjd:
This PR originally was updating go-yaml but the overnight CI updated the libcalico-go pin and did most of the work. This PR now cleans up go.sum. Mainly to make sure we don't have any references to the v1 version of go-yaml, I also removed a bunch of old libcalico-go entries too.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


go.mod
```diff
@@ -84,8 +84,7 @@ require (
 	gopkg.in/inf.v0 v0.9.1
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7
 	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
-	gopkg.in/yaml.v1 v1.0.0-20140924161607-9f9df34309c0 // indirect
-	gopkg.in/yaml.v2 v2.2.2
+	gopkg.in/yaml.v2 v2.2.5
```
bcreane: Hey @tmjd - is an update to `yaml.v2` required for the fix, or was this just a byproduct of updating the `go-yaml-wrapper` pin?

tmjd: The fix was to update yaml.v2 to at least v2.2.4. I figured it would be good to stop using v1.0.0 also and found that mergo was the only thing that needed updating to remove the v1.0.0 inclusion.
Is there a hesitation around these updates?

PR #2083, opened 2019-11-15
node diags: add ip(6)tables fallback

joshti:
## Description
ip(6)tables-save was/is an optional requirement for some distributions - add a fallback mode generated from ip(6)tables.

Full disclosure: I didn't know Go until a couple hours ago so this could be... something. Totally happy to take modifications / feedback.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fall back to ip(6)tables if ip(6)tables-save is not present when gathering diags
```

### DIFF
--- a/calicoctl/commands/node/diags.go
+++ b/calicoctl/commands/node/diags.go
@@ -24,6 +24,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"bytes"
 
 	"github.com/docopt/docopt-go"
 	log "github.com/sirupsen/logrus"
@@ -222,6 +223,13 @@ func writeDiags(cmds diagCmd, dir string) {
 	content, err := exec.Command(parts[0], parts[1:]...).CombinedOutput()
 	if err != nil {
 		fmt.Printf("Failed to run command: %s\nError: %s\n", cmds.cmd, string(content))
+		if strings.Compare(cmds.cmd, "iptables-save -c") == 0 {
+			fmt.Printf("Fallback: iptables\n")
+			content = walkIPTables("/proc/net/ip_tables_names", "iptables")
+		} else if strings.Compare(cmds.cmd, "ip6tables-save -c") == 0 {
+			fmt.Printf("Fallback: ip6tables\n")
+			content = walkIPTables("/proc/net/ip6_tables_names", "ip6tables")
+		}
 	}
 
 	// This is for the commands we want to run but don't want to save the output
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/diags.go
```diff
@@ -222,6 +223,13 @@ func writeDiags(cmds diagCmd, dir string) {
 	content, err := exec.Command(parts[0], parts[1:]...).CombinedOutput()
 	if err != nil {
 		fmt.Printf("Failed to run command: %s\nError: %s\n", cmds.cmd, string(content))
+		if strings.Compare(cmds.cmd, "iptables-save -c") == 0 {
```
caseydavenport: If you want to do exact equality, I think it's more common in go to just do `==`. 

e.g., `cmds.cmd == "iptables-save -c"`

calicoctl/commands/node/diags.go
```diff
@@ -235,3 +243,39 @@ func writeDiags(cmds diagCmd, dir string) {
 		log.Errorf("Error writing diags to file: %s\n", err)
 	}
 }
+
+func walkIPTables(file string, command string)([]byte) {
```
caseydavenport: Looks like you want to run `goimports` across this file to fix up some minor formatting / spacing issues.

calicoctl/commands/node/diags.go
```diff
@@ -235,3 +243,39 @@ func writeDiags(cmds diagCmd, dir string) {
...
+		fmt.Printf("Failed to open: %s\n", file)
+	}
+
+	command, err = exec.LookPath(command)
+	if err != nil {
+		fmt.Printf("Failed to run command: %s\nError: Not found in PATH\n", command)
+	}
+
+	scanner := bufio.NewScanner(tablesFile)
+	scanner.Split(bufio.ScanLines)
```
caseydavenport: Is this just setting the split function to be the default? If so, do we need this line at all?

calicoctl/commands/node/diags.go
```diff
@@ -235,3 +243,39 @@ func writeDiags(cmds diagCmd, dir string) {
 		log.Errorf("Error writing diags to file: %s\n", err)
 	}
 }
+
```
caseydavenport: Would appreciate a quick function comment to explain the purpose of `waslkIPTables` for future readers.

calicoctl/commands/node/diags.go
```diff
@@ -235,3 +243,39 @@ func writeDiags(cmds diagCmd, dir string) {
...
+func walkIPTables(file string, command string)([]byte) {
+
+	tablesFile, err := os.Open(file)
+	if err != nil {
+		fmt.Printf("Failed to open: %s\n", file)
```
caseydavenport: I think we need to return if we hit one of these errors, otherwise we'll try to continue below.

calicoctl/commands/node/diags.go
```diff
@@ -235,3 +243,39 @@ func writeDiags(cmds diagCmd, dir string) {
...
+		err = cmd.Run()
+
+		if err != nil {
+			fmt.Printf("ERROR in %s: %s", table, errbuf.String())
+		} else {
+			persist += fmt.Sprintf("*%s\n%sCOMMIT\n", table, rules.String())
+		}
+	}
+
+	tablesFile.Close()
```
caseydavenport: It's more common in go to use `defer` for closing files, especially when you might have several exit paths to a function.

You can just do `defer tablesFile.Close()` right after opening the file to make sure it will always get closed.

joshti:
Closing in favour of https://github.com/projectcalico/calicoctl/pull/2096 which fixes this as a byproduct.

PR #2087, opened 2019-11-28
Move image build/push targets back to local Makefile

rafaelvanoni:


rafaelvanoni:
> Issue in CI, so presumably the reversion isn't quite right yet.

Fixed !

PR #2088, opened 2019-12-04
move all k8s.io packages to kubernetes v1.16.3

rafaelvanoni:


tests/st/utils/data.py
```diff
@@ -261,14 +261,6 @@
             'name': 'endpoint1',
             'uid': 'test-uid-change',
         }],
-        'initializers': {
```
tmjd: Can you provide some reasoning why this removal is needed? I think removing it was for fixing a couple tests but I'd like to know why that removal is ok.

rafaelvanoni: That field was deprecated in 1.13 and removed in 1.16.0

PR #2098, opened 2020-01-03
Add checksystem test

caseydavenport:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```


tests/st/calicoctl/test_checksystem.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2015-2017 Tigera, Inc. All rights reserved.
```
lmm: nit: year isn't current

PR #2119, opened 2020-04-22
ipam configure strictaffinity

gianlucam76:
calicoctl ipam configure --strictaffinity=<true|false>

adds capability to configure IPAM StrictAffinity.

When set to true, if IPAM on a node does not have any IP address left in
its blocks and cannot allocate a new block, no IP will be borrowed from
matching blocks assigned to other nodes. This means, in such a scenario
POD will remain in ContainerCreating state with no IP.

calicoctl ipam show --show-borrowed
provides a detailed list of which PODs on which nodes have received an
IP address borrowed from a block assigned to different node.

## Todos
- [ ] Documentation

## Release Note

```release-note
None required
```


calicoctl/commands/ipam/show.go
```diff
@@ -138,3 +136,56 @@ Description:
...
+	passedIP := parsedArgs["--ip"]
+	showBlocks := parsedArgs["--show-blocks"].(bool)
+	showBorrewed := parsedArgs["--show-borrowed"].(bool)
+
+	if passedIP != nil {
+		return showIP(ctx, ipamClient, passedIP)
+	} else if showBlocks {
+		return showBlockUtilization(ctx, ipamClient, true)
+	} else if showBorrewed {
+		return showBorrewedDetails(ctx, ipamClient)
```
caseydavenport, line 187: typo: "borrewed" (and a few other places as well)

calicoctl/commands/ipam/show.go
```diff
@@ -21,83 +21,135 @@ import (
...
-	// Create a new backend client from env vars.
-	cf := parsedArgs["--config"].(string)
-	client, err := clientmgr.NewClient(cf)
+func showBorrowedDetails(ctx context.Context, ippoolClient clientv3.IPPoolInterface, bc bapi.Client) error {
+	details, err := getPoolDetails(ctx, ippoolClient, bc)
 	if err != nil {
 		return err
 	}
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader([]string{"POOL", "BLOCK", "NODE OWNING BLOCK", "POD", "NAMESPACE", "NODE BORROWING IP"})
```
caseydavenport, line 96: I think we need to include the IP address itself in this output. And actually, I'm not sure if we need to show the pool at all?

I was imagining something like this:

```
IP                   BORROWING-NODE         BLOCK               BLOCK OWNER      TYPE    ALLOCATED-TO
192.168.0.1          my-node-1              192.168.0.0/26      my-node-2        pod     kube-system/coredns-pod-2
192.168.0.2          my-node-4              192.168.0.0/26      my-node-2        vxlan   my-node-4
192.168.0.3          my-node-8              192.168.0.0/26      my-node-2        N/A     N/A
 
```

This shows the three types of IP allocations we might have:
- Pod allocation
- tunnel address allocation (in this case VXLAN, but it could also be IPIP)
- unknown - i.e., doesn't have attributes attached to it

calicoctl/commands/ipam/show.go
```diff
@@ -21,83 +21,135 @@ import (
...
-	parsedArgs, err := docopt.Parse(doc, args, true, "", false, false)
+	allPools, err := ippoolClient.List(ctx, options.ListOptions{})
 	if err != nil {
-		return fmt.Errorf("Invalid option: 'calicoctl %s'. Use flag '--help' to read about a specific subcommand.", strings.Join(args, " "))
+		return nil, err
 	}
-	if len(parsedArgs) == 0 {
-		return nil
+
+	for _, pool := range allPools.Items {
```
caseydavenport, line 50: Could we simplify our data structures here?

Instead of a nested structure, with pools at the top level and then another list below, what about something like this?

```go
type BorrowedIP struct {
  addr string
  borrowingNode string
  block string
  blockOwner string
  allocationType string
  allocatedTo string
  pool string
}
```

And then we just use a `[]BorrowedIP` to populate the table? This is all assuming we even need IP pool in the table output. Is there a requirement for that?

calicoctl/commands/ipam.go
```diff
@@ -29,9 +29,10 @@ func IPAM(args []string) error {
 	doc := constants.DatastoreIntro + `Usage:
   calicoctl ipam <command> [<args>...]
 
-    release      Release a Calico assigned IP address.
-    show         Show details of a Calico assigned IP address,
-                 or of overall IP usage.
+    release          Release a Calico assigned IP address.
+    show             Show details of a Calico assigned IP address,
+                     or of overall IP usage.
+    strictaffinity   true/false
```
caseydavenport, line 35: I think we should have a general `configure` subcommand here, in case we want to expose other configuration options here in the future. 

e.g. `calicoctl ipam configure --strictaffinity` rather than `calicoctl ipam strictaffinity` 

WDYT?

calicoctl/commands/ipam/strictaffinity.go
```diff
@@ -0,0 +1,95 @@
...
+     --true             Set StrictAffinity to true. Borrowing IP addresses from other
+                        blocks is not allowed. 
+     --false            Set StrictAffinity to false. Borrowing IP addresses from other
+                        blocks when needed is allowed.
+  -c --config=<CONFIG>  Path to the file containing connection configuration in
+                        YAML or JSON format.
+                        [default: ` + constants.DefaultConfigPath + `]
+
+Description:
+  When a new POD is scheduled on a node, IPAM assigns an IP address to that POD
```
caseydavenport, line 65: POD - > "Pod"

calicoctl/commands/ipam/show.go
gianlucam76: That's much better. Thank you!

calicoctl/commands/ipam.go
gianlucam76: Makes sense. It is better if in future as you mentioned we want to make something else configurable.

calicoctl/commands/ipam/show.go
gianlucam76: There is no requirement for ippool so used your suggested structure removing pool.

calicoctl/commands/ipam/configure.go
```diff
@@ -0,0 +1,98 @@
...
+Options:
+  -h --help                        Show this screen.
+     --strictaffinity=<true/false> Set StrictAffinity to true/false. When StrictAffinity
+                                   is true, borrowing IP addresses is not allowed.
+  -c --config=<CONFIG>             Path to the file containing connection configuration in
+                                   YAML or JSON format.
+                                   [default: ` + constants.DefaultConfigPath + `]
+
+Description:
+  When a new Pod is scheduled on a node, IPAM assigns an IP address to that Pod
```
caseydavenport, line 64: I think this docstring probably doesn't make sense any more, for a few reasons:

- IPAM is used for more than just pods
- The configure subcommand is more generic than just IP borrowing (in theory). 

It could probably be as simple as something like "Modify and view configuration for Calico IP address management" 

Which makes me think, do we have a way for the user to view what the current state is?

Maybe we need a `calicoctl ipam show configuration` subcommand?

calicoctl/commands/ipam/show.go
```diff
@@ -24,80 +24,136 @@ import (
...
+				// Allocations[i] if not nil:
+				// - b.OrdinalToIP(i) is the allocated IP address
+				// - *b.Allocations[i] is the index of the corresponding b.Attributes
 
-	// Create a new backend client from env vars.
-	cf := parsedArgs["--config"].(string)
-	client, err := clientmgr.NewClient(cf)
+				attributes := b.Attributes[*b.Allocations[i]]
+
+				if b.Affinity != nil {
```
caseydavenport, line 67: Do we want to show IP addresses that are allocated in blocks that have no affinity? That is potentially also considered a form of borrowing.

If we _do_ want to show that, then we'll need to restructure this a little bit.

calicoctl/commands/ipam/show.go
```diff
@@ -24,80 +24,136 @@ import (
...
+}
 
-Options:
-  -h --help             Show this screen.
-     --ip=<IP>          Report whether this specific IP address is in use.
-     --show-blocks      Show detailed information for IP blocks as well as pools.
-  -c --config=<CONFIG>  Path to the file containing connection configuration in
-                        YAML or JSON format.
-                        [default: ` + constants.DefaultConfigPath + `]
+func getPoolDetails(ctx context.Context, ippoolClient clientv3.IPPoolInterface, bc bapi.Client) ([]*borrowedIP, error) {
```
caseydavenport, line 48: Should probably rename this function now that it has nothing to do with pools :)

maybe `getBorrowedIPs`?

calicoctl/commands/ipam/show.go
```diff
@@ -24,80 +24,136 @@ import (
...
+							host = parts[0]
+						}
+
+						if host != borrowingNode {
+							bIP := borrowedIP{block: b.CIDR.IPNet.String(), blockOwner: host, borrowingNode: borrowingNode}
+							bIP.addr = b.OrdinalToIP(i).String()
+							if _, ok := attributes.AttrSecondary[model.IPAMBlockAttributePod]; ok {
+								bIP.allocatedTo = fmt.Sprintf("%s/%s", attributes.AttrSecondary[model.IPAMBlockAttributeNamespace],
+									attributes.AttrSecondary[model.IPAMBlockAttributePod])
+								bIP.allocationType = "model.IPAMBlockAttributePod"
```
caseydavenport, line 86: Since this is user facing, I think it would be better to just set the type to "Pod" rather than "model.IPAMBlockAttributePod", since users don't know what that is

calicoctl/commands/ipam/show.go
gianlucam76: Wouldn't that be that too many on large cluster?

caseydavenport: We don't expect an exceptionally large number of these. At least not any different than we expect for other borrowed addresses.

Fundamentally, on a large cluster, there might be a lot of IP address allocations! But I think we should probably show these since most likely the reason they are there is because they were borrowed originally.

gianlucam76: Yeah I had misunderstood your feedback. Then after re-reading it I had deleted my question :-)

calicoctl/commands/ipam/show.go
```diff
@@ -19,85 +19,145 @@ import (
...
+
+				blockOwner := ""
+				if b.Affinity != nil {
+					// Affinity is in the form host:mgianluc-bz-09s4-kadm-node-2
+					// Remove "host:"
+					parts := strings.Split(*b.Affinity, ":")
+					if len(parts) == 2 {
+						blockOwner = parts[1]
+					} else {
+						blockOwner = parts[0]
```
caseydavenport, line 80: hm, when do we expect this to be a single part? I would have thought the value is always either `host:<somehost>` or empty.

calicoctl/commands/ipam/show.go
```diff
@@ -19,85 +19,145 @@ import (
...
+					}
+				}
+
+				if borrowingNode, ok := attributes.AttrSecondary[model.IPAMBlockAttributeNode]; ok {
```
caseydavenport, line 84: One thought - for _really_ old IP allocations, we used to not set the `AttrSecondary[model.IPAMBlockAttributeNode]` at all.

That means it's possible there are IPs in the data store that don't have this set, and for those addresses we have no way to tell if they are borrowed or not. 

I think it's probably OK to do what you have in this PR - don't show them at all. 

However, we might want to consider notifying the user in some way that we detected addresses that we can't make a decision on? WDYT? I'm not sure how we'd do that.

calicoctl/commands/ipam/show.go
gianlucam76: Thanks. As you mentioned, I also have always seen host:<somehost> but was not sure if that was always the case. I can remove such a check if not needed.

calicoctl/commands/ipam/show.go
caseydavenport: Maybe a message at the end would be good enough for now.

e.g. "Note: found N IP allocations without an explicit node association. Unable to detect if they are borrowed"

calicoctl/commands/ipam/show.go
gianlucam76: Thinking more I still need this. For borrowing host we don't use "host" prefix and I need to compare blockOwner with borrowing node (so I need to get rid of the host prefix here). This check will help.

caseydavenport:
Docs PR: https://github.com/projectcalico/calico/pull/3471

PR #2133, opened 2020-05-27
Add commands to migrate datastore

mgleung:
## Description
Adds the following commands to calicoctl in order to migrate from an etcdv3 datastore to a kubernetes one:
- export
- import
- lock
- unlock

## Todos
- [x] Tests
- [x] Documentation
- [x] Clean up code
- [x] Refactor code for a `migrate` subcommand

## Release Note

```release-note
None required
```


calicoctl/commands/import.go
```diff
@@ -0,0 +1,252 @@
...
+		return fmt.Errorf("Error while reading migration file: %s\n", err)
+	}
+
+	// Apply v3 API resources
+	err = updateV3Resources(v3Yaml)
+	if err != nil {
+		return fmt.Errorf("Failed to import v3 resources: %s\n", err)
+	}
+
+	// Create the cluster info resource. This should not exist since calico-node was only writing to the etcd datastore.
```
caseydavenport, line 115: I think the comment here is out of date

calicoctl/commands/import.go
```diff
@@ -0,0 +1,252 @@
...
+		fmt.Printf("Successfully applied %d IPAM resource(s)\n", results.numHandled)
+	} else {
+		if results.numHandled != 0 && len(results.resErrs) > 0 {
+			fmt.Printf("Partial success: ")
+			fmt.Printf("applied the first %d out of %d resources:\n", results.numHandled, results.numResources)
+		}
+		return fmt.Errorf("Hit error(s): %v", results.resErrs)
+	}
+
+	fmt.Print("Datastore information successfully imported. To complete the datastore migration, run `calicoctl unlock` and modify your calico configuration to match the kubenretes datastore.\n")
```
caseydavenport: I think this instruction is wrong - users will need to do a rolling update to point calico at kdd before running unlock.

calicoctl/commands/migrateipam.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	blockAffinities := []*BlockAffinityKVPair{}
+	for _, item := range blockAffinityKVList.KVPairs {
+		modelBlockAffinityKey, ok := item.Key.(model.BlockAffinityKey)
+		if !ok {
+			return fmt.Errorf("Could not convert %+v to a BlockAffinityKey", item.Key)
+		}
+
+		// Convert this to a key that has vlaues for json encoding.
```
caseydavenport: typo: vlaues

calicoctl/commands/migrateipam.go
```diff
@@ -0,0 +1,301 @@
...
+		}
+
+		blockAffinity, ok := item.Value.(*model.BlockAffinity)
+		if !ok {
+			return fmt.Errorf("Could not convert %+v to a BlockAffinity", item.Value)
+		}
+		blockAffinities = append(blockAffinities, &BlockAffinityKVPair{
+			Key:      &blockAffinityKey,
+			Value:    blockAffinity,
+			Revision: item.Revision,
```
caseydavenport, line 152: Do we care about revision and TTL information for the purposes of migration? I think probably not?

calicoctl/commands/migrateipam.go
```diff
@@ -0,0 +1,301 @@
...
+
+	ipamConfigKV, err := m.client.Get(ctx, model.IPAMConfigKey{}, "")
+	if err != nil {
+		// If the resource does not exist, do not throw the error
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			return err
+		}
+	}
+
+	// Convert all of the abstract KV Pairs into the appropriate types.
```
caseydavenport, line 118: I don't understand why we need to convert these from the abstract kvpairs to typed variants. Could you explain that to me?

calicoctl/commands/migrateipam.go
```diff
@@ -0,0 +1,301 @@
...
+
+	// Store the information
+	m.BlockAffinities = blockAffinities
+	m.IPAMBlocks = blocks
+	m.IPAMHandles = ipamHandles
+	m.IPAMConfig = ipamConfig
+	return nil
+}
+
+// TODO: Cannot use the backend client for this. Need a better way
```
caseydavenport: why not? Is this comment right?

calicoctl/commands/export.go
```diff
@@ -0,0 +1,275 @@
...
+			"<KIND>":   r,
+			"<NAME>":   []string{},
+			"--config": cf,
+			"--export": true,
+			"--output": "yaml",
+			"get":      true,
+		}
+		results := executeConfigCommand(mockArgs, actionGetOrList)
+		for _, resource := range results.resources {
+			// Remove relevant metadata because the --export flag does not for lists.
```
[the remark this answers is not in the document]
mgleung: meant to mean that the metadata isn't removed for lists. I'll update the comment

calicoctl/commands/migrateipam.go
mgleung: comment isn't correct and I forgot to remove it.

calicoctl/commands/migrateipam.go
mgleung: Revision information we definitely don't need. Not quite sure about TTL? Not sure if there was a value in the old datastore if we would want it moved over or since it's a new datastore, we don't care about it.

calicoctl/commands/migrateipam.go
mgleung: It is less for going from the golang representation to JSON than it is for going from JSON back to the golang representation. The KVPairs need to have the keys and values as concrete types for going from JSON to the golang objects since otherwise go does not know what object it is actually deserializing into. I could have created a separate `migrateIPAM` object that has the concrete types and used that one solely for going from JSON -> golang and then erase this logic and use this `migrateIPAM` object solely for going from golang -> JSON but I decided that having one object with a little more logic around it looked better to me.

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	} else {
+		if results.numHandled != 0 && len(results.resErrs) > 0 {
+			fmt.Printf("Partial success: ")
+			fmt.Printf("applied the first %d out of %d resources:\n", results.numHandled, results.numResources)
+		}
+		return fmt.Errorf("Hit error(s): %v", results.resErrs)
+	}
+
+	// TODO: Be more specific on how to do the calico configuration change.
+	fmt.Print("Datastore information successfully imported. To complete the datastore migration, run `calicoctl unlock` and modify your calico configuration to match the kubernetes datastore.\n")
```
caseydavenport, line 163: I don't think we should print this out here - unlocking the data store should only happen after the rolling update is complete.

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+func splitImportFile(filename string) ([]byte, []byte, []byte, error) {
+	// Get the appropriate file to read from
+	fname := filename
+	if filename == "-" {
+		fname = os.Stdin.Name()
+	}
+
+	b, err := ioutil.ReadFile(fname)
+	if err != nil {
+		return []byte{}, []byte{}, []byte{}, err
```
caseydavenport, line 177: I think this can just be `nil, nil, nil, err`?

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	err = applyV3(mockArgs)
+	if err != nil {
+		return fmt.Errorf("Failed to import v3 resources: %s\n", err)
+	}
+
+	return nil
+}
+
+func importCRDs(cfg *apiconfig.CalicoAPIConfig) error {
+	// TODO: Add the apiextensions clientset and handling logic to the libcalico-go kubeclient
```
caseydavenport, line 296: I don't think we want to do this TODO.

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	_, err = c.ClusterInformation().Update(ctx, clusterinfo, options.SetOptions{})
+	if err != nil {
+		return fmt.Errorf("Error updating current cluster info for migration: %s", err)
+	}
+
+	return nil
+}
+
+func updateV3Resources(cf string, data []byte) error {
+	// Create tempfile so the v3 resources can be created using Apply
```
caseydavenport, line 271: This is probably fine, but I was originally imagining we'd do this all through libcalico-go rather than writing to file and pushing it through all of the calicoctl processing.

Again, probably fine for now if it's working. Might find a reason later to do it differently though, e.g., if there is some calicoctl validation that we don't want in this codepath.

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	return nil
+}
+
+func importCRDs(cfg *apiconfig.CalicoAPIConfig) error {
+	// TODO: Add the apiextensions clientset and handling logic to the libcalico-go kubeclient
+	// Start a kube client
+	// Use the kubernetes client code to load the kubeconfig file and combine it with the overrides.
+	configOverrides := &clientcmd.ConfigOverrides{}
```
caseydavenport, line 299: Could we replace all of this code by just importing this public function? https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/k8s.go#L213

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	return nil
+}
+
+func applyV3(args map[string]interface{}) error {
+	results := common.ExecuteConfigCommand(args, common.ActionApply)
+	log.Infof("results: %+v", results)
+
+	if results.FileInvalid {
```
caseydavenport, line 376: This feels like a lot of duplication in code. 

Could we instead keep these fields private on the results structure, and just commonize it via a method?

e.g., 

```
if err := results.Error(); err != nil {
  return err
}
```

caseydavenport: To be honest, I don't think it really matters much to the user if some resources applied and some didn't for this use-case anyway. I think it's a boolean - did it work? Or did it fail?

So, we can reduce some of complexity of tracking that and just return an error or not.

Becasue, if they hit an error, they need to clear and start over again anyway.

calicoctl/commands/migrate/unlock.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/docopt/docopt-go"
+
+	"github.com/projectcalico/calicoctl/calicoctl/commands/clientmgr"
+	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
+	"github.com/projectcalico/libcalico-go/lib/options"
+)
+
+func Unlock(args []string) error {
+	doc := `Usage:
+  calicoctl migrate unlock [--config=<CONFIG>]
```
caseydavenport, line 31: I think we said this should be `calicoctl datastore migrate lock`?

and `calicoctl datastore migrate import/export`?

calicoctl/commands/migrate/import.go
mgleung: I thought it would be nicer to not have what is essentially client level code living in calicoctl since it feels like it would get more use in libcalico-go. Granted, we don't have a use case for adding it yet, but thought it would be a nice thing to have.

calicoctl/commands/migrate/import.go
mgleung: Only reason I did it this way is that it let me reuse a lot of the code as it was written so I wouldn't need to duplicate code to handle things like partially applying some of the resources or parsing the YAML files into the appropriate resources before applying them.

calicoctl/commands/migrate/import.go
mgleung: Unfortunately the only difference between the client that we already have is that it does not include the apiextensions clientset which is what we require in order to modify CRDs. Ideally following up with the `TODO` for the function would add that clientset so we can use that function.

calicoctl/commands/migrate/import.go
mgleung: A lot of the separate handling code differs slightly depending on the command it is being used from. The only way to keep the fields private would be to add separate `Error` functions i.e. `ErrorGetOrList`, `ErrorApply`, `ErrorCreate`, etc. which I opted against since the only truly duplicate piece of code is this invocation here and the one used by Apply.

calicoctl/commands/migrate/unlock.go
mgleung: Oops, glanced over it. I'll fix it.

calicoctl/commands/migrate/import.go
mgleung: Would swapping the order of the instructions work? I feel like it is useful to be as prescriptive as possible since people may forget to unlock the datastore.

calicoctl/commands/migrate/import.go
caseydavenport: I'd prefer to put code closer to where it's used - nothing in libcalico uses apiextensions.

calicoctl/commands/migrate/import.go
caseydavenport: But barring that, I think it should just say "refer to the upgrade docs for next steps"

calicoctl/commands/migrate/import.go
caseydavenport: yeah, understood. thinks this is good for now.

calicoctl/commands/migrate/import.go
```diff
@@ -0,0 +1,409 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/docopt/docopt-go"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
```
caseydavenport, line 28: I think this is the wrong package to use for clientset - should use the one in client-go.

calicoctl/commands/crds/crds.go
```diff
@@ -0,0 +1,283 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package crds
+
+import (
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
caseydavenport, line 18: Same here - I don't think we should be importing the apiextensions-apiserver anywhere in calicoctl.

caseydavenport: Should be `k8s.io/api`

calicoctl/commands/migrate/import.go
caseydavenport: I think we can still re-use it just to build the config, though.

e.g.,

```go
cfg, _, err := k8s.CreateKubernetesClientSet()
if err != nil {...}

cs := clientset.NewForConfig(cfg)

File: calicoctl/commands/migrate/import.go (hunk 15 above)
Comment (mgleung) [reply calicoctl/commands/migrate/import.go:28]: Talked about this offline. Clarified that the apiextensions CRD object is not available as a part of the normal client-go clientset.

File: calicoctl/commands/crds/crds.go (hunk 16 above)
Comment (mgleung) [reply calicoctl/commands/crds/crds.go]: Talked about this offline. The apiextensions CRD objects are not found in the k8s.io/api.

File: calicoctl/commands/migrate/import.go (hunk 13 above)
Comment (mgleung) [reply calicoctl/commands/migrate/import.go:347]: I shortened a lot of the error handling since, as you pointed out, it's not really necessary for this use case.

File: calicoctl/calicoctl.go [hunk 17]
```diff
@@ -28,20 +28,23 @@ func main() {
...
+    get                  Get a resource identified by file, stdin or resource type and
+                         name.
+    label                Add or update labels of resources.
+    convert              Convert config files between different API versions.
+    ipam                 IP address management.
+    node                 Calico node management.
+    version              Display the version of calicoctl.
+    export               Export the Calico datastore objects for migration
+    import               Import the Calico datastore objects for migration
+    datastore migrate    Migrate the Calico datastore resources.
```
caseydavenport, line 47: should this just be `datastore` here?

PR #2134, opened 2020-06-01
POC: extract env vars from kubeconfig

p-tang:
## Description
This is a possible starting point to check kubeconfig file for environment variables related to datastore type, if it is not set elsewhere. Manually tested using an AWS EKS cluster.

### DIFF
--- a/calicoctl/commands/clientmgr/client.go
+++ b/calicoctl/commands/clientmgr/client.go
@@ -17,8 +17,11 @@ package clientmgr
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	log "github.com/sirupsen/logrus"
+
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/projectcalico/calicoctl/calicoctl/commands/constants"
 	"github.com/projectcalico/libcalico-go/lib/apiconfig"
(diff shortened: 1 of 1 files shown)


calicoctl/commands/clientmgr/client.go
```diff
@@ -30,6 +33,37 @@ import (
...
+
+	// Look up datastore type
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		log.Info("Error loading config")
+		return nil, err
+	}
+
+	// If datastore type env var found, set it
+	datastoreType := apiconfig.EtcdV3
```
caseydavenport, line 54: Maybe we should default to "Kubernetes" unless "ETCD_ENDPOINTS" is specified in the config?

calicoctl/commands/clientmgr/client.go
```diff
@@ -30,6 +33,37 @@ import (
...
+	if !exists {
+		kubeconfigPath = filepath.Join(
+			os.Getenv("HOME"), ".kube", "config",
+		)
+	}
+	cfg.Spec.KubeConfig.Kubeconfig = kubeconfigPath
+
+	// Look up datastore type
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
```
robbrockbank, line 48: IIUC this will fail if the kubeconfig does not exist, and therefore this is changing the behavior of calicoctl such that it will only be useable if a kubeconfig is present. I don't think that is correct.  If the kubeconfig is not present then we should just pull the environments from the host - this is already what happens in LoadClientConfig() at line 35, so I think if the kubeconfig does not exist then we just use the original client config.

calicoctl/commands/clientmgr/client.go
```diff
@@ -30,6 +33,37 @@ import (
...
+		log.Info("Error loading config")
+		return nil, err
+	}
+
+	// If datastore type env var found, set it
+	datastoreType := apiconfig.EtcdV3
+	for _, v := range config.ExecProvider.Env {
+		if v.Name == "DATASTORE_TYPE" {
```
robbrockbank, line 56: We need to copy across more than just the datastore type.

robbrockbank:
If we are trying to allow calicoctl to be run as it's own binary and as a plugin for kubectl, but to have the exec environments speyified in kubeconfig - then I would change the logic a little.  I also agree with @caseydavenport that if DATASTORE_TYPE is not set *AND* ETCD_ENDPOINTS is not set then we should default to kubernetes as the datastore type.

-  Load the kubeconfig if it exists, otherwise just continue.
-  If kubeconfig exists, grab the ExecConfig and set the local environments (if this is running as a plugin, this will just copy over the environments with the same value), if not running as a plugin
-  Load the Calico config.  This will load from the updated environments.

I would change the default datastore logic in libcalico-go LoadClientConfig

caseydavenport:
We have since merged PRs that change the datastore default, as well as how kubeconfig and datastore type is determined

PR #2146, opened 2020-06-26
Use scratch based Image, with only required files.

krishgobinath:
## Description

Build scratch based Image, with required tools, utils, and configuration files from UBI image.

- Create an empty root file system on base Image.

- Copy only required Linux tools, utils from base Image, and Include all dependencies in the root FS.

-  Map temporary root filesystem as scratch based image filesystem.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -12,7 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1 as ubi
+RUN microdnf update
+
+# Create Empty root filesystem on /rootfs, copy only required linux utils
+# and supporting files.This filesystem act as placeholder for all
+# files required in container.
+
+WORKDIR /rootfs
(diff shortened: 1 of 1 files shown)


krishgobinath:
> @krishgobinath does this approach preserve enough package metadata for CVE scanners to identify vulnerabilities?

This approach doesn't have any metadata related to package or distro information. Hence scanner cannot find any vulnerabilities through packaging information. However, the scanner can scan through all binaries and dependencies present in the final image.

bcreane:
@krishgobinath - my take is this approach of hand-crafting `find`, `grep` etc. statements to manually copy specific binaries and libraries is not robust. When the dependencies change for a specific binary, or our `grep` statement suddenly includes more (or less) than we expect because of a change in the distro, we'll be debugging tricky file system issues or searching for improperly added dependencies that expand our CVE exposure.

For simple images that only include a statically linked golang binary and maybe a few directories, this approach of hand crafting the directories + files seems fine.

But calicoctl doesn't fall into that category. 

Let's discuss the approach we want to take for building scratch images that include more than just a few directories and a statically linked golang binary. @gianlucam76 and @doublek should be involved as well. My guess is that it looks similar to the approach @fasaxc took with node.

krishgobinath:
Closing this PR, opening new PR with inbuilt application to replace  `sh` utilis.

PR #2148, opened 2020-07-10
Fix CVEs, use scratch based Image

krishgobinath:
## Description
Build scratch based Image, with only statically built applications[calicoctl & calicoctl-wait]
1) Built small application to wait for in an indefinite loop.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -12,7 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2 as ubi
+
+RUN mkdir /licenses
+COPY LICENSE /licenses
+
+FROM scratch
 ARG GIT_VERSION=unknown
 
 LABEL name="Calico CLI tool" \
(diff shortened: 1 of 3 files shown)


Dockerfile.amd64
```diff
@@ -24,9 +24,7 @@ LABEL name="Calico CLI tool" \
       maintainer="maintainers@projectcalico.org"
 
 ADD bin/calicoctl-linux-amd64 /calicoctl
-
-RUN mkdir /licenses
```
robbrockbank, line 28: I'm not sure what this was originally for so I've no idea if removing it is sensible.  Do you have any more details on this?

calicoctl-wait/calicoctl-wait.go
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2020 Tigera, Inc. All rights reserved.
```
robbrockbank, line 1: What is this for?

Dockerfile.amd64
krishgobinath: I don't know why that file is required. I'll bring back the original file.

calicoctl-wait/calicoctl-wait.go
krishgobinath: The calico-wait tool is used to replace Linux utils used in the container image and go for a pure scratch image.

calicoctl-wait/calicoctl-wait.go
```diff
@@ -0,0 +1,26 @@
...
+package main
+
+import (
+	"time"
+)
+
+func main() {
+	// Spin continously in tight loop
+	for {
+		time.Sleep(500 * time.Millisecond)
```
doublek, line 24: This seems too low, should we just sleep longer?

bcreane: @krishgobinath - I prematurely merged - didn't notice that @doublek was pending. Can you open another PR to address this comment?

krishgobinath: I had unit tested with 500ms and 1 second, I didn't spot any difference between those 2 in terms of CPU utilization rate. The worst case is if we don't sleep then CPU utilization rate is 100%.

krishgobinath: @bcreane If required I'll do that.

bcreane: I chatted with @doublek - he's okay skipping this change. So ... never mind.

PR #2150, opened 2020-07-11
Fix issue making gen-crds

joshuactm:
## Description
Running `make gen-crds` was failing.  Tracked issue down to EXTRA_DOCKER_ARGS needing to be declared before including Makefile.common.

## Todos
Now that `make gen-crds` works, it highlights the issue that `commands/crds/crds.go` is out of date.

## Release Note
```release-note
None required
```


joshuactm:
Issue was actually caused by not setting LOCAL_BUILD

PR #2158, opened 2020-08-07
Calico util to apply manifests

krishgobinath:
## Description
1) Operator based Open shift Install requires linux `shell` to install few manifests. Replace this `shell` requirement with a `calicoctl-util `tool.


caseydavenport:
@tmjd does that sound like the right approach to you?

tmjd:
I thought it was going to be a new feature of calicoctl to be able to apply the files in a directory. I did not think there was going to be a new binary.
The caveat is that doing the apply multiple times does not error if it is redone.

PR #2161, opened 2020-08-12
Add ability to specify directory as filename parameter

robbrockbank:
## Description
Added the ability to specify manifests by directory.

If the file is a directory, calicoctl will search for .yaml, .yml and .json files in the immediate directory, and optionally recursively search sub directories.  An additional option is also added to allow calicoctl to run without error if the specific file or directory are empty (note that if --filename points to a non-existent 

## Todos
- [X] Tests (ST)
- [ ] Documentation
- [ ] Release note

Could benefit from some STs I think.  We have UTs testing the actual file loading behavior.  Not planning to add STs for 3.16 though.

## Release Note

```release-note
Various commands using file input now support specifying a directory to support actions on multiple manifests, these commands also include optional recursive searches for .yaml, .yml and .json files, and additional options to disable failing if the files or directory contain no resources.
```

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -29,14 +29,14 @@ func main() {
 	doc := `Usage:
   calicoctl [options] <command> [<args>...]
 
-    create       Create a resource by filename or stdin.
-    replace      Replace a resource by filename or stdin.
-    apply        Apply a resource by filename or stdin.  This creates a resource
+    create       Create a resource by file, directory or stdin.
+    replace      Replace a resource by file, directory or stdin.
+    apply        Apply a resource by file, directory or stdin.  This creates a resource
                  if it does not exist, and replaces a resource if it does exists.
     patch        Patch a pre-exisiting resource in place.
-    delete       Delete a resource identified by file, stdin or resource type and
+    delete       Delete a resource identified by file, directory, stdin or resource type and
                  name.
-    get          Get a resource identified by file, stdin or resource type and
+    get          Get a resource identified by file, directory, stdin or resource type and
                  name.
     label        Add or update labels of resources.
     convert      Convert config files between different API versions.
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -20,8 +20,9 @@ import (
 	"strings"
 
 	"github.com/docopt/docopt-go"
-	"github.com/projectcalico/calicoctl/calicoctl/commands"
 	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calicoctl/calicoctl/commands"
 )
 
 func main() {
--- a/calicoctl/commands/apply.go
+++ b/calicoctl/commands/apply.go
@@ -27,7 +27,8 @@ import (
 
 func Apply(args []string) error {
 	doc := constants.DatastoreIntro + `Usage:
-  calicoctl apply --filename=<FILENAME> [--recursive] [--config=<CONFIG>] [--namespace=<NS>]
+  calicoctl apply --filename=<FILENAME> [--recursive] [--skip-empty]
+                  [--config=<CONFIG>] [--namespace=<NS>]
 
 Examples:
   # Apply a policy using the data in policy.yaml.
@@ -43,6 +44,8 @@ Options:
                             invoked for each .json .yaml and .yml file within that directory,
                             terminating after the first failure.
   -R --recursive            Process the filename specified in -f or --filename recursively.
+     --skip-empty           Do not error if any files or directory specified using -f or --filename contain no
+                            data.
   -c --config=<CONFIG>      Path to the file containing connection
                             configuration in YAML or JSON format.
                             [default: ` + constants.DefaultConfigPath + `]
@@ -100,10 +103,14 @@ Description:
 
 	if results.FileInvalid {
 		return fmt.Errorf("Failed to execute command: %v", results.Err)
+	} else if results.NumResources == 0 {
+		// No resources specified. If there is an associated error use that, otherwise print message with no error.
+		if results.Err != nil {
+			return results.Err
+		}
+		fmt.Println("No resources specified")
 	} else if results.NumHandled == 0 {
-		if results.NumResources == 0 {
-			return fmt.Errorf("No resources specified in file")
-		} else if results.NumResources == 1 {
+		if results.NumResources == 1 {
 			return fmt.Errorf("Failed to apply '%s' resource: %v", results.SingleKind, results.ResErrs)
 		} else if results.SingleKind != "" {
 			return fmt.Errorf("Failed to apply any '%s' resources: %v", results.SingleKind, results.ResErrs)
--- a/calicoctl/commands/apply.go
+++ b/calicoctl/commands/apply.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016-2019 Tigera, Inc. All rights reserved.
+// Copyright (c) 2016-2020 Tigera, Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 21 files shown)


bcreane:
😮

calicoctl/commands/apply.go
```diff
@@ -92,40 +95,42 @@ Description:
...
-	} else if len(results.ResErrs) == 0 {
-		if results.SingleKind != "" {
-			fmt.Printf("Successfully applied %d '%s' resource(s)\n", results.NumHandled, results.SingleKind)
-		} else {
-			fmt.Printf("Successfully applied %d resource(s)\n", results.NumHandled)
-		}
-	} else {
-		if results.NumHandled-len(results.ResErrs) > 0 {
-			fmt.Printf("Partial success: ")
+	return file.Iter(parsedArgs, func(updatedArgs map[string]interface{}) error {
```
robbrockbank, line 98: This is the main change in each command.  I'm wrapping the command in this iter and using the "updatedArgs" in place of the parsedArgs.

calicoctl/commands/file/iter.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// File name is specified.
+	f, ok := parsedArgs["--filename"].(string)
+	if !ok {
+		// Handle invalid filename by invoking standard processing.
+		return cb(parsedArgs)
+	}
+
+	if parsedArgs["--filename"] == "-" {
```
doublek, line 26: Probably use `f` instead of `parsedArgs["--filename"]`.

calicoctl/commands/file/iter.go
```diff
@@ -0,0 +1,73 @@
...
+	// Handle directory by walking the directory contents and performing the action on each manifest file.
+	return filepath.Walk(f,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if info.IsDir() {
+				return nil
+			}
```
doublek, line 45: Since we aren't returning a SkipDir here, this means we will process this directory as well?

calicoctl/commands/apply.go
```diff
@@ -39,7 +40,9 @@ Examples:
 Options:
   -h --help                 Show this screen.
   -f --filename=<FILENAME>  Filename to use to apply the resource.  If set to
-                            "-" loads from stdin.
+                            "-" loads from stdin. If filename is a directory, this command is
+                            invoked for each .json .yaml and .yml file within that directory,
+                            terminating after the first failure.
```
doublek, line 45: Do we process sub-directories or rather recursively?

calicoctl/commands/file/iter.go
robbrockbank: Oooh that was not intentional.  I will fix.

calicoctl/commands/apply.go
robbrockbank: I don't personally think this should recurse

doublek: I was looking at kubectl's behavior and [kubectl will only recurse on an explicit `--recursive` flag](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#create). So I am okay with not recursing here. Can you update the doc/comment to make the explicit?

calicoctl/commands/get.go
```diff
@@ -43,7 +43,10 @@ Examples:
 Options:
   -h --help                    Show this screen.
   -f --filename=<FILENAME>     Filename to use to get the resource.  If set to
-                               "-" loads from stdin.
+                               "-" loads from stdin. If filename is a directory, this command is
+                               invoked for each .json .yaml and .yml file within that directory,
+                               terminating after the first failure
```
caseydavenport, line 48: missing a period at the end of this sentence

calicoctl/commands/replace.go
```diff
@@ -40,7 +41,10 @@ Examples:
 Options:
   -h --help                  Show this screen.
   -f --filename=<FILENAME>   Filename to use to replace the resource.  If set
-                             to "-" loads from stdin.
+                             to "-" loads from stdin. If filename is a directory, this command is
+                             invoked for each .json .yaml and .yml file within that directory,
+                             terminating after the first failure
```
caseydavenport, line 46: same here - need that "full stop" to really drive it home.

calicoctl/commands/create.go
```diff
@@ -39,7 +39,10 @@ Examples:
 Options:
   -h --help                 Show this screen.
   -f --filename=<FILENAME>  Filename to use to create the resource.  If set to
-                            "-" loads from stdin.
+                            "-" loads from stdin. If filename is a directory, this command is
+                            invoked for each .json .yaml and .yml file within that directory,
+                            terminating after the first failure
```
caseydavenport, line 44: Another one needs a dot

calicoctl/commands/delete.go
```diff
@@ -46,7 +46,10 @@ Options:
   -s --skip-not-exists      Skip over and treat as successful, resources that
                             don't exist.
   -f --filename=<FILENAME>  Filename to use to delete the resource.  If set to
-                            "-" loads from stdin.
+                            "-" loads from stdin. If filename is a directory, this command is
+                            invoked for each .json .yaml and .yml file within that directory,
+                            terminating after the first failure
```
caseydavenport, line 51: I wasn't going to comment on all of the missing `.`, but here I am.

calicoctl/commands/file/iter_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"errors"
+	"io/ioutil"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// Check parsing of BIRD peer lines
```
doublek, line 28: outdated comment.

calicoctl/commands/common/resources.go
```diff
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/projectcalico/calicoctl/calicoctl/commands/file"
```
doublek, line 23: Does this need to be goimported? I see different grouping below?

robbrockbank: it was - but that only shifts things out of place it doesn't necessarily re-group things.  I've fixed.

calicoctl/commands/file/iter_test.go
robbrockbank: oops

calicoctl/commands/delete.go
robbrockbank: hahahaha

doublek:
@robbrockbank can you add a release note to the description.

calicoctl/calicoctl.go
```diff
@@ -20,8 +20,9 @@ import (
 	"strings"
 
 	"github.com/docopt/docopt-go"
-	"github.com/projectcalico/calicoctl/calicoctl/commands"
 	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calicoctl/calicoctl/commands"
```
doublek, line 25: One more comment while I was looking at docs. Can you update the usage doc below for the appropriate sub commands to state "filename, directory or stdin"?

calicoctl/commands/get.go
```diff
@@ -179,6 +184,9 @@ Description:
 
 	if results.FileInvalid {
 		return fmt.Errorf("Failed to execute command: %v", results.Err)
+	} else if results.NumResources == 0 {
+		// No resources is generally an error case unless the --no-error-on-empty option is explicitly set.
```
doublek, line 188: nit: flag in comment needs to be updated.

calicoctl/commands/get.go
```diff
@@ -179,6 +184,9 @@ Description:
...
+		fmt.Println("No resources specified in file")
```
doublek, line 189: Looks like this has broken a couple of tests.

PR #2169, opened 2020-09-02
Syntax Validity Checker Option

nuxeric:
## Description
Hello! This is my first time contributing and I am very excited!
This PR was made in relation to this issue: https://github.com/projectcalico/calicoctl/issues/2117
I added in an option called `--dry-run` which allows a user to add a step to their CI/CD pipeline maybe or something along those lines that allows users to check if their file has no syntactical issues.
Similar to terraforms's validate : https://www.terraform.io/docs/commands/validate.html

I made changes such that before calicoctl accesses any kubernetes APIs or creates the calicoctl API client it "break"s the flow of execution and returns a statement that states that the dry run has passed successfully. If it throws an error earlier the user will recieve that error message as per usual

Thoughts and critiques would be very much appreciated!

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
I'm not sure if I need a release note 

```release-note
None required?
```


lmm:
Hey @nuxeric, sorry for the delay in getting to your PR. Thanks for taking an early stab :)

Rather than have calicoctl do nothing if given `--dry-run`, I think we might want calicoctl to do a "dry run" application of the provided input.

So I'm not actually familiar with the existing Kubernetes dry-run functionality so I played with it a bit as described in this blog: https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/

I think we need to figure out what we want (client-side or server-side dry-run) and what sort of dry-run experience we want to provide through calicoctl. It seems like k8s client-side(?) dry-run functionality is already available as options. E.g. https://github.com/kubernetes/apimachinery/blob/2a468d8caa4629500c086c621f2cb3402168c6a3/pkg/apis/meta/v1/types.go#L606

@caseydavenport might know about k8s dry-run and have some ideas on this.

nuxeric:
@lmm 
Super excited to work on this! I will wait on @caseydevnport (:

caseydavenport:
So calicoctl has some interesting distinctions from kubectl that we need to take into consideration here. Namely that it supports writing data both as k8s CRDs, but also directly into etcd.

For the etcd case, I don't think we can do server-side validation, because there is no "server" per-se. 

There are two types of validation:
- validating that the yaml is valid
- validating the the content of the yaml is correct.

For the latter, most of our validation is currently performed client side in the `lib/validator` package in libcalico-go: https://github.com/projectcalico/libcalico-go/tree/master/lib/validator/v3

So, I think we'll want to use that here as well in order to guarantee we're running the same validation for `--dry-run` as we would for a normal request. 

Really, I'd guess we want to do almost exactly the same work, with the exception that for `--dry-run` we don't want to actually store the data!

If I understand this PR correctly, it's performing the first type of validation ("is this valid yaml?") but it returns before it can get to the content validation, since that is currently done within the CRUD operations in libcalico-go. I think we'll want it to perform both types.

caseydavenport:
@nuxeric what do you think about calling the validation package functions on objects after they have been loaded?

That would catch more nuanced errors (e.g., a typo in a value) in addition to just yaml syntax errors.

nuxeric:
> @nuxeric what do you think about calling the validation package functions on objects after they have been loaded?
> 
> That would catch more nuanced errors (e.g., a typo in a value) in addition to just yaml syntax errors.

I will totally look into this (I am super sorry it's been taking so long school has just been super crazy)

I will investigate over the weekend and private msg you on Slack if I run into any issues!

caseydavenport:
> I am super sorry it's been taking so long school has just been super crazy)

Sorry, I didn't mean to rush you! Please take your time. I just wanted to make sure you weren't blocked on me.

caseydavenport:
@nuxeric is this something you're still planning to work on? It looks like it could use a rebase.

nuxeric:
> @nuxeric is this something you're still planning to work on? It looks like it could use a rebase.

Yes, I will submit new code over this coming weekend, thank you for the reminder!!

caseydavenport:
Moving to v3.20 - please give a shout when this is ready for re-review.

nuxeric:
> Moving to v3.20 - please give a shout when this is ready for re-review.

I am finally finished all my school work (finals included!!) and am ready to work on this!

While I was working on this before I messed up my current development branch (I think I merged my version of master into the development branch or somethign weird), I am going to close this PR and reopen it with another branch with all the changes that we discused here)

PR #2186, opened 2020-09-21
Fix datastore migration to properly assign Kubernetes nodenames in the migrated IPAM data

mgleung:
## Description

Fixes issues where the etcd node names did not match the Kubernetes datastore node names. This migrates all of the relevant host information in IPAM to the correct Kubernetes datastore node name. This inserts the new Kubernetes datastore node names into the following IPAM fields:
- IPAM block allocation attributes
- IPAM block block affinity field
- IPAM block affinity host field
- IPAM block handle ID for `ipip-tunnel-addr` handles
- IPAM handle ID for `ipip-tunnel-addr` handles

This also fixes a similar issue where node names were not migrated properly for per-node FelixConfigurations.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix a bug with mismatched node names when migrating IPAM data from etcd to Kubernetes datastores.
```

### DIFF
--- a/calicoctl/commands/datastore/migrate/export.go
+++ b/calicoctl/commands/datastore/migrate/export.go
@@ -39,7 +39,6 @@ var allV3Resources []string = []string{
 	"ippools",
 	"bgpconfig",
 	"bgppeers",
-	"felixconfigs",
 	"globalnetworkpolicies",
 	"globalnetworksets",
 	"heps",
@@ -47,6 +46,7 @@ var allV3Resources []string = []string{
 	"networkpolicies",
 	"networksets",
 	"nodes",
+	"felixconfigs",
 }
 
 var resourceDisplayMap map[string]string = map[string]string{
@@ -237,6 +237,28 @@ Description:
 				})
 				if err != nil {
 					return fmt.Errorf("Unable to process metadata for export for Node resource: %s", err)
+				}
+			}
+
+			// Felix configs may also need to be modified if node names do not match the Kubernetes node names.
+			if r == "felixconfigs" {
+				err := meta.EachListItem(resource, func(obj runtime.Object) error {
+					felixConfig, ok := obj.(*apiv3.FelixConfiguration)
+					if !ok {
+						return fmt.Errorf("Failed to convert resource to FelixConfiguration object for migration processing: %+v", obj)
+					}
+
+					if strings.HasPrefix(felixConfig.GetObjectMeta().GetName(), "node.") {
+						etcdNodeName := strings.TrimPrefix(felixConfig.GetObjectMeta().GetName(), "node.")
+						if nodename, ok := etcdToKddNodeMap[etcdNodeName]; ok {
+							felixConfig.GetObjectMeta().SetName(fmt.Sprintf("node.%s", nodename))
+						}
+					}
+
+					return nil
+				})
+				if err != nil {
+					return fmt.Errorf("Unable to process metadata for export for FelixConfiguration resource: %s", err)
 				}
 			}
 		}
--- a/calicoctl/commands/datastore/migrate/export.go
+++ b/calicoctl/commands/datastore/migrate/export.go
@@ -147,6 +147,7 @@ Description:
 	}
 
 	rp := common.ResourcePrinterYAML{}
+	etcdToKddNodeMap := make(map[string]string)
 	// Loop through all the resource types to retrieve every resource available by the v3 API.
 	for _, r := range allV3Resources {
 		mockArgs := map[string]interface{}{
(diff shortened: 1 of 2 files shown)


calicoctl/commands/datastore/migrate/migrateipam.go
```diff
@@ -120,6 +126,35 @@ func (m *migrateIPAM) PullFromDatastore() error {
...
+			for i, allocationAttribute := range block.Attributes {
+				// Update the node name if it has a corresponding Kubernetes node name
+				attrNodeName := strings.ReplaceAll(allocationAttribute.AttrSecondary["node"], ".", "-")
+				if nodeName, ok := m.nodeMap[attrNodeName]; ok {
+					block.Attributes[i].AttrSecondary["node"] = nodeName
+				}
+
+				// Update the handle ID for ipip tunnel addresses
+				if allocationAttribute.AttrPrimary != nil {
+					if strings.HasPrefix(*allocationAttribute.AttrPrimary, "ipip-tunnel-addr-") {
```
caseydavenport, line 141: What about the VXLAN and wireguard tunnel addresses? Probably need those here as well.

calicoctl/commands/datastore/migrate/export.go
```diff
@@ -297,6 +321,7 @@ Description:
 
 	// Use the v1 API in order to retrieve IPAM resources
 	ipam := NewMigrateIPAM(client)
+	ipam.SetNodeMap(etcdToKddNodeMap)
```
caseydavenport, line 324: Feels like this should just be another argument to the NewMigrateIPAM, given its required?

calicoctl/commands/datastore/migrate/migrateipam.go
```diff
@@ -120,6 +128,38 @@ func (m *migrateIPAM) PullFromDatastore() error {
 		if !ok {
 			return fmt.Errorf("Could not convert %+v to an AllocationBlock", item.Value)
 		}
+
+		// Update node names in the block to match the Kubernetes node
+		if m.nodeMap != nil {
```
caseydavenport, line 133: Is this ever going to actually be nil?

calicoctl/commands/datastore/migrate/migrateipam.go
```diff
@@ -120,6 +128,38 @@ func (m *migrateIPAM) PullFromDatastore() error {
...
+			for i, allocationAttribute := range block.Attributes {
+				// Update the node name if it has a corresponding Kubernetes node name
+				if nodeName, ok := m.nodeMap[allocationAttribute.AttrSecondary["node"]]; ok {
+					block.Attributes[i].AttrSecondary["node"] = nodeName
+				}
+
+				// Update the handle ID for ipip tunnel addresses
```
caseydavenport, line 140: "ipip tunnel address" -> "any tunnel addresses"

calicoctl/commands/datastore/migrate/migrateipam.go
```diff
@@ -120,6 +128,38 @@ func (m *migrateIPAM) PullFromDatastore() error {
...
+				if allocationAttribute.AttrPrimary != nil {
+					handlePrefixReplaced := false
```
caseydavenport, line 142: Why do we want this variable here? 

IIUC, it will make sure that we only replace one tunnel address handle and skip any subsequent ones. Is that right? 

I think if we have multiple tunnel addresses present we should replace all of them.

calicoctl/commands/datastore/migrate/migrateipam.go
```diff
@@ -147,10 +201,27 @@ func (m *migrateIPAM) PullFromDatastore() error {
 
 	ipamHandles := []*IPAMHandleKVPair{}
 	for _, item := range ipamHandleKVList.KVPairs {
-		handleKey, err := model.KeyToDefaultPath(item.Key)
+		// Update IPAM handle ID for IPIP tunnel to include the Kubernetes node name.
```
caseydavenport, line 204: IPIP tunnel -> tunnel

calicoctl/commands/datastore/migrate/export.go
mgleung: I didn't add it as an argument since it was really only required for export. Since I reuse the object for import, I didn't want to have to pass in an empty mapping just for initialization purposes there.

calicoctl/commands/datastore/migrate/migrateipam.go
mgleung: Yup, it's for skipping once one is found. I thought we could only have one tunnel so I added it, if we can have multiple (which it sounds like) then this is wrong and I need to remove it.

calicoctl/commands/datastore/migrate/migrateipam.go
mgleung: It *shouldn't* ever be nil but I thought I should add the check just in case since it'll cause things to throw nil pointer exceptions.

caseydavenport (approved):
@mgleung this looks good to me, but I'd really like to see a test or two added as part of this.

mgleung:
Oops, sorry, hit merge before reading through the comment. I'll see what I can add in a follow up PR

PR #2199, opened 2020-10-05
Add IPAM migration unit tests and node handling for BGP configurations

mgleung:
## Description
Follow up to https://github.com/projectcalico/calicoctl/pull/2186 to add some UTs

This PR also fixes the following bugs in the migration logic:
- BGP configurations were missing handling for per-node BGP configuration migrations in the case where the etcd node name differs from the Kubernetes datastore node name
- Long running clusters may store Felix configuration iptables fields in the Calico v1 API style. This converts those fields to match with the v3 validations.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/datastore/migrate/migrateipam_test.go
+++ b/calicoctl/commands/datastore/migrate/migrateipam_test.go
@@ -31,10 +31,10 @@ import (
 )
 
 var (
-	nodeName           = "original-node"
-	newNodeName        = "changed-node"
+	nodeName           = "etcdNodeName"
+	newNodeName        = "k8sNodeName"
 	blockAffinityField = fmt.Sprintf("host:%s", nodeName)
-	ipipTunnelHandle   = "ipip-tunnel-addr-original-node"
+	ipipTunnelHandle   = "ipip-tunnel-addr-etcdNodeName"
 )
 
 var _ = Describe("IPAM migration handling", func() {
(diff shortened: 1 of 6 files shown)


calicoctl/commands/datastore/migrate/migrateipam_test.go
```diff
@@ -0,0 +1,342 @@
...
+		Expect(*migrateIPAM.IPAMBlocks[0].Value.Attributes[0].AttrPrimary).To(Equal(fmt.Sprintf("ipip-tunnel-addr-%s", nodeName)))
+		Expect(migrateIPAM.IPAMBlocks[0].Value.Attributes[0].AttrSecondary["node"]).To(Equal(nodeName))
+
+		// Check that the block affinity attributes were changed correctly
+		newAffinityKeyPath, err := model.KeyToDefaultPath(affinity1.Key)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(migrateIPAM.BlockAffinities).To(HaveLen(1))
+		Expect(migrateIPAM.BlockAffinities[0].Key).To(Equal(newAffinityKeyPath))
+
+		// Check that the IPAM handle attributes were changed correctly
```
caseydavenport, line 166: same here

caseydavenport (changes requested):
A couple minor comments, but overall looks good.

Do we need a test for per-node FelixConfiguration / BGPConfiguration as well?

mgleung:
Forgot to even add the handling for the BGP configs. Added that and the appropriate tests for that and the felix configs.

caseydavenport (approved):

PR #2212, opened 2020-10-20
--context support to calicoctl

frozenprocess:
Using https://github.com/projectcalico/libcalico-go/pull/1335 this PR adds `--context` argument to calicoctl.
## Todos
- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calicoctl/commands/delete.go
+++ b/calicoctl/commands/delete.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/docopt/docopt-go"
@@ -30,7 +31,7 @@ func Delete(args []string) error {
 	doc := constants.DatastoreIntro + `Usage:
   <BINARY_NAME> delete ( (<KIND> [<NAME>...]) |
                    --filename=<FILE> [--recursive] [--skip-empty] )
-                   [--skip-not-exists] [--config=<CONFIG>] [--namespace=<NS>]
+                   [--skip-not-exists] [--config=<CONFIG>] [--namespace=<NS>] [--context=<context>]
 
 Examples:
   # Delete a policy using the type and name specified in policy.yaml.
@@ -60,6 +61,7 @@ Options:
                             Only applicable to NetworkPolicy and WorkloadEndpoint.
                             Only applicable to NetworkPolicy, NetworkSet, and WorkloadEndpoint.
                             Uses the default namespace if not specified.
+  --context=<context>       The name of the kubeconfig context to use.
 
 Description:
   The delete command is used to delete a set of resources by filename or stdin,
@@ -110,6 +112,9 @@ Description:
 	}
 	if len(parsedArgs) == 0 {
 		return nil
+	}
+	if context := parsedArgs["--context"]; context != nil {
+		os.Setenv("K8S_CURRENT_CONTEXT", context.(string))
 	}
 
 	results := common.ExecuteConfigCommand(parsedArgs, common.ActionDelete)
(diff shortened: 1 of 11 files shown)


frozenprocess:
I was not sure if there is a way to pass arguments into libcalico [Client](https://github.com/projectcalico/calicoctl/blob/be82d923e77efb5c44f8489ba55ada67f3bf2e83/calicoctl/commands/clientmgr/client.go#L31) so I decided to convert `--context` value into an environment variable.

@caseydavenport @robbrockbank

robbrockbank:
Hi Reza,

I would have included the field value in the CalicoAPIConfigSpec and then fill in the configOverrides.

see https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/k8s.go#L220

Also, you might need to change how we handle using the "default" namespace. We have this hardcoded in calicoctl here: https://github.com/projectcalico/calicoctl/blob/master/calicoctl/commands/common/resources.go#L361

In terms of where the new option lives.  I would put it at the same level as `--config`.

@caseydavenport :  are you ok with these suggestions?

calicoctl/commands/delete.go
```diff
@@ -16,7 +16,8 @@ package commands
 
 import (
 	"fmt"
-	"strings"
+  "strings"
```
caseydavenport, line 19: Looks like you'll need to run `gofmt` on this file

calicoctl/commands/get.go
```diff
@@ -61,7 +62,8 @@ Options:
   -a --all-namespaces          If present, list the requested object(s) across all namespaces.
   --export                     If present, returns the requested object(s) stripped of
                                cluster-specific information. This flag will be ignored
-			       if <NAME> is not specified.
```
caseydavenport, line 64: Indentation looks wrong here too? Although seems it was wrong in the other direction to begin with :D

caseydavenport:
> @caseydavenport : are you ok with these suggestions?

Yep, that all sounds right to me

calicoctl/commands/get.go
frozenprocess: Nice catch!

frozenprocess:
> I would have included the field value in the CalicoAPIConfigSpec and then fill in the configOverrides.
About this one

I've added a new string value `K8sCurrentContext` to `apiconfig.go` file in `kubeconfig` section for this specific override, since context is only going to affect `kubeconfig` scenarios I thought it would be best to do it in that section.

I talked to @robbrockbank about this one and he seemed to be satisfied with my explanation.


>In terms of where the new option lives. I would put it at the same level as --config.
My last commit supports `--context` regardless of its position. (more like kubectl)
```
calicoctl --context mycontext
calicoctl get --context mycontext node
calicoctl get node --context mycontext
```
@caseydavenport WDYT?

calicoctl/commands/delete.go
frozenprocess: I Love `gofmt`
Thank you!

frozenprocess:
Added test cases to check `--context` in multiple scenarios.

Extended `CalicoctlMayFail` function in `tests/fv/utils/calicoctl.go` to support multicluster. (`KUBECONFIG` got a lower priority than`K8S_API_ENDPOINT` that is why this change was needed.)

frozenprocess:
I've updated my base commit to #2223.

caseydavenport (approved):
@frozenprocess this looks good to me! I think you need to run `gofmt` though, there are one or two places the spacing looks a bit off.

PR #2223, opened 2020-11-12
comply with the go mod versioning constrains

svyotov:
## Module contains a go.mod file, so module path should be github.com/projectcalico/calicoctl/v3

fixes: #2126 

Go mod has as a requirement to put the code in directories named based on the release of the tool.
This project does not seem to support  or intend to support multiple versions.  As a result a simple solution of the `go mod` versioning and vendoring issues is to set the project versions as is in the root of this GitHub repository to v3: `github.com/projectcalico/calicoctl/v3`.

The possible solutions are:
1. Keep it as is and force users to use hashes to vendor (error prone and not user friendly) - IMO Bad
2.  Remove go.mod and revert to another vendoring system (dep, glide, ..) - IMO Bad
3. Refactor all the code under a new directory - Unnecessary based on my understanding of this projects goals/targets
4. The second best solution after the re-factoring, to set the version of the root module to v3.

Example of other projects that I have seen do this: https://github.com/go-git/go-git/blob/master/go.mod

## Todos
(not applicable - no new features or special documentation/testing/release note requirements as far as I can see)
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

Comply with the `go mod` module naming requirements for projects with version >= v2.x.x

```release-note
None required
```


caseydavenport:
@svyotov I'm not sure I understand the motivation for this. `calicoctl` is not a library, i.e., not intended to be imported into other projects.

> "If an old package and a new package have the same import path, the new package must be backwards compatible with the old package."

In other words, we are not trying to make any claims of compatibility between versions of the golang functions / structs / etc in this repository, so I don't think the motivation for that "rule" applies to us here, and would be promising something to users that is just not true.

Am I missing something?

svyotov:
> @svyotov I'm not sure I understand the motivation for this. `calicoctl` is not a library, i.e., not intended to be imported into other projects.
For better or worse, calicoctl has been imported by other tools as a library. It contains a lot of code that people decide to use (example https://github.com/projectcalico/calicoctl/issues/2126). In my case another library import this, and since they are not planing to drop it as that would be a considerate amount of work for them, I was thinking the second best is to "improve" the versioning here with this PR.

Is there any reason to no comply with the way `go mod` works, since it is used as a dependency vendoring tool for this project, or do you think the changes proposed will break something?

caseydavenport:
@svyotov well, I'm not aware of anything within the Calico project that imports calicoctl, and we nominally already don't expect other users to be doing this, so I think I'm OK merging this.

It does mean that anyone to _does_ happen to be importing this will have to adjust their import paths, but that's probably OK. If we get a ton of pushback post-merge we can revert (from a look through the list provided in the linked issue, I don't think there really are any. There is one really old fork of the calico-containers repo that hasn't been updated in 4 years). 

Thanks for pushing on this!

PR #2243, opened 2020-12-31
calicoctl get -A option support 

frozenprocess:
This PR changes `-a` option to `-A` in docs it also adds backward compatibility for `-a`.

```release-note
Introduces new `-A` option to calicoctl and deprecates the earlier `-a` flag, which will be removed in a subsequent release.
```
@caseydavenport


caseydavenport:
LGTM - worth a test that verifies both options work as expected.

frozenprocess:
A compilation test file?

caseydavenport:
I'm just thinking a test that calls `calicoctl get -A` and `calicoctl get -a` and verifies the output is the same for both.

caseydavenport:
@frozenprocess looks like you need a rebase on this guy

PR #2247, opened 2021-01-19
Additional IPAM sub commands

caseydavenport:
## Description

- Add's "ipam check" command which does a consistency check against Calico's IPAM data, writes a report.
- Enhance "ipam release" to be able to release inconsistent addresses from the generated report.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add calicoctl ipam check command to check IPAM consistency
```
```release-note
Allow releasing of IP addresses based on output report from calicoctl ipam check command.
```

### DIFF
--- a/calicoctl/commands/ipam/check.go
+++ b/calicoctl/commands/ipam/check.go
@@ -44,12 +44,13 @@ import (
 )
 
 // IPAM takes keyword with an IP address then calls the subcommands.
-func Check(args []string) error {
+func Check(args []string, version string) error {
 	doc := constants.DatastoreIntro + `Usage:
-  <BINARY_NAME> ipam check [--config=<CONFIG>] [--show-all-ips] [--show-problem-ips]
+  <BINARY_NAME> ipam check [--config=<CONFIG>] [--show-all-ips] [--show-problem-ips] [-o <FILE>]
 
 Options:
   -h --help                Show this screen.
+  -o --output=<FILE>       Path to output report file. Default: ipam-check-report.json
   -c --config=<CONFIG>     Path to the file containing connection configuration in
                            YAML or JSON format.
                            [default: ` + constants.DefaultConfigPath + `]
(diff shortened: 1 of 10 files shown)


calicoctl/commands/ipam/check.go
```diff
@@ -0,0 +1,576 @@
...
+func (c *IPAMChecker) recordAllocation(b *model.AllocationBlock, ord int) {
+	ip := b.OrdinalToIP(ord).String()
+	alloc := Allocation{IP: ip, Block: b, Ordinal: ord}
+
+	node := ""
+	blockAffinity := ""
+	if b.Affinity != nil {
+		affinity := *b.Affinity
+		if strings.HasPrefix(affinity, "host:") {
+			node = affinity[5:]
```
mgleung, line 412: If I'm understanding this correctly, this will include the `:` character in the node name. Is it intended for node to be `host:` here?

calicoctl/commands/ipam/check.go
```diff
@@ -0,0 +1,576 @@
...
+				numWEPIPs++
+			}
+		}
+		fmt.Printf("Found %d workload IPs.\n", numWEPIPs)
+		fmt.Printf("Workloads and nodes are using %d IPs.\n", len(c.inUseIPs))
+		fmt.Println()
+	}
+
+	{
+		const numNodesToPrint = 20
```
mgleung, line 266: Should we make this a configurable value?

calicoctl/commands/ipam/check.go
```diff
@@ -0,0 +1,576 @@
...
+						for _, owner := range owners {
+							fmt.Printf("  %s in use by %v is not in any active IP pool.\n", ip, owner.FriendlyName)
+						}
+					}
+					nonCalicoIPs = append(nonCalicoIPs, ip)
+					continue
+				}
+				if c.showProblemIPs {
+					for _, owner := range owners {
+						fmt.Printf("  %s in use by %v and in active IPAM pool but has no IPAM allocation.\n", ip, owner.FriendlyName)
```
mgleung, line 339: Should we also write these into the report if "show the problem IPs" option is set?

calicoctl/commands/ipam/check.go
caseydavenport: eh, I think it's not super important.

calicoctl/commands/ipam/check.go
caseydavenport: Yes, I think that we probably should. Although we can't actually do anything about these addresses if we find them - we just need to instruct the user to delete whatever thing is using the IP, most likely.

caseydavenport: I think this should be done in a follow-up, though. The main purpose of this is to get the initial "check and release" workflow functioning, which this isn't needed for.

mgleung (approved):
LGTM

PR #2253, opened 2021-02-02
Add tests for projectcalico/libcalico-go#1361

electricjesus:
- `make fv` now starts two apiservers (:8080, :8082) with two nodes (node4, node8 respectively)
- test scenario described in projectcalico/calico#4312
	- e.g. `KUBECONFIG=~/.kube/config:~/.kube/kubeconfig_2 calicoctl version`
- fix LOCAL_BUILD: run `go mod vendor` after `go mod edit` so that the replace directive gets synched to `vendor/modules.txt` when using your local ../libcalico-go repo

## Todos
- [x] Tests
- [x] Documentation
- [X] Release note

## Release Note

```release-note
`calicoctl` now supports multiple kubeconfig files via KUBECONFIG env var, just as `kubectl` would:

Example  usage:

    export KUBECONFIG=~/.kube/config:~/.kube/kubeconfig_2
    
    calicoctl get node
    calicoctl --context config2 get node
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,6 @@ LOCAL_BUILD_DEP:=set-up-local-build
 EXTRA_DOCKER_ARGS+=-v $(CURDIR)/../libcalico-go:/go/src/github.com/projectcalico/libcalico-go:rw
 $(LOCAL_BUILD_DEP):
 	$(DOCKER_RUN) $(CALICO_BUILD) go mod edit -replace=github.com/projectcalico/libcalico-go=../libcalico-go
-	$(DOCKER_RUN) $(CALICO_BUILD) go mod vendor
 endif
 
 include Makefile.common
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@ GO_BUILD_VER=v0.49
 
 SEMAPHORE_PROJECT_ID?=$(SEMAPHORE_CALICOCTL_PROJECT_ID)
 
+KUBE_MASTER_PORT?=8080
+KUBE_MOCK_NODE_MANIFEST?=mock-node.yaml
+
 ###############################################################################
 # Download and include Makefile.common
 #   Additions to EXTRA_DOCKER_ARGS need to happen before the include since
@@ -27,6 +30,7 @@ LOCAL_BUILD_DEP:=set-up-local-build
 EXTRA_DOCKER_ARGS+=-v $(CURDIR)/../libcalico-go:/go/src/github.com/projectcalico/libcalico-go:rw
 $(LOCAL_BUILD_DEP):
 	$(DOCKER_RUN) $(CALICO_BUILD) go mod edit -replace=github.com/projectcalico/libcalico-go=../libcalico-go
+	$(DOCKER_RUN) $(CALICO_BUILD) go mod vendor
 endif
 
 include Makefile.common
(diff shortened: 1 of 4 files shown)


electricjesus:
Test failures are expected for now until projectcalico/libcalico-go#1361 is merged

Makefile
```diff
@@ -3,6 +3,9 @@ GO_BUILD_VER=v0.49
 
 SEMAPHORE_PROJECT_ID?=$(SEMAPHORE_CALICOCTL_PROJECT_ID)
 
+KUBE_MASTER_PORT?=8080
```
fasaxc, line 6: Following Kubernetes lead, we're trying to move away from using terms like master/slave, suggest swapping this out for `KUBE_API_PORT`

Makefile
```diff
@@ -27,6 +30,7 @@ LOCAL_BUILD_DEP:=set-up-local-build
 EXTRA_DOCKER_ARGS+=-v $(CURDIR)/../libcalico-go:/go/src/github.com/projectcalico/libcalico-go:rw
 $(LOCAL_BUILD_DEP):
 	$(DOCKER_RUN) $(CALICO_BUILD) go mod edit -replace=github.com/projectcalico/libcalico-go=../libcalico-go
+	$(DOCKER_RUN) $(CALICO_BUILD) go mod vendor
```
fasaxc, line 33: I'm wary of `go mod vendor`, having a stale vendor directory around can lead to build breakage later IME; let's chat about it; my understanding was that replacing should be enough.

Makefile
```diff
@@ -216,10 +220,12 @@ ut: $(LOCAL_BUILD_DEP) bin/calicoctl-linux-amd64
 ## Run the tests in a container. Useful for CI, Mac dev.
 fv: $(LOCAL_BUILD_DEP) bin/calicoctl-linux-amd64
 	$(MAKE) run-etcd-host
-	$(MAKE) run-kubernetes-master
+	$(MAKE) run-kubernetes-master KUBE_MASTER_PORT=8080 KUBE_MOCK_NODE_MANIFEST=mock-node.yaml
```
fasaxc, line 223: Worth a comment `# We start two API servers in order to test multiple kubeconfig support` ?

tests/st/manifests/mock-node-second.yaml
```diff
@@ -0,0 +1,14 @@
+# This is a Node used by the k8s ST tests.  The migration tests
```
fasaxc, line 1: Comment incorrect for this new file.

Makefile
```diff
@@ -280,55 +286,58 @@ stop-etcd:
...
-                        --master=127.0.0.1:8080 \
+                        --master=127.0.0.1:${KUBE_MASTER_PORT} \
                         --min-resync-period=3m \
                         --allocate-node-cidrs=true \
                         --cluster-cidr=10.10.0.0/16 \
                         --v=5
 
 	# Create a Node in the API for the tests to use.
-	docker run \
+	while ! docker run \
```
fasaxc, line 317: Presumably you saw this fail?  Bit odd after we successfully get the nodes above?

fasaxc (changes requested):
Looking good, just a few minor comments from me.  Think your simple test does just enough to make sure this won't get regressed :+1:

Makefile
electricjesus: Yes I think because as this node is being made and requested on the second apiserver, the second apiserver hasn't yet registered at this point so it fails with a ResourceNotFound

tests/st/manifests/mock-node-second.yaml
electricjesus: Should I update the other one as well? it's also incorrect

electricjesus: Oh wait, no.. it's still true (applies to migration tests in ST)

Makefile
electricjesus: I commented on the rest of the steps too (incoming commit in a while) as the block now seems to be large enough for comments to help with readability

fasaxc (approved):

PR #2255, opened 2021-02-03
Check system update

frozenprocess:
Update 

Issue reported here https://github.com/projectcalico/calico/issues/4288

## TODOS
[x] add bootconfig file to the mix of checks
[x] add kernel variables of required modules as a backup
[x] Add option to get the bootconfig override from user input

## Maybe
- Add `compress/gzip` in order to read `/boot/config.tar.gz` included in some linux flavors
- a windows checkfile system
```release-note
Improve calicoctl checksystem's ability to discover kernel dependencies
```

### DIFF
--- a/calicoctl/commands/node/checksystem.go
+++ b/calicoctl/commands/node/checksystem.go
@@ -34,17 +34,48 @@ import (
 // namespaces and veth pairs.
 const minKernelVersion = "2.6.24"
 
-// Required kernel modules to run Calico
-var requiredModules = []string{"ip_set", "ip_tables", "ip6_tables", "ipt_REJECT", "ipt_rpfilter", "ipt_set", "nf_conntrack_netlink",
-	"xt_addrtype", "xt_conntrack", "xt_icmp", "xt_icmp6", "xt_ipvs", "xt_mark", "xt_multiport", "xt_rpfilter", "xt_set", "xt_u32"}
+// Required kernel modules to run Calico are saved in a two dimentional map variable.
+// Keys are used to search in `lsmod` results or `modules.dep` and `modules.builtin` files.
+// Values are used to search the the `kernel config` or `ip_tables_matches` file.
+var requiredModules = map[string]string{
+	"ip_set":               "CONFIG_IP_SET",
+	"ip6_tables":           "CONFIG_IP6_NF_IPTABLES",
+	"ip_tables":            "CONFIG_IP_NF_IPTABLES",
+	"ipt_ipvs":             "CONFIG_NETFILTER_XT_MATCH_IPVS",
+	"vfio-pci":             "CONFIG_VFIO",
+	"xt_bpf":               "CONFIG_BPF",
+	"ipt_REJECT":           "CONFIG_NFT_REJECT",
+	"ipt_rpfilter":         "CONFIG_IP_NF_MATCH_RPFILTER",
+	"xt_rpfilter":          "CONFIG_IP_NF_MATCH_RPFILTER",
+	"ipt_set":              "CONFIG_NET_EMATCH_IPSET",
+	"nf_conntrack_netlink": "CONFIG_NF_CT_NETLINK",
+	"xt_addrtype":          "CONFIG_NETFILTER_XT_MATCH_ADDRTYPE",
+	"xt_conntrack":         "CONFIG_NETFILTER_XT_MATCH_CONNTRACK",
+	"xt_icmp":              "icmp",
+	"xt_icmp6":             "icmp",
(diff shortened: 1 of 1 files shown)


calicoctl/commands/node/checksystem.go
```diff
@@ -133,6 +138,12 @@ func checkKernelModules() error {
 	// File path to Builtin kernel modules
 	modulesBuiltinPath := fmt.Sprintf("/lib/modules/%s/modules.builtin", kernelVersionStr)
 
+	// File path to module configs in boot time
+	modulesBootPath := findBootFile(kernelVersionStr)
+
+	// File path for loaded iptables modules
+	modulesLoadedIPtables := "/proc/net/ip_tables_matches"
```
frozenprocess, line 145: I rather use another method but this was the only way that I could've think of iptables modules `"xt_icmp","xt_icmp6"`.

frozenprocess:
@caseydavenport I think this is done for now.
Please let me know if you got any suggestions.

calicoctl/commands/node/checksystem.go
```diff
@@ -35,16 +35,27 @@ import (
...
 
 // Checksystem checks host system for compatible versions
 func Checksystem(args []string) error {
 	doc := `Usage:
-  <BINARY_NAME> node checksystem
+  <BINARY_NAME> node checksystem [--filename=<FILENAME>]
 
 Options:
   -h --help                 Show this screen.
+  -f --filename=<filename>	Override kernel bootfile location, checksystem
```
caseydavenport, line 56: I think this argument isn't super clear to the user - what format is the file supposed to be? Where are the default locations that are checked? etc. 

I think this is referring to the kernel's build configuration file, rather than bootfile? I could be wrong though.

caseydavenport: Probably worth calling the argument `--kernel-config` or something rather than `--filename` since it's more exact.

caseydavenport (changes requested):

calicoctl/commands/node/checksystem.go
```diff
@@ -196,7 +219,7 @@ func checkKernelModules() error {
 // as an argument exists for the provided kernelVersion
 func checkModule(filename, module, kernelVersion string) error {
 
-	regex, err := regexp.Compile(fmt.Sprintf("\\/%s.ko", module))
+	regex, err := regexp.Compile(fmt.Sprintf("(?:\\/|^)%s(?:\\.ko|=.|$)", module))
```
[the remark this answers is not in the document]
frozenprocess: I implemented both suggested ways and the alternative one seemed a bit more clear and precise to me.

frozenprocess (commented):
@caseydavenport I think I implemented all the changes that you suggested, let me know if there is room for improvement!

Thanks!

calicoctl/commands/node/checksystem.go
```diff
@@ -34,17 +34,48 @@ import (
...
+	"xt_u32":               "CONFIG_NETFILTER_XT_MATCH_U32"}
+
+// Variable to override bootfile location.
+var overrideBootFile = ""
 
 // Checksystem checks host system for compatible versions
 func Checksystem(args []string) error {
 	doc := `Usage:
-  <BINARY_NAME> node checksystem
+	<BINARY_NAME> node checksystem [--kernel-config=<kernel-config>]
```
caseydavenport, line 67: Did you mean to indent this? I think it should probably go back to the previous indentation to match other commands.

calicoctl/commands/node/checksystem.go
```diff
@@ -34,17 +34,48 @@ import (
...
-  -h --help                 Show this screen.
+  -h --help                             Show this screen.
+  -f --kernel-config=<kernel-config>    Override the Kernel config file location.
+                                        Expected format is plain text.
+                                        default search locations:
+                                          "/usr/src/linux/.config",
+                                          "/boot/config-kernelVersion,
+                                          "/usr/src/linux-kernelVersion/.config",
+                                          "/usr/src/linux-headers-kernelVersion/.config",
+                                          "/lib/modules/kernelVersion/build/.config"
```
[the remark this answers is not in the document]
caseydavenport: Looks good

caseydavenport (approved):
One minor comment about indentation but then LGTM

calicoctl/commands/node/checksystem.go
frozenprocess: oops 😅 seems like instead of shift+tab I just pushed tab.

caseydavenport (approved):

PR #2313, opened 2021-04-20
Build all architectures for calicoctl

mgleung:
## Description

Ensures that all of the architectures that we support are built and uploaded on each release for the calicoctl utility. Fixes https://github.com/projectcalico/calicoctl/issues/2269

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,9 @@ BUILD_IMAGE?=calico/ctl
 PUSH_IMAGES?=$(BUILD_IMAGE) quay.io/calico/ctl
 RELEASE_IMAGES?=
 
+# Remove any excluded architectures since for calicoctl we want to build everything.
+EXCLUDEARCH?=
+
 include Makefile.common
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


caseydavenport (approved):

Brian-McM (approved):

Makefile
```diff
@@ -72,6 +72,10 @@ update-pins: update-libcalico-pin
 ###############################################################################
 # Building the binary
 ###############################################################################
+
+# Reassign VALIDARCHES since for calicoctl we do not want to exclude any builds.
+VALIDARCHES=$(patsubst Dockerfile.%,%,$(wildcard Dockerfile.*))
```
Brian-McM, line 77: This needs to be above the include for the common makefile otherwise the targets in the common makefile won't be created for all the valid arches

Brian-McM (changes requested):
Could you please move VALIDARCHES above the include for the common makefile

Brian-McM:
So my other comment is slightly wrong, what you need to do is set `EXCLUDEARCH` to empty before the common makefile include:
```
ARCHES ?= $(patsubst docker-image/Dockerfile.%,%,$(wildcard docker-image/Dockerfile.*))

# Some repositories keep their Dockerfile(s) in the root directory instead of in
# the 'docker-image' subdir. Make sure ARCHES gets filled in either way.
ifeq ($(ARCHES),)
    ARCHES=$(patsubst Dockerfile.%,%,$(wildcard Dockerfile.*))
endif

# list of arches *not* to build when doing *-all
#    until s390x works correctly
EXCLUDEARCH ?= s390x
VALIDARCHES = $(filter-out $(EXCLUDEARCH),$(ARCHES))
```
^^ that's in the common Makefile. The "sub" targets that are created on the include require the VALIDARCHES be set before the include.

Brian-McM (approved):

PR #2332, opened 2021-05-22
change logs of 'node status' command for easy debug

zgfh:
change logs of 'node status' command for easy debug

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
I'm worried about changing the output of this command, as there may be existing tools relying on this output.

Plus, there are several processes that this _could_ check (although it's only checking felix today). Not sure we want to expose that implementation detail.

Maybe instead of changing the output, we could add a debug log with more detailed information? WDYT?

zgfh:
thanks for explaining, i will close this

PR #2353, opened 2021-07-13
Add client/cluster version mismatch verification

coutinhop:
## Description

Make `calicoctl` verify if the cluster version is the same as the client version before running. Add `--allow-version-mismatch` command line arg to override this verification if needed. The verification is also bypassed if the command being run is `datastore migrate import`, since the ClusterInformation will not be set in this case.

Add go utility app calico_version_helper, which is used to set the calico version in the ClusterInformation for test environments.

Add st coverage for the feature, and make multiple changes to st and fv tests to set the version so that the tests continue performing correctly.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Added calicoctl checks for version mismatch between itself and Calico in the cluster, with optional bypass argument '--allow-version-mismatch'.
```

### DIFF
--- a/calicoctl/calicoctl.go
+++ b/calicoctl/calicoctl.go
@@ -50,10 +50,11 @@ func main() {
     datastore    Calico datastore management.
 
 Options:
-  -h --help               Show this screen.
-  -l --log-level=<level>  Set the log level (one of panic, fatal, error,
-                          warn, info, debug) [default: panic]
-  --context=<context>	  The name of the kubeconfig context to use.
+  -h --help                 Show this screen.
+  -l --log-level=<level>    Set the log level (one of panic, fatal, error,
+                            warn, info, debug) [default: panic]
+  --context=<context>       The name of the kubeconfig context to use.
+  --allow-version-mismatch  Allow client and cluster versions mismatch.
 
 Description:
   The %s is used to manage Calico network and security
@@ -75,7 +76,7 @@ Description:
 	if err != nil {
 		if _, ok := err.(*docopt.UserError); ok {
 			// the user gave us bad input
-			fmt.Printf("Invalid option: '%s'. Use flag '--help' to read about a specific subcommand.\n", strings.Join(os.Args[1:], " "))
+			fmt.Printf("Invalid or incomplete arguments: '%s'. Use flag '--help' to read about a specific subcommand.\n", strings.Join(os.Args[1:], " "))
 		}
 		os.Exit(1)
 	}
@@ -99,6 +100,20 @@ Description:
 	if arguments["<command>"] != nil {
 		command := arguments["<command>"].(string)
 		args := append([]string{command}, arguments["<args>"].([]string)...)
+
+		// A "datastore migrate import" command should skip version mismatch checking,
+		// since the datastore will not have the cluster version set
+		isImport := args[0] == "datastore" && args[1] == "migrate" && args[2] == "import"
+
+		// Check for client/cluster version mismatch. If a mismatch occurs, check for
+		// --allow-version-mismatch arg to override/fail.
+		allowMismatch, ok := arguments["--allow-version-mismatch"].(bool)
+		if (!ok || !allowMismatch) && !isImport {
+			if err = commands.VersionMismatch(args); err != nil {
+				fmt.Fprintf(os.Stderr, "%s\n Use --allow-version-mismatch to override.", err)
+				os.Exit(1)
+			}
+		}
 
 		var err error
 
--- a/calicoctl/commands/version.go
+++ b/calicoctl/commands/version.go
@@ -130,3 +130,102 @@ Description:
 
 	return err
 }
+
+func VersionMismatch(args []string) error {
+	// We need to "look ahead" to see if config or context have been passed in the args
+	doc := `Usage:
+  <BINARY_NAME> [options] [<args>...]
+
+Options:
+  -h --help                 Show this screen.
+  -c --config=<config>      Path to the file containing connection
+                            configuration in YAML or JSON format.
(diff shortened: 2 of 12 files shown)


calicoctl/calicoctl.go
```diff
@@ -100,6 +101,16 @@ Description:
 		command := arguments["<command>"].(string)
 		args := append([]string{command}, arguments["<args>"].([]string)...)
 
+		// Check for client/cluster version mismatch. If a mismatch occurs, check for
+		// --allow-version-mismatch arg to override/fail.
+		if err = commands.VersionMismatch(args); err != nil {
```
mgleung, line 106: Instead of always running this, it might be better to run it only when `--allow-version-mismatch` is not passed in

calicoctl/commands/version.go
```diff
@@ -130,3 +130,38 @@ Description:
...
+	// We need to "look ahead" to see if --config has been passed in the args
+	for i := range args {
+		if args[i] == "--config" && i < len(args)-1 {
+			config = args[i+1]
+		}
+	}
+
+	client, err := clientmgr.NewClient(config)
+	if err != nil {
+		return fmt.Errorf("Unable to verify version mismatch: %w", err)
```
mgleung, line 147: you may want to give these error messages something a little different so you can tell what happened apart in case you're debugging based on the error message

calicoctl/calicoctl.go
coutinhop: Right, good point! Will do, thanks!

calicoctl/commands/version.go
coutinhop: Will do!

tests/fv/helper/calico_version_helper.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Get the backend client for updating cluster info and migrating IPAM.
+	client, err := clientv3.New(*cfg)
+	if err != nil {
+		fmt.Printf("Could not create client: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Import Calico CRDs
+	err = migrate.ImportCRDs(cfg)
```
mgleung, line 61: Do we need to do this for etcd clusters as well? I figured that the CRDs would only be needed for KDD clusters.

mgleung: I **think** we could also apply just the `ClusterInformation` CRD we need in the K8s test cluster we're using too via `kubectl` (or some other method of hitting the K8s API)? Then we could avoid needing a blanket import as well.

calicoctl/commands/version.go
```diff
@@ -130,3 +130,79 @@ Description:
...
 	return err
 }
+
+func GetConfigAndContext(args []string) (string, string) {
+	// Possible arg formats:
+	// -cVALUE, -c VALUE, --config VALUE, --config=VALUE,
+	// --context VALUE, --context=VALUE
+	var config, context string
+
+	for i, arg := range args {
```
mgleung, line 140: You might've tried this, but I'm wondering instead of doing all the parsing manually which opens us up to a lot of edge cases if we could add a fake docopt string that displays all the options across all the commands (to avoid errors from unknown options) so we could catch the `--config` option. Of course this also has the downside of needing to be updated if we add new calicoctl commands with different options but it does give us a fairly effortless way of getting the token parsing for free (without pulling relevant code from docopt). WDYT?

calicoctl/calicoctl.go
```diff
@@ -100,6 +101,20 @@ Description:
 		command := arguments["<command>"].(string)
 		args := append([]string{command}, arguments["<args>"].([]string)...)
 
+		// A "datastore migrate import" command should skip version mismatch checking,
+		// since the datastore will not have the cluster version set
+		isImport := strings.HasPrefix(strings.Join(args, " "), "datastore migrate import")
```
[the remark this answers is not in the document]
coutinhop: hm... you're right, let me try that, thanks!

tests/fv/helper/calico_version_helper.go
coutinhop: hm... I just figured we already had a way of updating the CRDs when they change, and since "utility" app is only used on tests, maybe this was fine... I'll look into only importing the `ClusterInformation` CRD

calicoctl/commands/version.go
coutinhop: hm... tbh it didn't cross my mind to have every single argument possible in the docstring, I'm gonna try that. thanks!

calicoctl/calicoctl.go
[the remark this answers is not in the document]
coutinhop: Sorry for returning to this, but I think it will work that way as is. `args` is already created by joining `arguments["<command>"]` and `arguments["<args>"]` (which will contain the subcommand). So this will already have removed "top-level" calicoctl options (such as `-l debug`), and I don't think it's possible to have options between the command and subcommands due to the docopts parsing of the `datastore` command...

tests/fv/helper/calico_version_helper.go
coutinhop: as we sync'ed up offline, I'll import the CRDs only on kdd via the Makefile for the test environments, so we won't need to worry about making `importCRDs()` public in `migrate`, or about importing them in the go code. Thanks for helping me out with this!

calicoctl/calicoctl.go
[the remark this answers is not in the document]
coutinhop: Made the changes per our offline conversation, thanks for the improvement suggestions!

mgleung (approved):
LGTM

mgleung (approved):
LGTM

PR #2370, opened 2021-07-28
Skip version mismatch check when ClusterInformation does not exist

neiljerram:
This is needed as a follow-up to #2353 , as many of Tigera's internal tests do not create any ClusterInformation.

### DIFF
--- a/calicoctl/commands/version.go
+++ b/calicoctl/commands/version.go
@@ -211,6 +211,10 @@ Description:
 
 	ci, err := client.ClusterInformation().Get(ctx, "default", options.GetOptions{})
 	if err != nil {
+		if strings.Contains(err.Error(), "does not exist") {
+			// ClusterInformation does not exist, so skip version check.
+			return nil
+		}
 		return fmt.Errorf("Unable to get Cluster Information to verify version mismatch: %w", err)
 	}
 
--- a/calicoctl/commands/version.go
+++ b/calicoctl/commands/version.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


calicoctl/commands/version.go
```diff
@@ -211,6 +211,10 @@ Description:
 
 	ci, err := client.ClusterInformation().Get(ctx, "default", options.GetOptions{})
 	if err != nil {
+		if strings.Contains(err.Error(), "does not exist") {
```
fasaxc, line 214: Think we return a dedicated, reliable error type here; it should be `ErrorResourceDoesNotExist` from our errors package

neiljerram: Is there an explicit contract somewhere for this, though?  If it's not robust, it doesn't feel obviously better than checking for a string.

song-jiang: Seems we are checking NotExist error like this https://github.com/projectcalico/node/blob/master/pkg/lifecycle/startup/startup.go#L433

song-jiang: Suggest printing out information before return nil so that user knows cluster information does not exist.

neiljerram: I guess I can do that.  My concern is: there are at least 7 levels of `interface` usage between `client.ClusterInformation().Get` and what actually happens, and **none** of those interfaces says "if the resource is not found, the error will be `ErrorResourceDoesNotExist`.  So assuming that feels no less fragile to me than a hacky string check, and I do at least know that the string check will fix our CI because I've seen the string "does not exist" in the CI output.

(Aside: this feels like a structural weakness in golang.)

song-jiang (approved):
LGTM. Thanks @neiljerram

PR #2371, opened 2021-07-28
Skip version mismatch check when CalicoVersion not specified in cluster

neiljerram:
A further follow-up to #2353


coutinhop (approved):
LGTM, thanks!

PR #2390, opened 2021-07-30
 Refactor version mismatch checking to not need argument parsing

coutinhop:
## Description

Refactor version mismatch checking to run after argument parsing in the
applicable commands, dismissing the need for its own fragile argument parsing.

Add --allow-version-mismatch arg to every command (no-op when not applicable).

Adjust version mismatch st test accordingly.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Fixed usage message output when invalid arguments are passed to calicoctl.
```

### DIFF
--- a/calicoctl/commands/common/version_mismatch.go
+++ b/calicoctl/commands/common/version_mismatch.go
@@ -1,0 +1,82 @@
+// Copyright (c) 2021 Tigera, Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/projectcalico/calicoctl/v3/calicoctl/commands/clientmgr"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/options"
+)
+
+var VERSION string
+
+func CheckVersionMismatch(configArg, allowMismatchArg interface{}) error {
+	if allowMismatch, _ := allowMismatchArg.(bool); allowMismatch {
+		log.Infof("Skip version mismatch checking due to '--allow-version-mismatch' argument")
+
+		return nil
+	}
+
+	cf, _ := configArg.(string)
+
+	client, err := clientmgr.NewClient(cf)
+	if err != nil {
+		// If we can't connect to the cluster, skip the check. Either we're running a command that
+		// doesn't need API access, in which case the check doesn't need to be run, or we'll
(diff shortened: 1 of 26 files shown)


mgleung (approved):

mgleung (changes requested):
On second thought, we may want to figure out what to do in the case where the version mismatch options haven't been updated, but the appropriate logic in the command itself has been updated since this will allow the version mismatch checks to be skipped.

coutinhop:
> On second thought, we may want to figure out what to do in the case where the version mismatch options haven't been updated, but the appropriate logic in the command itself has been updated since this will allow the version mismatch checks to be skipped.

I updated the PR, moving the version checking to the commands themselves, so now we don't need a separate argument parsing for it, which had shown to be fragile and buggy. Thanks, @mgleung !

calicoctl/commands/node/run.go
```diff
@@ -69,6 +69,7 @@ func Run(args []string) error {
                      [--no-default-ippools]
                      [--dryrun]
                      [--init-system]
+					 [--allow-version-mismatch]
```
mgleung, line 72: Looks like there's a spacing error here, probably need to change it from tabs to spaces?

calicoctl/commands/common/version_mismatch.go
```diff
@@ -0,0 +1,83 @@
...
+		return
+	}
+
+	clusterv = strings.Split(clusterv, "-")[0]
+
+	clientv := strings.Split(VERSION, "-")[0]
+
+	if clusterv != clientv {
+		fmt.Fprintf(os.Stderr, "Version mismatch.\nClient Version:   %s\nCluster Version:  %s\nUse --allow-version-mismatch to override.\n", VERSION, clusterv)
+		os.Exit(1)
```
mgleung, line 81: nit: I know that exiting here is less code, but I think it's a little less clear from a glance if you were to see the `CheckVersionMismatch` function called that you would know it would prevent the rest of the code from running. I think the more common way is to have it return an error and then have each command pass that error until it finally gets handled in `calicoctl.go` (where it'll exit). I think it makes it easier for anyone using the function to reuse it and leaves us options to handle the error in the future separately.

calicoctl/commands/node/run.go
coutinhop: Oh, I missed it in my editor, thanks for catching it, will fix!

calicoctl/commands/common/version_mismatch.go
coutinhop: Got it, yeah it makes sense... I'll make the changes, thanks!

calicoctl/commands/version.go
```diff
@@ -44,15 +42,16 @@ func init() {
...
-                        [default: ` + constants.DefaultConfigPath + `]
-     --poll=<POLL>      Poll for changes to the cluster information at a frequency specified using POLL duration
-                        (e.g. 1s, 10m, 2h etc.). A value of 0 (the default) disables polling.
+  -h --help                 Show this screen.
+  -c --config=<CONFIG>      Path to the file containing connection configuration in
+                            YAML or JSON format.
+                            [default: ` + constants.DefaultConfigPath + `]
+     --poll=<POLL>          Poll for changes to the cluster information at a frequency specified using POLL duration
+                            (e.g. 1s, 10m, 2h etc.). A value of 0 (the default) disables polling.
+  --allow-version-mismatch  Allow client and cluster versions mismatch.
```
mgleung, line 54: nit: spacing doesn't align with `--poll` above

coutinhop: Fixed, thanks for catching it!

mgleung (approved):
LGTM

PR #2415, opened 2021-09-15
Multiarch image adjustments

frozenprocess:
Fix: This PR updates all non `amd64` dockerfiles for a multiarch release.

```release-note
None required
```

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -15,13 +15,14 @@ ARG QEMU_IMAGE=calico/go-build:v0.55
 FROM ${QEMU_IMAGE} as qemu
 
 FROM --platform=linux/arm64 registry.access.redhat.com/ubi8/ubi-minimal:8.2 as ubi
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be the first RUN command in this file!
 COPY --from=qemu /usr/bin/qemu-aarch64-static /usr/bin/
 
 RUN mkdir /licenses
 COPY LICENSE /licenses
 
 FROM scratch
-COPY --from=qemu /usr/bin/qemu-aarch64-static /usr/bin/
 
 ARG GIT_VERSION=unknown
 
--- a/Dockerfile.armv7
+++ b/Dockerfile.armv7
@@ -1,4 +1,11 @@
+ARG QEMU_IMAGE
+FROM --platform=linux/arm/v7 ${QEMU_IMAGE} as qemu
+
 FROM arm32v7/alpine:3.12
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be the first RUN command in this file!
+COPY --from=qemu /usr/bin/qemu-arm-static /usr/bin/
+
 MAINTAINER Marc Crébassa <aalaesar@gmail.com>
 
 RUN apk --no-cache upgrade apk-tools
@@ -7,6 +14,8 @@ ADD bin/calicoctl-linux-armv7 /calicoctl
 
 ENV CALICO_CTL_CONTAINER=TRUE
 ENV PATH=$PATH:/
+# Delete qemu binaries
+RUN rm /usr/bin/qemu-*
 
 WORKDIR /root
 ENTRYPOINT ["/calicoctl"]
--- a/Dockerfile.ppc64le
+++ b/Dockerfile.ppc64le
@@ -1,5 +1,11 @@
+ARG QEMU_IMAGE
+FROM ${QEMU_IMAGE} as qemu
+
 FROM ppc64le/alpine:3.10
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be the first RUN command in this file!
 MAINTAINER Tom Denham <tom@projectcalico.org>
+COPY --from=qemu /usr/bin/qemu-ppc64le-static /usr/bin/
 
 RUN apk --no-cache upgrade apk-tools
 
@@ -7,6 +13,9 @@ ADD bin/calicoctl-linux-ppc64le /calicoctl
 
 ENV CALICO_CTL_CONTAINER=TRUE
 ENV PATH=$PATH:/
+
+# Delete qemu binaries
+RUN rm /usr/bin/qemu-*
 
 WORKDIR /root
 ENTRYPOINT ["/calicoctl"]
--- a/Dockerfile.s390x
+++ b/Dockerfile.s390x
@@ -1,4 +1,12 @@
+ARG QEMU_IMAGE
+# Copying qemu binary files from amd64 container.
+# latest `go-build` containers got no candidate for `s390x` architecture.
+FROM --platform=linux/amd64 ${QEMU_IMAGE} as qemu
+
 FROM s390x/alpine:3.10
+# Enable non-native builds of this image on an amd64 hosts.
+# This must be the first RUN command in this file!
+COPY --from=qemu /usr/bin/qemu-s390x-static /usr/bin/
 MAINTAINER LoZ Open Source Ecosystem (https://www.ibm.com/developerworks/community/groups/community/lozopensource)
 
 RUN apk --no-cache upgrade apk-tools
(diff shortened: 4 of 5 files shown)


Dockerfile.arm64
```diff
@@ -35,6 +35,8 @@ LABEL name="Calico CLI tool" \
 
 COPY --from=ubi /licenses /licenses
 ADD bin/calicoctl-linux-arm64 /calicoctl
+# Overwrite qemu binary
+ADD LICENSE /usr/bin/qemu-aarch64-static
```
frozenprocess, line 39: At this point our `arm64` image doesn't have executables to invoke normal commands for removing `qemu-aarch64-static` binary, so I just overwrite the file with a text file.

Makefile
```diff
@@ -138,7 +140,7 @@ remote-deps: mod-download
 image: $(CALICOCTL_IMAGE)
 $(CALICOCTL_IMAGE): $(CTL_CONTAINER_CREATED)
 $(CTL_CONTAINER_CREATED): register Dockerfile.$(ARCH) bin/calicoctl-linux-$(ARCH)
-	docker build -t $(CALICOCTL_IMAGE):latest-$(ARCH) $(TARGET_PLATFORM) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --build-arg GIT_VERSION=$(GIT_VERSION) -f Dockerfile.$(ARCH) .
+	docker build --pull -t $(CALICOCTL_IMAGE):latest-$(ARCH) $(TARGET_PLATFORM) --build-arg QEMU_IMAGE=$(CALICO_BUILD) --build-arg GIT_VERSION=$(GIT_VERSION) -f Dockerfile.$(ARCH) .
```
frozenprocess, line 143: `--pull`  = Always attempt to pull a newer version of the image

Makefile
```diff
@@ -49,6 +49,8 @@ Makefile.common.$(MAKE_BRANCH):
 include Makefile.common
 
 ifeq ($(ARCH),arm64)
+# Forces ARM64 build image to be used in a crosscompilation run.
+CALICO_BUILD:=$(CALICO_BUILD)-$(ARCH)
```
frozenprocess, line 53: When docker pulls `go-build:v0.55` for `arm64` it tags it incorrectly and use it for all other architectures which causes problems.
In this line we make sure that image pulled by docker for `arm64` process is tagged correctly in local repo. `go-build:v0.55-arm64`

Dockerfile.arm64
```diff
@@ -11,8 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-ARG QEMU_IMAGE=calico/go-build:v0.55
-FROM ${QEMU_IMAGE} as qemu
+ARG QEMU_IMAGE
+FROM ${QEMU_IMAGE}-arm64 as qemu
```
frozenprocess, line 15: In order to tag the image with correct achitecture we are using `--platform` with the `docker build` command causing docker to pull the `arm64` image to the local repository and tag it as `v0.55` then docker uses this localy stored image to build all other architectures causing an issue in binary build step.

Dockerfile.s390x
```diff
@@ -1,4 +1,9 @@
-FROM s390x/alpine:3.10
+ARG QEMU_IMAGE
+FROM ${QEMU_IMAGE} as qemu
+
+FROM s390x/alpine:3.12 as base
```
neiljerram, line 4: Do we need "as base" here?  Please remove if not.

Dockerfile.ppc64le
```diff
@@ -1,4 +1,9 @@
-FROM ppc64le/alpine:3.10
+ARG QEMU_IMAGE
+FROM ${QEMU_IMAGE} as qemu
+
+FROM ppc64le/alpine:3.12 as base
```
neiljerram, line 4: Do we need "as base" here? Please remove if not.

Dockerfile.armv7
```diff
@@ -1,5 +1,10 @@
-FROM arm32v7/alpine:3.12
-MAINTAINER Marc Crébassa <aalaesar@gmail.com>
+ARG QEMU_IMAGE
+FROM ${QEMU_IMAGE} as qemu
+
+FROM arm32v7/alpine:3.12 as base
```
neiljerram, line 4: Do we need "as base" here? Please remove if not.

Dockerfile.armv7
```diff
@@ -1,5 +1,10 @@
...
+COPY --from=qemu /usr/bin/qemu-arm-static /usr/bin/
+
+MAINTAINER LoZ Open Source Ecosystem (https://www.ibm.com/developerworks/community/groups/community/lozopensource)
```
neiljerram, line 7: Why this change?  Probably better to have a separate PR for this, so that the reason for the change is clearly described.

Dockerfile.arm64
neiljerram: How exactly does that help?  Who is it that checks for the existence of /usr/bin/qemu-aarch64-static and then tries to run it?  With this solution, will they still try to run it, and then hit an error because it's just plain text?

Dockerfile.arm64
neiljerram: I'm sorry, I don't understand.  Why do we need to add "-arm64" here, but we don't do that for the other architectures?

> In order to tag the image

Which image do you mean?

> with correct architecture we are using --platform with the docker build command causing docker to pull the arm64 image

Which image do you mean here?

> to the local repository and tag it as v0.55

Huh?  It should only be calico-build that has the v0.55 tag, but I don't think you mean the calico-build image here, do you?

> then docker uses this locally stored image to build all other architectures causing an issue in binary build step.

neiljerram (changes requested):

Dockerfile.arm64
frozenprocess: > Why do we need to add "-arm64" here, but we don't do that for the other architectures?

It is because of the `--platform=linux/arm64/v8` pulls an `arm64` image and tags it as both `calico/go-build:v0.55` and `calico/go-build:v0.55-arm64`.


In `arm64` dockerfile `FROM SCRATCH` pulls an empty `amd64` layer in the build process which results in an incorrect architecture tag for the output image (aka `calico/ctl`).`--platform` prevents docker from tagging image (`calico/ctl`) architecture as `amd64`.

Here is an example from cni repo:
https://github.com/projectcalico/cni-plugin/blob/f6cd1b86daec16b04e095b04abf0af3bffb73a89/Dockerfile.arm64#L1
```
docker inspect calico/cni:v3.20.1-arm64 | egrep -i arch
        "Architecture": "amd64",
```

> Which image do you mean?

Output image aka `calico/ctl:vxxxxx-arm64`


> Which image do you mean here?

`calico/go-build:v0.55-arm64` will be pulled and stored as both `calico/go-build:v0.55` and `calico/go-build:v0.55-arm64`

> Huh? It should only be calico-build that has the v0.55 tag, but I don't think you mean the calico-build image here, do you?

No, I meant `calico/go-build:v0.55-arm64` will be saved as `calico/go-build:v0.55`
```
root@build-dummy:~/calicoctl# docker image ls
REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
root@build-dummy:~/calicoctl# docker pull --platform=linux/arm64/v8 calico/go-build:v0.55 &> /dev/null
root@build-dummy:~/calicoctl# docker image ls
REPOSITORY        TAG       IMAGE ID       CREATED      SIZE
calico/go-build   v0.55     633df77ba924   5 days ago   1.74GB
root@build-dummy:~/calicoctl# docker inspect 633df77ba924 | egrep -i arch
        "Architecture": "arm64",
```

Dockerfile.s390x
frozenprocess: We don't need it and It can be removed.

Dockerfile.armv7
frozenprocess: Since `arm32v7/alpine` is an `arm` based image and in it executable files are `arm` only we neeed the qemu binary to run the commands like the following one
https://github.com/projectcalico/calicoctl/blob/136af0e96783676c7faba39c71b1ebf390d04dd6/Dockerfile.armv7#L9

caseydavenport (approved):
Alright this has been broken long enough. Up up and away!

PR #2458, opened 2021-11-19
Remove TTY flag from docker exec

algamaes:
## Description
In recent changes for the "make test" command, the flag -ti was added to the docker exec command. However, this caused that running the test in some pipelines (specifically Azure DevOps) failed, because it doesn't enable TTY input in the enviroment.

There is no reason to have both the flags -t and -i, so this PR removes the first, to fix that problem.

The PR was tested running "make test" command, that is the one modified


caseydavenport (approved):
LGTM, although you'll need to sign the CLA to make the bot happy before we can merge.

